PR #7, opened 2017-11-06
a simple layer configuration

nhanvtran:


nhanvtran:
Here's a first attempt ...

I added the layer configuration as a constructor parameter and not a template input mostly because it was simpler, but if you know how to do it with the templated functions, that would be great.

ejk43:
I really like the idea of using a configuration settings to tweak unroll parameters! Can't review until this evening 11/7 but I'll take a look as soon as possible.

Looks like the directives are commented out. Have you tested with the HLS compiler yet? 

I far prefer function inputs over template inputs, but there's a fair amount of parameters that need to be compile-time decisions for the hls compiler to work, so templates becomes a necessary evil. For example, array/loop sizes need to be templates for the compiler to figure everything out. I'm very curious to try with unroll factors.

nhanvtran:
Thanks, so after some more testing, this is NOT working -- as you say, we cannot just pass constructor inputs instead of template inputs.  But hopefully we can use this as a testbed/discussion tool.  We (with @benjaminkreis) are playing around with adding as a template input, but not obvious how to make it work

benjaminkreis:
To add one bit of information, we were able to control the unrolls successfully if the template parameter was an int, similar to N_IN or N_OUT [here](https://github.com/hls-fpga-machine-learning/hls-fpga-machine-learning/blob/master/nnet_utils/nnet_layer.h#L28).  

We are just missing something when it comes to writing a templated function that accepts structs.

nhanvtran:
Nice! we were struggling with the syntax today but this looks great.  Also ran into the issue of having to make an intermediate value.  So we should test/merge this and then you want to make a centralized repo for the git submodule?

ejk43:
> So we should test/merge this and then you want to make a centralized repo
for the git submodule?

Sure, I'd say give it a shot and see if it makes sense. I don't want to
hold up your development though so definitely don't wait on me for the
submodule-- I'll just pull in whatever you're using when I start to
integrate.

By the way, I also added a full_unroll option to the struct in case we
wanted to specify pragmas for a full unroll... Any other ideas for
optimizations that would be useful to include in the struct-template API?



On Nov 7, 2017 8:54 PM, "nhanvtran" <notifications@github.com> wrote:

Nice! we were struggling with the syntax today but this looks great. Also
ran into the issue of having to make an intermediate value. So we should
test/merge this and then you want to make a centralized repo for the git
submodule?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/hls-fpga-machine-learning/hls-fpga-machine-learning/pull/7#issuecomment-342684481>,
or mute the thread

nhanvtran:
Ok, I think it's cleaned up and ready to go, thanks! 

I added a few ideas of other things we'd like to configure going forward. Thoughts?  
- roll factor in and out
- @ejk43 you commented out the storing of arrays in LUT or BRAMs, should this be configurable if you have a lower latency design?
- if we do not have a full unroll, do we want to do a cyclic partition of the arrays with a commensurate factor? 
- I thought about a parameter which prunes low weights but maybe we do that in the translation and then put in a flag to not perform the operation in the HLS?

@benjaminkreis you merging?  we didn't set up a formal procedure other than probably if you made the PR, don't merge it :)

benjaminkreis:
will merge in a bit after checking

benjaminkreis:
Okay, it looks good to me.  I might have the same question as @nhanvtran, which is I'm not how these complete array partition pragmas will interact with partial unrolls, but I suppose we will find out.  Merging

PR #14, opened 2017-11-11
WIP: More struct usage for Input Typenames

ejk43:
To my surprise, it seems that we can actually define typenames inside a template struct. (Relevant nnet changes here: https://github.com/hls-fpga-machine-learning/hls-fpga-machine-learning/compare/ejk/more-struct-usage?expand=1#diff-b37c065f136460b015788b96b5c25102)

I put together a new proof of concept that pushes the bias, weights, and accumulator typedefinitions into the config structure. I have not edited the python generation script yet, so it's not fully compatible with everything else (hence the Work In Progress), but I thought I'd float the idea to see how far you'd like to go with the struct input.

Here's example code from the main function:

```C++
nnet::compute_layer<input_t, layer1_t, config1>(data, logits1, w1, b1);
nnet::relu<layer1_t, layer1_t, N_LAYER_1>(logits1, layer1_out);
```

Personally I think it cleans up the interface nicely and becomes very explicit about the type definitions per layer (ie, each layer has their OWN config structure, with their own internal typedefs). Thoughts?


example-prjs/simple-demo/firmware/parameters.h
```diff
@@ -0,0 +1,46 @@
...
+#include "ap_int.h"
+#include "ap_fixed.h"
+#include "nnet_layer.h"
+
+//hls-fpga-machine-learning insert numbers
+#define N_INPUTS 10
+#define N_LAYER_1 32
+#define N_OUTPUTS 1
+
+typedef ap_fixed<32,10> accum_t;
```
nhanvtran: So these 7 lines go away and the default values are specified in the YAML?

nhanvtran:
Looks great to me.  We did talk early on about how important it would be to change the precision by input and by layer.  I'm only a little worried about if its overkill for the user -- so I included a specific question about how the user would configure this in the YAML.  Then they could go in after the fact and really Tweak things by layer if they really wanted, but say a casual user wouldn't have to worry about it.

Also I like the example-prjs directory idea -- it's nice to have something people can look at.

example-prjs/simple-demo/firmware/parameters.h
ejk43: I havent gotten that far yet! If anything, I think the general list of typedefs would go away, to be replaced by specific typedefs inside each struct.

ejk43:
Sure-- or we could perhaps configure the defaults using a global typedef, that the user could then override later if needed? Example code below of what this could perhaps looks like?

```C++
typedef ap_fixed<32,8> input_t;
typedef ap_fixed<32,8> result_t;
typedef ap_fixed<32,8> layer1_t;

typedef ap_fixed<32,10> accum_default_t;
typedef ap_fixed<32,8> weight_default_t;
typedef ap_fixed<32,8> bias_default_t;

//hls-fpga-machine-learning insert layer-config
struct config1 : nnet::layer_t {
        static const unsigned n_in = N_INPUTS;
        static const unsigned n_out = N_LAYER_1;
        [ etc ]
        // TODO: Override default types with specific typing as needed
        typedef accum_default_t acc_t;       
        typedef bias_default_t bias_t;
        typedef weight_default_t weight_t;
        };
``` 



By the way, I added an issue to try to capture this question of how to interact with data typing: https://github.com/hls-fpga-machine-learning/hls-fpga-machine-learning/issues/15

nhanvtran:
Yeah, I like this approach with the global default typedef that could be overridden later

ejk43:
This PR has been superceded by new PR, which includes and adds to this branch: https://github.com/hls-fpga-machine-learning/hls-fpga-machine-learning/pull/18

Closing the PR now- leaving unmerged

PR #18, opened 2017-11-14
WIP: compute_layer resource-reuse-API for parallel inputs

ejk43:
First draft of the compute_layer "resource reuse API" for parallel inputs only. 

Still to do:
1. Add option for "serial" inputs
2. Code generation from keras with new struct-based API

Thought I'd open the PR so we can look at the potential edits. 

Works pretty well for the "simple-demo" example project (just change the unroll_factor to achieve a lower DSP usage)

Feel free to suggest any edits here! Especially API related.


nhanvtran:
Wow!  this is awesome.  With basically the `ALLOCATION` pragma, HLS figures out the II and everything.  

I'd say we go with this implementation.  I'm happy with RollFactor -> ReuseFactor, especially since we allocate the resources by hand and we kept getting confused between roll and unroll factor ;).  

Otherwise, I see things just need to be updated in the translation script.  So in the YAML, we should define the default `ReuseFactor`  and `Precision` or something like that?  

Finally, I have one resource usage question (see the line-by-line comment)

nnet_utils/nnet_layer.h
```diff
@@ -20,67 +20,85 @@
...
     }
 
-    NewInput: for(int ii = 0; ii < CONFIG_T::n_in; ii++) {
-        #pragma HLS UNROLL factor=unroll_factor_in
-        data_cache = data[ii];
-        Product: for(int jj = 0; jj < CONFIG_T::n_out; jj++) {
-        #pragma HLS UNROLL factor=unroll_factor_out
-            acc[jj] += data_cache * weights[ii][jj];
+    // Accumulate multiplication result
+    Accum1: for(int ii = 0; ii < CONFIG_T::n_in; ii++) {
```
nhanvtran: For this 2D loop do we have to do anything special for the resource usage?  The accumulator is not a big worry from the resource-usage side (what is it using, just LUTs?)?

ejk43: Yes, accumulator is just using logic-- it will get big for lots of input features!

I believe for this case we do not need to do anything special. I seem to remember trying a dedicated PIPELINE II=4 to match the resource reuse factor, but the logic consumed was unchanged (might be good to verify). If true, that would suggest the compiler is recognizing the higher II and already capturing that without an explicit directive. 

We may need to add extra pragmas for the serial IO situation

ejk43:
> Otherwise, I see things just need to be updated in the translation script. So in the YAML, we should define the default ReuseFactor and Precision or something like that?

Yes, sounds good to have a default reuse factor and precision in the yaml. 

 I can't promise I won't want to change things again later, especially for different later types, but I'm pretty happy with this cut of the API. Feel free to integrate if you'd like I won't be able to work on it for a few days

nhanvtran:
Ok sounds good.  I can try to integrate this.  Might not get to it until tomorrow afternoon or Thurs.  

We can certainly update it if a different layer type needs a different API.  Maybe we'll have a totally different method too...

nhanvtran:
Ok, I think I've reproduced your example project integrated into the scripts.  I pulled it into another branch, will make that PR

PR #19, opened 2017-11-16
integrated version of PR18

nhanvtran:
Has the new layer and activation configurations.  A few bug fixes too


nhanvtran:
Ah, so I also tried to implement to io_serial case.  It seems the straightforward way to do this is just in the top function to make all the arrays into streams.  What do y'all think?

e.g. `#pragma HLS STREAM variable=data,logits1,layer1_out`

ejk43:
That's a good find! We should probably use that in general, since it dictates how we transfer data between different layers. I believe in most cases we'll have sequential accesses, so we definitely want to use a streaming interface rather than a ping-pong buffer.

benjaminkreis:
I'm back and had a look at all the developments today with @nhanvtran and @jmduarte.  A lot of progress!

There is one problem I am observing, which is that according to the cosim, when the reuse factor is 1, the block accepts an input every clock, but it only gives an output every two clocks.  So with each new set of inputs, the time you wait for the corresponding output increases by one clock.  I'm not even sure how this is possible on an FPGA -- seems like at some point it would just fill up.  I played around a bit and found that if I replace the dataflow pragma with `HLS PIPELINE II=1` we get the expected behavior -- one input every clock and one output every clock.  I also see the FF and LUT usage go down jus a bit.  Then, if I change the reuse factor to 2 and keep the new pipeline pragma unchanged, we give one input every 2 clocks, get an output every 2 clocks, and use half the number of DSPs.  So apparently it's not clobbering reuse factor, which is good, because it means we could just leave it there.  @ejk43 do you foresee any problems with making this change?

ejk43:
@benjaminkreis  What was your test case to see this dataflow error?

The issue with replacing dataflow with pipeline is that it fully unrolls all loops in subfunctions. Should be fine for now, but could cause compatibility problems with other architectures.

Issue #13, opened 2017-11-09
RNN/LSTM in HLS Library


ejk43:
Fill out the plausible types of network layers:

- Convolutional
- Recurrent

What else am I missing?

nhanvtran:
This isn't necessarily a layer type but one thing that we definitely want to have is a regression example.  This is important for fitting/determining physics parameters and already are many use cases.

ejk43:
What do you mean by regression example?

I'm happy to start working on these new layer implementations. My typical process here is:
1. Start with a tensorflow example, which I'd create and train
2. Implement the functionality of the TF function in C++ code. Confirm equivalence using floating point types
3. Edit typing and compiler directives to achieve a decent resource target in HLS

If you have concrete examples ready (for example, something like this article here: https://www.tensorflow.org/tutorials/recurrent), that could save some time on Step 1 and help us be clear about what *exactly* we'd like to implement. Usually I just try to implement the tensorflow function (matmul is a single fully-connected layer. I targeted tf.nn.conv1d for the 1-dimensional convolution. It looks like LSTM would be what we need for recurrent NN). I'm not too familiar with Keras yet, but I'd be happy with examples there too. Thoughts?

By the way, the RNN may be tricky in HLS-- the compiler tends to choke on feedback like that. But we'll see.. should be able to make it work

nhanvtran:
Sorry still learning :)... @jmduarte is traveling but usually sets us (me) straight on these things.  So nothing new for the regression is needed (just the linear activation @benjaminkreis put in) -- so we just want to add an example weight file?  

In any case, on the examples, we plan to have a repo for setting up the training too with common data files, etc.  -- I put this under the team "Project" tab.  @jmduarte is working on the Keras training so that at least could be common (hopefully in the next few days), but it is also in the Issues to include a TensorFlow implementation as well.  

We'll try to get you data files asap but starting with anything is totally fine.  Starting to think about a RNN would be great!  We still didn't port over your CNN example either...

ejk43:
I'm learning too! I still dont know what exactly you mean by the regression..? Does this refer to a logistic regression?

Keras implementation would definitely be fine as a reference, if/when that's available. 

> We still didn't port over your CNN example either...

Convolution was not fully fleshed out, anyway! I only implemented a small subset of possible 1D convolutions. This needs lots of work :)

nhanvtran:
Yes, basically we were discussing something like this
https://deeplearning4j.org/logistic-regression

[...]

PR #22, opened 2017-12-05
first implementation of Conv1D from EJ

jmduarte:
Related to issue #13  

Took @ejk43's implementation of Conv1D layer from https://github.com/Xilinx/RFNoC-HLS-NeuralNet/blob/master/rfnoc/hls/nnet_lib/nnet_conv.h

Things changed:
- removed `hls::streams`
- changed output logic so output size is the same as input size (zero padding added only to the left of inputs)

Things to update/make configurable:
- configurable "zero" padding: valid, same, full, etc.
- number of filters
- reuse factor

@benjaminkreis also has some ideas about trying to make everything more parallelizable


nhanvtran:
is there some code for testing?

jmduarte:
added the test project code at bf723b9

ejk43:
> changed output logic so output size is the same as input size (zero padding added only to the left of inputs)

Does this represent the "same" padding? 

As a related question, what other requirements would you like to see from conv1d and conv2d, especially regarding items like number of input/output channels for the convolution? One thing I found earlier,  at least with the "full serial" implementation, was that more options means more ways the HLS directives can go wrong...

pierinim:
In terms of inputs to the 1D, we could imagine passing a list of 10 particles, each consisting of four numbers (energy + the angles for direction and one ID number for the kind of particle). The same input file could be used for LSTM. I would start with a binary classifier, so that the number of output layers is minimal

benjaminkreis:
@ejk43 this is "same" padding in the sent the the number of inputs equals the number of outputs, but padding the data with zeros only on the "left" side might not be the standard.  We'll have to look at TensorFlow etc. to see how they split the zeros between the left and right sides.

I leave it to @jmduarte to answer about channels etc.

benjaminkreis:
Here's the answer about same padding in TF: "In this case, the bottom and right sides always get the one additional padded pixel." (https://www.tensorflow.org/versions/r0.12/api_docs/python/nn/convolution)

benjaminkreis:
The HLS csynth runs fine, so I'm going to go ahead a merge this to serve as our starting point.  
@jmduarte figured out that for each filter index, convolutional layers sum over the channels and apply one bias.

benjaminkreis:
Sorry, I clicked comment too soon.

The HLS csynth runs fine, so I'm going to go ahead a merge this to serve as our starting point.  

At least for Keras/Tensorflow, @jmduarte figured out that for each filter index, convolutional layers sum over the channels and apply one bias.  I'm going to work on making this change for a next PR, which will not only affect the internal math but also the output shape.  @ejk43, let us know if you have a different understanding from your own experience with Tensorflow!

ejk43:
> convolutional layers sum over the channels and apply one bias (see here for e.g.).

Interesting-- It's been a while, but I think I was under the impression each channel operated independently-- I must have gotten confused somewhere. Good catch! This makes more intuitive sense; that each *filter* is independent, but the channels are combined (for example, multiple channels in a color image).  Is that a correct interpretation?

benjaminkreis:
Yeah, I think that is right.  One thing I'm not sure about is how you interpret the output when you connect it to another convolutional layer.  I suppose the different filter outputs from the first layer become channels for the second layer?

PR #23, opened 2017-12-19
handling zeros from compression/pruning

nhanvtran:
an adaptation of @benjaminkreis findings.  we agree that it's doing what we want it to do to limit DSPs.  Further compression optimization can still be achieved but this gets the major DSP usage down


benjaminkreis:
I just committed a few cleanups, the main one being that I removed the array for nzero and put it into the layer dictionary.

And just to add to @nhanvtran's description, this limits the DPSs to the maximum it should use assuming the zeros are optimally aligned when there is a reuse factor >1.  It does not prevent HLS from making multipliers with FFs and LUTs, which does seem to happen to a degree -- this is the further compression optimization to be achieved.

PR #30, opened 2018-01-03
Take project name and part number from yaml config

benjaminkreis:
(also added comment with zero counter result to top of weight and bias files)


nhanvtran:
oh crap, #32 and this one are in conflict...will wait for ben to come back to resolve this. Sorry @benjaminkreis!

benjaminkreis:
ok, done @nhanvtran 

just a reminder that this now configures project name, part number, and clock period from yaml (name of PR doesn't include clock period)

PR #32, opened 2018-01-10
Use relative paths for the Keras-to-HLS conversion

ejk43:
Just a small PR that converts the relative paths from keras-config.yml into absolute paths. I ran into this issue when I recently tried to recreate a new project. 

I'm not really wedded to this idea, but it means you can point to the keras-to-hls.py file from elsewhere-- ie not in the "keras-to-hls" folder-- and still get a HLS project that works. It also calculates relative pathing so the HLS project setup points to the nnet_utils folder in the HLS4ML repo.

As an example, I added keras-config.yml to a new "demo-conversion" folder in example-prjs. 

One potential use-case that's not covered here would be if keras-config.yml contains absolute paths rather than relative paths..?

### DIFF
--- a/example-prjs/demo-conversion/README.md
+++ b/example-prjs/demo-conversion/README.md
@@ -1,0 +1,9 @@
+# Instructions to run
+
+Copy content from the keras-to-hls directory:
+
+```cp -r ../../keras-to-hls/example-keras-model-files ./```
+
+Then run the conversion script:
+
+```python ../../keras-to-hls/keras-to-hls.py -c keras-config.yml```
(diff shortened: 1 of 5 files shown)


nhanvtran:
I'm mildly concerned about not having the ability to put the absolute path but is the simple solution just to check if the path starts with `/`?

ejk43:
Oh, interesting, I just stumbled on this command: 

https://stackoverflow.com/questions/3320406/how-to-check-if-a-path-is-absolute-path-or-relative-path-in-cross-platform-way-w

```
os.path.isabs(my_path)
```

So to capture the absolute path condition we'll add a check to see if the specified path in keras-config.yml is absolute or relative.

nhanvtran:
nice!  perfect

ejk43:
New commit now allows absolute paths -- just check if its an absolute path before writing. I've tested that successfully and then entered the generated directory and built HLS.

nhanvtran:
sorry left this hanging, looks graet -- now merged

PR #40, opened 2018-02-08
Conv1D support in hls_writer 

jmduarte:
First attempt at Conv1D layer support in `hls_writer.py`

There is still some issue with this though, where the CSIM fails for certain values of the hyperparameters of the first Conv1D layer. I committed two small example models to demonstrate this. One with the first Conv1D filter size of 4 (and 4 input channels) and the other with the first Conv1D filter size of 5 (and 4 input channels). The latter goes through CSIM without a problem, while the former fails CSIM. It seems to be related to the fact that the number of filters and the number of input channels is the same (it also fails for filter size of 5 with 5 input channels for example), but the issue is not so clear.

```
# filter size of 4
example-keras-model-files/KERAS_conv1d_small.json
example-keras-model-files/KERAS_conv1d_small_weights.h5
# filter size of 5
example-keras-model-files/KERAS_conv1d_small_nfilt5.json
example-keras-model-files/KERAS_conv1d_small_nfilt5_weights.h5
```

If you try to run it the default (filter size of 4):
```
python keras-to-hls.py -c keras-config.yml
cd my-hls-dir-test
vivado_hls -f build_prj.tcl 
```
you get this CSIM failure:
```
INFO: [SIM 211-2] *************** CSIM start ***************
In file included from ../../../../firmware/parameters.h:8:0,
                 from ../../../../firmware/myproject.cpp:21:
/home/jduarte1/hls-fpga-machine-learning/nnet_utils/nnet_conv.h: In function ‘void nnet::conv_1d(data_T (*)[CONFIG_T:: n_chan], res_T (*)[CONFIG_T:: n_filt], typename CONFIG_T::weight_t (*)[CONFIG_T:: n_chan][CONFIG_T:: n_filt], typename CONFIG_T::bias_t*) [with data_T = ap_fixed<18, 8>, res_T = ap_fixed<18, 8>, CONFIG_T = config1, typename CONFIG_T::weight_t = ap_fixed<18, 8>, typename CONFIG_T::bias_t = ap_fixed<18, 8>]’:
/home/jduarte1/hls-fpga-machine-learning/nnet_utils/nnet_conv.h:87:25: internal compiler error: Segmentation fault
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://gcc.gnu.org/bugs.html> for instructions.
make: *** [obj/myproject.o] Error 1
4
    while executing
"source [lindex $::argv 1] "
    ("uplevel" body line 1)
    invoked from within
```

even though it compiles and runs outside of `vivado_hls`:
```
g++ -I/home/jduarte1/hls-fpga-machine-learning/nnet_utils -I/data/xilinx/Vivado_HLS/2016.4/include -I/data/xilinx/Vivado_HLS/2016.4/include/etc -I/data/xilinx/Vivado_HLS/2016.4/lnx64/tools/auto_cc/include -I/data/xilinx/Vivado_HLS/2016.4/include/ap_sysc -I/data/xilinx/Vivado_HLS/2016.4/lnx64/tools/systemc/include -I/home/jduarte1/hls-fpga-machine-learning/keras-to-hls/my-hls-dir-test -O0 -g3 -Wall -c -fmessage-length=0  myproject_test.cpp
g++ -I/home/jduarte1/hls-fpga-machine-learning/nnet_utils -I/data/xilinx/Vivado_HLS/2016.4/include -I/data/xilinx/Vivado_HLS/2016.4/include/etc -I/data/xilinx/Vivado_HLS/2016.4/lnx64/tools/auto_cc/include -I/data/xilinx/Vivado_HLS/2016.4/include/ap_sysc -I/data/xilinx/Vivado_HLS/2016.4/lnx64/tools/systemc/include -I/home/jduarte1/hls-fpga-machine-learning/keras-to-hls/my-hls-dir-test -O0 -g3 -Wall -c -fmessage-length=0  firmware/myproject.cpp 
g++ -o conv1d myproject_test.o myproject.o
./conv1d
```

If anyone @ejk43 @nhanvtran has a clue about what's going on let me and @benjaminkreis know!

Thanks!


nhanvtran:
So I'm not totally sure I understand what's going on with the seg fault in the different between
`KERAS_conv1d_small_nfilt5.json`
and 
`KERAS_conv1d_small.json`

But I see HLS choking on the access to the 4-d array, this piece:
`mult[ii][ff][cc][jj]`

I found a remedy this by turning all the >2-d arrays into 1-d arrays like here:
https://github.com/hls-fpga-machine-learning/HLS4ML/blob/nvt/conv1d/nnet_utils/nnet_conv.h

what do you think?

nhanvtran:
@jmgduarte also concerning csim of really big arrays for the really big CNN, apparently the arrays were too big and you created a stack overflow!
Not sure if there are other solutions but I found this
```- Segmentation fault in C simulation of the pre-synthesis test bench and design under test (DUT) or test bench in post-0synthesis co-simulation. This can occur if the code contains array which are large, as these require a large stack space. Two potential work-arounds are available:
- Using a static qualifier will move the arrays out of stack. 
- The shell command, ulimit -s, can also be used to increase stack size.```

jmduarte:
@nhanvtran, yeah i think your solution for unrolling weight/multiplication arrays that are bigger than >=3D in 0a7be79 makes sense and makes both of the small Conv1D models synthesize correctly. 

So I cherry picked that commit and added this to this PR. I also modified keras-to-hls to unroll the weight array if it's >=3D.

https://github.com/hls-fpga-machine-learning/hls4ml/pull/40/commits/78a435bcfa4734ae534dc1d3d912578941124bd7#diff-f437afd09e9b6c1edeaf9767cc871c44R54

Also, the big Conv1D model now gets past csim (after increasing the stack size with `ulimit -s 65532`), but it still gets stuck in csynth and still complains about large memory usage. So very big models might still need some work (and @benjaminkreis and I had some ideas like only initializing non-zero weights or using `static const` arrays)... but I think we're in a mergable state now.

nhanvtran:
all right let's merge it!

PR #42, opened 2018-02-10
Softmax with invert LUT

jmduarte:
Modified the softmax activation function to use an invert LUT (default 10 bits from 0 to 64 - this might need to be tuned).

The latency for the softmax layer in the 50% pruned three-layer example goes down from 35 to 8, but it's still not as good as a simple sigmoid which is 2.

softmax:
```
+ Latency (clock cycles): 
    * Summary: 
    |   35|   35|    1|    1| function |
```

softmax with invert LUT:
```
+ Latency (clock cycles): 
    * Summary: 
    |    8|    8|    1|    1| function |
```

sigmoid:
```
+ Latency (clock cycles): 
    * Summary: 
    |    2|    2|    1|    1| function |
```

### DIFF
--- a/nnet_utils/nnet_activation.h
+++ b/nnet_utils/nnet_activation.h
@@ -95,8 +95,8 @@ inline float sigmoid_fcn_float(float input) {
     return 1.0 / (1 + exp(-input));
 }
 
-template<class data_T, int N_TABLE>
-void init_sigmoid_table(data_T table_out[N_TABLE])
+template<typename CONFIG_T, int N_TABLE>
+void init_sigmoid_table(typename CONFIG_T::table_t table_out[N_TABLE])
 {
     // Default logistic sigmoid function:
     //   result = 1/(1+e^(-x))
@@ -104,7 +104,7 @@ void init_sigmoid_table(data_T table_out[N_TABLE])
         // First, convert from table index to X-value (signed 8-bit, range -8 to +8)
         float in_val = 2*8.0*(ii-float(N_TABLE)/2.0)/float(N_TABLE);
         // Next, compute lookup table function
-        data_T real_val = sigmoid_fcn_float(in_val);
+        typename CONFIG_T::table_t real_val = sigmoid_fcn_float(in_val);
         //std::cout << "Lookup table In Value: " << in_val << " Result: " << real_val << std::endl;
         table_out[ii] = real_val;
     }
@@ -114,8 +114,8 @@ template<class data_T, class res_T, typename CONFIG_T>
 void  sigmoid(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 {
     // Initialize the lookup table
-    res_T sigmoid_table[CONFIG_T::table_size];
-    init_sigmoid_table<res_T, CONFIG_T::table_size>(sigmoid_table);
+    typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
+    init_sigmoid_table<CONFIG_T, CONFIG_T::table_size>(sigmoid_table);
 
     if (CONFIG_T::io_type == io_parallel){
         #pragma HLS PIPELINE
@@ -130,7 +130,7 @@ void  sigmoid(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
         index = data_round + 8*CONFIG_T::table_size/16;
         if (index < 0)   index = 0;
         if (index > CONFIG_T::table_size-1) index = CONFIG_T::table_size-1;
-        res[ii] = sigmoid_table[index];
+        res[ii] = (res_T) sigmoid_table[index];
     }
 }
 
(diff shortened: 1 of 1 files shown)


nhanvtran:
cool! is the performance functionally the same or do you have plots of comparisons for accuracy?  (clearly I didn't test it yet)

nhanvtran:
ah nevermind, I see you flipped the LUT to remove the divide...

nhanvtran:
ok, last comment maybe...

what if you do the invertLUT but doing a floating point multiplication instead of fixed point, like @jngadiub was trying to recover performance and save latency?

jmduarte:
@nhanvtran I haven't extensively tested these choices of table size/range/precision, but it works for our testbed example (all zero inputs).  Running through the full set of testing inputs (and maybe other models like the conv1d) and checking that it works for each one would be worthwhile. 

Also, yeah floating point might work better. Is @jngadiub's floating point multiplication example committed somewhere?

nhanvtran:
how do i point to a specific comment in an issue thread?  anyway, it's @jngadiub comment from early this morning
https://github.com/hls-fpga-machine-learning/HLS4ML/issues/31

jmduarte:
I brought in @jngadiub's commit b6480aa8cb52b0d6958fc6bd7556c01b0b1f6ab6 and added the invert LUT approach there too (also with floating point)

Then the latency becomes large again:
```
+ Latency (clock cycles): 
    * Summary: 
    |   59|   59|    1|    1| function |
```

Another possibility is to decouple the precision of the LUT calculation from the input/result precision. Like having it be `ap_fixed<18,8>` always (or configurable) regardless of the other precision.

I guess this adds another dimension to the scan, but might be worth it since the latency is 8 for `ap_fixed<18,8>` vs 59 for floating point...

benjaminkreis:
@jmduarte, when you say "precision of the LUT calculation", you mean the precision of exp_res and exp_res_sum_invert, right?  Can you check the latency for a couple other values of ap_fixed?  Say <24,10> and <32,12>?

jngadiub:
@jmduarte do we need the floating point softmax definition if you have changed such that one can anyways configure the preferred precision in the variable table_t here

https://github.com/hls-fpga-machine-learning/HLS4ML/blob/dac6d2e0734220bc2106b73aa8c16d7674e1fe17/nnet_utils/nnet_activation.h#L137

One can then just typedef the table_t to float.

Can we instead configure the table_t variable from outside such that one can make an automatic scan?

jmduarte:
@jngadiub good idea, I'll update this PR to make table_t configurable

jmduarte:
`table_t` is now configurable... currently only used for exp and invert calculations... but we can do the same thing for the sigmoid and tanh

jmduarte:
Following @jngadiub's and @pierinim's suggestion for computing 1/p in the softmax layer first, then inverting. Let me know if this is what you guys were thinking... or feel free to fix it!

e.g. for 3 inputs
```
a1 = 1 + exp(x2-x1) + exp(x3-x1)
a2 = exp(x1-x2) + 1 + exp(x3-x2)
a3 = exp(x1-x3) + exp(x2-x3) + 1

p1 = 1/a1
p2 = 1/a2
p3 = 1/a3
```

pierinim:
Yes! Would be nice to know from @jngadiub if we gain anything with it. By eye it looks like...

jngadiub:
@jmduarte @pierinim I have summarized and compared the different softmax implementations we have tried so far

https://cernbox.cern.ch/index.php/s/tPzvM6MfusO8BhS

It seems that by doing the inversion there is a significant degradation in performance. I am investigating.

nnet_utils/nnet_activation.h
```diff
@@ -186,50 +186,47 @@ void  softmax(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
...
-        index = data_round + 8*CONFIG_T::table_size/16;
-        if (index < 0)   index = 0;
-        if (index > CONFIG_T::table_size-1) index = CONFIG_T::table_size-1;
-        exp_res[ii] = exp_table[index];
-        exp_res_sum += exp_table[index];
+      exp_res[ii] = 0;
+      for (int jj=0; jj<CONFIG_T::n_in; jj++) {
+	if (ii==jj) exp_diff_res[ii][jj] = 1;
+	else {
+	  data_round = (data[ii]-data[jj])*CONFIG_T::table_size/16;
```
jngadiub: Found one bug. This should be data[jj]-data[ii].

jmduarte: ack you're right! is it fast to redo the scan with this fix? i'm checking quickly that this gives the same output for the test bench

jngadiub: mmh, it takes about 3 hours. I already checked the results of the testbench for <18,8> and after the fix I get the same results as for the other softmax implementations. But there are other problems with the division which I am looking at now. For instance for <16,4> I see a lot of zeros that should not be there. i'll let you know what else I find out.

jmduarte: ok if we fix that bug we should be mostly there...

`master` branch, example-prjs/conv1-layer:
```
0.527591 (expected 0.528625, -0.195696 percent difference)
0.351451 (expected 0.352447, -0.282627 percent difference)
0.119575 (expected 0.117572, 1.70421 percent difference)
0.000261545 (expected 0.000254869, 2.61927 percent difference)
8.61883e-05 (expected 8.45194e-05, 1.97461 percent difference)
0.0010345 (expected 0.00101596, 1.82458 percent difference)
```

`jmgd/softmax-invert-lut` branch (bug)
```
0 (expected 0.528625, -100 percent difference)
0.0234375 (expected 0.352447, -93.3501 percent difference)
0 (expected 0.117572, -100 percent difference)
0.238281 (expected 0.000254869, 93391.5 percent difference)
0.726563 (expected 8.45194e-05, 859540 percent difference)
0.0585938 (expected 0.00101596, 5667.32 percent difference)
```

`jmgd/softmax-invert-lut` branch + bug fix:
```
0.533203 (expected 0.528625, 0.865951 percent difference)
0.355469 (expected 0.352447, 0.857234 percent difference)
0.118164 (expected 0.117572, 0.503923 percent difference)
0.0273438 (expected 0.000254869, 10628.5 percent difference)
0.015625 (expected 8.45194e-05, 18386.9 percent difference)
0 (expected 0.00101596, -100 percent difference
```

nnet_utils/nnet_activation.h
```diff
@@ -137,69 +141,93 @@ inline float exp_fcn_float(float input) {
...
+      }
     }
 
-    //Second loop to divide by sum of exponentials
+    //Second loop to invert
     for (int ii=0; ii<CONFIG_T::n_in; ii++) {
-      // #pragma HLS UNROLL
-      res[ii] = exp_res[ii]/exp_res_sum; //Note division used here!
+      int exp_res_index = exp_res[ii]*CONFIG_T::table_size/64;
+      if (exp_res_index < 0)   exp_res_index = 0;
```
jngadiub: It seems like here there is some sign flip happening on the exp_res due to the precision of the table which I am setting now as the precision of the rest. We know already about this problem. In fact, I plan to redo the scans by setting the precision of the table to something reasonable like <18,8>.

benjaminkreis: @jngadiub it sounds like you figured it out.  The only other thought is that `table_size` could be playing a role, in which case increasing it would effectively increase precision of the outputs.  If changing table_t to <18,8> doesn't get you all the there, I'd try that.

jngadiub:
I repeated the scans for the default softmax, the softmax invert 1 and softmax invert 2 by fixing the table precision to <18,8>

softmax_resources
softmax_resources

So now the performance for floating point softmax, softmax invert 2 and default softmax are very similar. While softmax invert 2 wins in the resource usage. Instead, I am not sure what's happening with the top tagger class.

I debugged a little what's happening with softmax invert 1 (the first implementation from @jmduarte) and I understood the following things:

* the softmax invert 2 is more protected against large exponential as we do the exp of a difference
* the exp itself or the sum of exps can generally become quite large and leave the range allowed by the number of integer bits set for the exp table
* one can increase the number of integer bits but one also has at the same time to increase the amplitude and precision of the table used for the division of a large exp

So, the softmax invert 1 needs some careful tuning because things seem to get very correlated. While the softmax invert 2 assures much more stability. Shall we use this for the results in the paper?

Then I would also like to check if using a different standardization method of the inputs would help to recover some performance. Right now the standard scaler is normalizing between -2 and +2 while a min-max scaler would normalize between -1 and 1. Something to try.
Last thing to check if it's increasing the precision of the tables can help as well, as suggested by @benjaminkreis.

nhanvtran:
@jngadiub very nice to see!  it's good to see this converging

> So, the softmax invert 1 needs some careful tuning because things seem to get very correlated. While the softmax invert 2 assures much more stability. Shall we use this for the results in the paper?

I would say yes! I'm still think we want to start with a value of <X,Y> where the value of Y returns back floating point performance at a given X.  So it looks like Y = 6 or 8 is better.  This is because of the standardization to -2,2 which I think is okay.  Moving to -1,1 might make <X,4> more reasonable but in the end, I'm guessing you would need more bits for X to achieve the same performance so it's a neutral change in the end and doesn't affect the qualitative results.

We should anyway make sure we write up these lessons to learn in the paper :)

jngadiub:
Right! I have done the <X,Y> scan for the floating point softmax some time ago

[scan_allAUCs.pdf](https://github.com/hls-fpga-machine-learning/hls4ml/files/1748228/scan_allAUCs.pdf)

I'll repeat it for the softmax invert 2 and with few more sampling points before the plateau.

nhanvtran:
Ah great!  Sorry I missed it, sounds like a good plane

violatingcp:
Nice. Does this mean we don't need to modify our Keras training for reduced
precision?

jmduarte:
One other possibility along the lines of making the softmax more stable is to always compute 
exp(xi-m) where m = max(xi) a la https://stats.stackexchange.com/a/304774

This is similar to what the softmax invert 2 is already doing... but it would actually be easier to implement as an improvement on the softmax invert 1:

```
m = max(x1, x2, x3)
s = e^(x1-m) + e^(x2-m) + e^(x3-m)
p1 = e^(x1-m) * 1/s
p2 = e^(x2-m) * 1/s
p3 = e^(x3-m) * 1/s
```

PR #45, opened 2018-02-19
serial mode, fixed II

nhanvtran:
This makes a very simplistic serial mode work.  Also explicitly declares the II inside of nn_layer.h to fix any weird pipeline optimizations.

do i need to add this fix to nn_conv.h and also any changes needed for PyTorch?

### DIFF
--- a/keras-to-hls/keras-config.yml
+++ b/keras-to-hls/keras-config.yml
@@ -5,6 +5,6 @@ ProjectName: myproject
 XilinxPart: xc7vx690tffg1927-2
 ClockPeriod: 5
 
-IOType: io_parallelv # options: io_serial/io_parallel
+IOType: io_parallel # options: io_serial/io_parallel
 ReuseFactor: 1
 DefaultPrecision: ap_fixed<18,8> 
--- a/keras-to-hls/keras-config.yml
+++ b/keras-to-hls/keras-config.yml
@@ -1,10 +1,10 @@
-KerasJson: example-keras-model-files/KERAS_conv1d_small.json
-KerasH5:   example-keras-model-files/KERAS_conv1d_small_weights.h5
-OutputDir: my-hls-dir-test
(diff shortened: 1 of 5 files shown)


keras-to-hls/keras-config.yml
```diff
@@ -1,10 +1,10 @@
...
-OutputDir: my-hls-dir-test
+KerasJson: example-keras-model-files/KERAS_1layer.json
+KerasH5:   example-keras-model-files/KERAS_1layer_weights.h5
+OutputDir: my-hls-test
 ProjectName: myproject
 XilinxPart: xc7vx690tffg1927-2
 ClockPeriod: 5
 
-IOType: io_parallel # options: io_serial/io_parallel
+IOType: io_parallelv # options: io_serial/io_parallel
```
benjaminkreis: typo?  parallelv

benjaminkreis:
PS your changes are in hls_writer which is used by both the keras and pytorch converter, so this should be all that's required

nhanvtran:
@benjaminkreis yes, I just keep the reshapes and partitions for now.  I don't think there's anything wrong with it particularly.  However, I do see that the partitioned arrays are still stored in BRAMs to use the resources efficiently.  

thanks for picking up on the typo -- it's fixed now.

for the conv layer, we need to add the configuration which asks for io_type

benjaminkreis:
This looks okay to me, at least as a first version.  As you know, the IP expects all inputs on one clock edge, so if the data is streaming in, the user would have to add a wrapper around it.  The long term goal could be to recover the streaming functionality, but no reason not to merge this now in my opinion.

ejk43:
Hey guys, take a look here as a follow-up PR: https://github.com/hls-fpga-machine-learning/hls4ml/pull/48

I think I've hit a good mix of directives that gets back towards the serialized behavior I hit last year in the dense layer.

Issue #46, opened 2018-02-19
store_in_bram functionality


nhanvtran:
So far we have 2 working modes -- one for LHC trigger (low reuse factor, 1-6, weights in the fabric) and one for "naive" serial mode (see PR #45).  

One interesting mode is a very large reuse factor and weights stored in the fabric.  This is particularly useful for really big networks and SDAccel-like use cases.  

However, playing around with the current code, I found that it's not so trivial to store weights in BRAMs and run with `#pragma HLS PIPELINE`.  HLS always wants to partition the weight array.  

@ejk43 you have any ideas here?  I thought we could remove the PIPELINE and go back to the days of DATAFLOW pragma...

nhanvtran:
> Storing weights in BRAMs in parallel mode (using high reuse factors) should be feasible. I may take a look at this. I doubt removing the top-level PIPELINE is a good idea -- it just works so well for so many algorithms-- but it sounds like it should be feasible.

This would be good to conceptualize.  We agree it _should_ be possible -- the question was how clunky it would be.  We thought that one would start by making small arrays with length == reuse factor and read off the arrays from BRAMs in that way, but there are complications with pruned models.  Maybe you have some more native way to implement this?  @GiuseppeDiGuglielmo was also thinking about the memory storage perhaps in other global schemes including this issue so maybe it'll be good to stay in touch here

> Does this issue only apply to dense layers? Or do you want to consider conv layers too? Any other high-priority layers that would want to save weights in BRAM (maybe LSTM)?

I think we want to extend this to many architectures but I think we agree that it's most transferrable from dense to conv.  This goes with your other comment about serial mode in conv.  The LSTM case is interesting ... in the `static` case it's not as important as the fully unrolled case.

violatingcp:
For the store in b-ram functionality. I think we can follow the sub-layer
setup. Basically when we start splitting the matrices into smaller ones to
limit eh multiplications we can store the weights of the elements not being
multiplied in block ram. I will run some quick tests to see how this works,
but it seems the most natural use for me. Also if this is the case, it
should be easy to add this to any of the setups that already using the
sublayer tech (dense/LSTM/conv?)

ejk43:
I have a possible implementation here: https://github.com/hls-fpga-machine-learning/hls4ml/commits/ejk/add-brams-to-parallel-mode

[...]

PR #48, opened 2018-02-20
Implement Serial mode directives in Dense Layer

ejk43:
I saw updates to the recent serial mode PR (https://github.com/hls-fpga-machine-learning/hls4ml/pull/45) -- Looks good!!

I noticed that the top-level IOs were still fully partitioned, and the dense layer directives were too aggressively serialized-- since essentially no directives were applied in serial mode at all, HLS defaults to *everything* being serialized. It's a conservative approach, but generally not very useful from a latency/throughput standpoint  :) 

This PR adds a few edits that make the dense layer in nnet_layer.h somewhat useful in serial mode! I'm pretty happy with how it turned out, and it's pretty close to what I achieved last year in my original library. I think it should be merge-able -- please review / provide feedback otherwise. 

Quick overview of edits:
 1. Removes all ARRAY_RESHAPE / ARRAY_PARTITION directives from the top level. This lets the top-level IOs remain serial inputs and outputs. I also converted the IOs to "handshake" type (ap_hs).
 2. Adds the DATAFLOW directive to the top level,  and corresponding HLS STREAM directives that indicate the variables should be implemented as streaming FIFOs instead of ping-pong ram (https://www.xilinx.com/html_docs/xilinx2017_2/sdaccel_doc/topics/pragmas/ref-pragma_HLS_stream.html)
 3. Set up DATAFLOW, PIPELINE, and ARRAY_PARTITION directives in the dense layer to achieve a "good" balance of resources vs latency for serial mode, including a reasonable usage of the reuse_factor.

Here's a quick overview of my assumptions/expectations for serial-mode:
 - The number of multipliers in the dense layer will be driven by the N_OUT parameter; that is, the matrix-multiply for-loop will be pipelined (by default) such that there are N_OUT parallelized multipliers.
 - The dense layer will be able to accept one new input per clock, perform the matrix multiply, and then clock out one output value at a time. The II for the serialized dense layer should be nominally "close" to the maximum of N_IN and N_OUT (I do see higher IIs, likely due to other glue logic)
 - Adding a reuse_factor will reduce the number of multipliers required for the dense layer, at the expense of a larger II into the matrix multiply. This does not necessarily mean the total II is higher for the dense layer... for example, the 10x32 dense layer in the basic example maintains the same total II when reuse_factor = 2 or 3, because this means it takes 20 or 30 samples to clock in the 10 inputs rather than 10 clocks, but the II was already > 32 because of the output size. 

Utilization result from the basic dense-layer example: 
```
* Summary: 
|Expression       |        -|      -|       -|       -|
|FIFO             |        0|      -|      15|      90|
|Instance         |       17|     33|    1781|    3749|
|Multiplexer      |        -|      -|       -|       -|
|Register         |        -|      -|       9|       -|
|Total            |       17|     33|    1805|    3839|
|Utilization (%)  |    ~0   |   ~0  |   ~0   |   ~0   |
```

And the ports show only 18-bit interfaces to data and res: 
```
+-----------------------+-----+-----+------------+----------------+--------------+
|       RTL Ports       | Dir | Bits|  Protocol  |  Source Object |    C Type    |
+-----------------------+-----+-----+------------+----------------+--------------+
|data_V                 |  in |   18|    ap_hs   |     data_V     |    pointer   |
|data_V_ap_vld          |  in |    1|    ap_hs   |     data_V     |    pointer   |
|data_V_ap_ack          | out |    1|    ap_hs   |     data_V     |    pointer   |
|res_V                  | out |   18|    ap_hs   |      res_V     |    pointer   |
|res_V_ap_vld           | out |    1|    ap_hs   |      res_V     |    pointer   |
|res_V_ap_ack           |  in |    1|    ap_hs   |      res_V     |    pointer   |
|const_size_in          | out |   16|   ap_vld   |  const_size_in |    pointer   |
|const_size_in_ap_vld   | out |    1|   ap_vld   |  const_size_in |    pointer   |
|const_size_out         | out |   16|   ap_vld   | const_size_out |    pointer   |
|const_size_out_ap_vld  | out |    1|   ap_vld   | const_size_out |    pointer   |
|ap_clk                 |  in |    1| ap_ctrl_hs |    myproject   | return value |
|ap_rst                 |  in |    1| ap_ctrl_hs |    myproject   | return value |
|ap_start               |  in |    1| ap_ctrl_hs |    myproject   | return value |
|ap_done                | out |    1| ap_ctrl_hs |    myproject   | return value |
|ap_idle                | out |    1| ap_ctrl_hs |    myproject   | return value |
|ap_ready               | out |    1| ap_ctrl_hs |    myproject   | return value |
+-----------------------+-----+-----+------------+----------------+--------------+
```

Timing and Latency also look nice:

```
+ Timing (ns): 
    * Summary: 
    +--------+-------+----------+------------+
    |  Clock | Target| Estimated| Uncertainty|
    +--------+-------+----------+------------+
    |ap_clk  |   5.00|      4.14|        0.62|
    +--------+-------+----------+------------+

+ Latency (clock cycles): 
    * Summary: 
    |   48|   48|   38|   38| dataflow |
```

Thoughts??

### DIFF
--- a/hls-writer/hls_writer.py
+++ b/hls-writer/hls_writer.py
@@ -37,7 +37,7 @@ def hls_writer(layer_list, yamlConfig):
                 newline += '    #pragma HLS INTERFACE ap_vld port=data,res \n'
                 newline += '    #pragma HLS PIPELINE \n'
             if yamlConfig["IOType"] == "io_serial":
-                newline += '    #pragma HLS INTERFACE ap_hs port=data,res \n'
+                newline += '    #pragma HLS INTERFACE axis port=data,res \n'
                 newline += '    #pragma HLS DATAFLOW \n'
 
         #Add layers
--- a/hls-writer/hls_writer.py
+++ b/hls-writer/hls_writer.py
@@ -37,9 +37,8 @@ def hls_writer(layer_list, yamlConfig):
                 newline += '    #pragma HLS INTERFACE ap_vld port=data,res \n'
                 newline += '    #pragma HLS PIPELINE \n'
             if yamlConfig["IOType"] == "io_serial":
-                newline += '    #pragma HLS ARRAY_RESHAPE variable=data complete dim=0 \n'
-                newline += '    #pragma HLS ARRAY_RESHAPE variable=res complete dim=0 \n'
-                newline += '    #pragma HLS INTERFACE ap_vld port=data,res \n'
+                newline += '    #pragma HLS INTERFACE ap_hs port=data,res \n'
+                newline += '    #pragma HLS DATAFLOW \n'
 
         #Add layers
         elif '//hls-fpga-machine-learning insert layers' in line:
@@ -91,7 +90,7 @@ def hls_writer(layer_list, yamlConfig):
                     elif layer_list[i-1]['class_name']=='Conv1D':
                         newline += '    {} layer{}_out[{}*{}];\n'.format(output_type,i,y_out,n_filt)
                     if yamlConfig["IOType"] == "io_parallel": newline += '    #pragma HLS ARRAY_PARTITION variable=layer{}_out complete dim=0\n'.format(i)
-                    if yamlConfig["IOType"] == "io_serial":   newline += '    #pragma HLS ARRAY_PARTITION variable=layer{}_out complete dim=0\n'.format(i)
+                    if yamlConfig["IOType"] == "io_serial":   newline += '    #pragma HLS STREAM variable=layer{}_out depth=1\n'.format(i)
 
                 #Compute Dense layer
                 if layer_list[i-1]['activation'] == "linear" and layer_list[i-1]['class_name']=='Dense':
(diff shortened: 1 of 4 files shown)


nhanvtran:
Thanks @ejk43!  A much better serial mode :)...what was in there for #45 was mostly "something that worked and was the most literal meaning of *serial*"

This is much more functional and mostly addresses what I was getting at in issue #46.  I have some more detailed questions (see next immediate post) but I wanted to make sure I understand the concept.  

For `io_parallel` (with reuse = 1), the underlying principle is "infer this network as fast as possible, I don't care how many resources this takes".

Do I understand right that `io_serial`, the underlying principle as you've implemented it here is "infer this network as fast as possible with the constraint that the inputs and outputs are streamed in and out"?

nhanvtran:
Technical questions on this PR

Main technical observation:  When I used a reuse=2 the latency and II blew up more than expected.  And it all comes from the small layer (32x1).  I tried digging around to see what was the cause, but I didn't find it yet.  Is this the intended behavior?
```
  * Summary: 
    |  164|  164|  162|  162| dataflow |

    + Detail: 
        * Instance: 
        +--------------------------------+---------------------+-----+-----+-----+-----+----------+
        |                                |                     |  Latency  |  Interval | Pipeline |
        |            Instance            |        Module       | min | max | min | max |   Type   |
        +--------------------------------+---------------------+-----+-----+-----+-----+----------+
        |grp_compute_layer_0_0_1_fu_164  |compute_layer_0_0_1  |   58|   58|   32|   32| dataflow |
        |grp_compute_layer_0_0_fu_173    |compute_layer_0_0    |  162|  162|  162|  162| dataflow |
        |grp_sigmoid_fu_181              |sigmoid              |    3|    3|    1|    1| function |
        |grp_relu_fu_190                 |relu                 |   34|   34|   34|   34|   none   |
        |StgValue_37_Block_proc_fu_196   |Block_proc           |    0|    0|    0|    0|   none   |
        +--------------------------------+---------------------+-----+-----+-----+-----+----------+
```

My other questions are really probably more for my own education...
a. why did you add the pipelines to the activation functions? it looks like that would do nothing.
b. why did you set the IO ports to `ap_hs` instead of `ap_fifo` or `axis`?
c. now i see that weights are stored in BRAMs which is great -- should that be the default and you store the weights in `lutram`s if specified by user?
d. how does compression get handled in this scheme?  it seems like it doesn't naturally take advantage of a sparse weight matrix without some major refactoring of the weights (or removing whole neurons at a time)

ejk43:
> Do I understand right that io_serial, the underlying principle as you've implemented it here is "infer this network as fast as possible with the constraint that the inputs and outputs are streamed in and out"?

Yes, that's exactly it, well said! This principle also propagates downstream to all intermediate layers -- the limiting factor on the overall II would be the maximum length of the input/output vectors. In addition, I also have a goal that each layer can run simultaneously (after layer 1 is done, layer 2 then runs in parallel with layer 1, and so on), which implies that we require the dataflow directive at the top layer. 

As I think about it, an alternative "advanced" usage could mix serial with parallel implementations, where say the first layer has a serial input, but then the output of layer 1 is parallelized. Subsquent layers use the io_parallel approach to consume multiple features per clock, while the final layer then re-serializes for clocking out. Assuming we have a streaming IO interface to the IP core, this might represent an optimally efficient solution, but it would definitely add another layer of complexity beyond full serial or full parallel implementations.

> When I used a reuse=2 the latency and II blew up more than expected. And it all comes from the small layer (32x1). I tried digging around to see what was the cause, but I didn't find it yet. Is this the intended behavior?

Good catch. I only tested reuse=2 by manually changing the parameters.h for layer 1, so I did not test on layer 2... Since layer 2 has only one output feature, I'm guessing the multiplier limit rounded down to 0 multipliers (perhaps an integer-math "feature"). Perhaps we need a check that removes the limitation if it would be limited to 0 multipliers? 

> why did you add the pipelines to the activation functions? it looks like that would do nothing

A pipeline directive in the activation for-loop allows the loop to hit II=1. Without pipeline, the for-loop II is equal to the latency -- something like 3 clocks or so, which increases the total II for the activation function to something like 3x the number of features it is consuming. My target II for the overall activation function would be equal to the number of features, with a couple clocks of extra latency. Does that make sense?

> why did you set the IO ports to ap_hs instead of ap_fifo or axis?

On second thought, I would definitely prefer `axis` here. I had selected `ap_hs` to remain "close" to the `ap_vld`, but `axis` would more explicitly demonstrate that this becomes a streaming interface. I would not want to use `ap_fifo` because I believe that instantiates some amount of memory on the interface, which I dont think is really necessary there.

> now i see that weights are stored in BRAMs which is great -- should that be the default and you store the weights in lutrams if specified by user?

Yes, I believe so. It would definitely be appropriate to implement the option to store in LUTRAM instead of BRAM, to help manualy balance BRAM vs fabric usage if necessary (I often found last year that BRAMs quickly became a limiting factor in serial mode). I did not implement those directives yet -- that's a good followup feature to add.

> how does compression get handled in this scheme? it seems like it doesn't naturally take advantage of a sparse weight matrix without some major refactoring of the weights (or removing whole neurons at a time)

I did not consider compression yet for serial mode, but agree it's likely an issue. How feasible is it to prune sparse matrices *before* converting to HLS? If you can identify at the keras level that some weights/multiplies are unnecessary, it would of course help to reduce network size before conversion... HLS likely will not be able to automatically convert BRAMs/ multiplies into logic operations like we saw with the fully parallel implementation.

nhanvtran:
Hi @ejk43, I did a few quick checks and have some updates for this PR:
https://github.com/hls-fpga-machine-learning/hls4ml/compare/ejk/serialized-dense-layer...nvt/serialized-dense-layer

   * changed from `ap_hs` to `axis`
   * made a check for multiplier limit = 0 
   * fixed a pipeline which had an explicit II and I don't think you intended that (copy/paste error?)

with the fixes, now a reuse = 2 gives:

```
+ Latency (clock cycles): 
    * Summary: 
    |   58|   58|   38|   38| dataflow |
```
```
|FIFO             |        0|      -|      15|      89|
|Instance         |       17|     17|    2126|    4412|
|Multiplexer      |        -|      -|       -|       2|
|Register         |        -|      -|       2|       -|
|Total            |       17|     17|    2143|    4507|
|Utilization (%)  |    ~0   |   ~0  |   ~0   |       1|
```

> I did not consider compression yet for serial mode, but agree it's likely an issue. How feasible is it to prune sparse matrices before converting to HLS? If you can identify at the keras level that some weights/multiplies are unnecessary, it would of course help to reduce network size before conversion... HLS likely will not be able to automatically convert BRAMs/ multiplies into logic operations like we saw with the fully parallel implementation.

For the parallel mode, we do all compression outside of Keras reducing weights and sparsifying matrices.  Can you see these [slides](https://indico.cern.ch/event/696998/contributions/2862625/attachments/1585627/2506976/IterativeRetrainingKeras_18Jan2018.pdf)?

Then HLS handles it all naturally just knowing not to use multipliers for the zeros.  I'm wondering what happens in serial mode cases, if resources are actually saved through compression because of the way the multiplications are streamed in.  I think this deserves it's own study, so I'd say that we should just merge this in (with the small updates) and discuss the serial mode compression in another issue/PR.

ejk43:
Awesome! Good fix. I'm happy to see the multiplier limit worked out of the box.

>For the parallel mode, we do all compression outside of Keras reducing weights and sparsifying matrices. Can you see these slides?
>Then HLS handles it all naturally just knowing not to use multipliers for the zeros

I do remember seeing these slides at some point, but I did not notice you were relying on HLS to trim the zero-weight multiplies -- I suppose this makes sense because it would optimally trim a sparse matrix in the fully pipelined case.

Which reminds me... Have you guys looked into sparsifying weights to a power of 2?? I wonder if HLS could squeeze out even more optimizations if the non-zero weights could be implemented as bit-shifts rather than DSPs. Also keep in mind this doesnt mean an "integer" power of 2, but would be driven by the fixed point bit-width of your selected data type. For example, if you used ap_fixed<18,8> (with 10 fractional bits), your smallest weight that could be a power of 2 would be 2^-10,  or you could do 2^-9, and so on... I wonder how many more DSPs could get trimmed from the parallel implementation? 

>I'm wondering what happens in serial mode cases, if resources are actually saved through compression because of the way the multiplications are streamed in. I think this deserves it's own study, so I'd say that we should just merge this in (with the small updates) and discuss the serial mode compression in another issue/PR

I agree about the separate PR -- it needs some study. If you can reduce layer sizes in Keras, that would be a clear win because it both reduces multiplier usage and lowers the input/output streaming requirements... but if it's just a more sparse matrix multiply, its not clear to me how that would get optimized. Any thoughts based on earlier optimizations? 

I manually pulled your updates into my branch so we can go from this PR, also cleaned up whitespace and killed the old PIPEINE. Looks good to merge in to me.

nhanvtran:
ok, I'm merging the PR, looks good!  

I'll open an issue on the the compression in Serial mode.  (including studying quantization to powers of 2).  I don't have any good ideas on how to handle sparse matrices in serial mode.  let's think about this more.  that said, i like the idea you had of mixing parallel and serial layers.  that could be one way around it -- especially if we have fairly large II's.

Issue #53, opened 2018-03-01
Compilation failure when result_t and data_t of compute_layer point to same data type


benjaminkreis:
In [this branch](https://github.com/hls-fpga-machine-learning/hls4ml/tree/muon) I've added a model from Sergo that's failing to compile.  If you do the normal `python keras-to-hls.py -c keras-config.yml` you'll pick up the model and see the error when you try to build the project.

The error is
```
INFO: [SIM 211-2] *************** CSIM start ***************
INFO: [SIM 211-4] CSIM will launch GCC as the compiler.
   Compiling ../../../../myproject_test.cpp in debug mode
   Compiling ../../../../firmware/myproject.cpp in debug mode
In file included from ../../../../firmware/parameters.h:7:0,
                 from ../../../../firmware/myproject.cpp:21:
/home/kreis/muon/hls4ml/nnet_utils/nnet_layer.h: In function ‘void nnet::compute_layer(data_T*, res_T*, typename CONFIG_T::weight_t (*)[CONFIG_T:: n_out], typename CONFIG_T::bias_t*) [with data_T = ap_fixed<18, 8>, res_T = ap_fixed<18, 8>, CONFIG_T = config4, typename CONFIG_T::weight_t = ap_fixed<18, 8>, typename CONFIG_T::bias_t = ap_fixed<18, 8>]’:
../../../../firmware/myproject.cpp:82:77:   instantiated from here
/home/kreis/muon/hls4ml/nnet_utils/nnet_layer.h:100:13: error: invalid use of incomplete type ‘class ap_fixed<18, 8>’
/data/xilinx/Vivado_HLS/2017.2/include/ap_int.h:318:7: error: declaration of ‘class ap_fixed<18, 8>’
/home/kreis/muon/hls4ml/nnet_utils/nnet_layer.h:100: confused by earlier errors, bailing out
make: *** [obj/myproject.o] Error 1
ERROR: [SIM 211-100] 'csim_design' failed: compilation error(s).
INFO: [SIM 211-3] *************** CSIM finish ***************
4
    while executing
"source [lindex $::argv 1] "
    ("uplevel" body line 1)
    invoked from within
"uplevel \#0 { source [lindex $::argv 1] } "

[kreis@rulinux04 my-hls
```

I've definitely seen this one before, but I can't remember the previous causes. 

[...]

PR #54, opened 2018-03-01
linear activation for issue 53

benjaminkreis:
temporary patch for sergo's model to run  #53 

makes a linear activation function that does nothing but gets around the C++ issue


nhanvtran:
Thanks @benjaminkreis ! @sergojin 's model works and resources don't change?

benjaminkreis:
Yes, @sergojin works now.  Use reuse>1 to have it fit in a vertex 7.

PR #55, opened 2018-03-05
fix ceil by casting to float first

benjaminkreis:
This fixes a bug in the multiplier limit calculation that was leading to strange initiation intervals (and possibly more).  The "ceil" in the multiplier limit wasn't doing anything because the objects inside were ints.  Now they are cast to float, the ceil does its thing, and we should have enough multipliers to hit the pipeline target.


nhanvtran:
Looks good!

Issue #56, opened 2018-03-28
Pruning conv1d


benjaminkreis:
I have a [branch](https://github.com/hls-fpga-machine-learning/hls4ml/tree/bjk/prune_conv) going to allow the conv1d to take advantage of pruning in training.  @jmduarte and I have been discussing this, and we think there isn't a simple formula for reducing the multiplier limit based on the number of weights equaling zero.  It gets pretty complicated due to the presence of padding.  

So the approach here is to just do the loops once in advance and count the number of multiplications by nonzero weights, then divide that by the reuse factor to get the multiplier limit.  

Perhaps the simplest would be to do these first loops for counting in separate code (e.g. our python), but I think it would be nice for the nnet_utils to have this feature.  So I've added the count [here](https://github.com/hls-fpga-machine-learning/hls4ml/blob/bjk/prune_conv/nnet_utils/nnet_conv.h#L108).  Unfortunately it looks to me like I haven't managed to fully decouple it from the firmware part.  I see a slightly different resource usage if I use this function versus just passing the limit from the outside world.  If anyone sees the problem, let me know!

Some test results:
Model: KERAS_conv1d_small
Precision: ap_fixed<18,8> 
HLS Version: v2017.2

Default weights:

Reuse | BRAM | DSP | FF | LUT | Lat | II
--- | --- | --- | --- | --- | --- | ---
1 | 13 | 547 | 47149 | 28161 | 24 | 1
2 | 13 | 398 | 47188 | 30769 | 26 | 2
3 | 13 | 266 | 47504 | 33031 | 30 | 3

Random 20% weights set to zero:

Reuse | BRAM | DSP | FF | LUT | Lat | II
--- | --- | --- | --- | --- | --- | ---
1 | 13 | 455 | 41276 | 23950 | 24 | 1
2 | 13 | 303 | 37926 | 24188 | 26 | 2
3 | 13 | 206 | 38074 | 26472 | 30 | 3

Random 50% weights set to zero:

Reuse | BRAM | DSP | FF | LUT | Lat | II
--- | --- | --- | --- | --- | --- | ---
1 | 13 | 279 | 24769 | 15877 | 22 | 1
2 | 13 | 200 | 27442 | 16876 | 25 | 2
3 | 13 | 99 | 17592 | 11563 | 25 | 3

[...]

PR #57, opened 2018-03-30
Multiplier limit for conv1d to take advantage of pruning

benjaminkreis:
Add new function to `nnet_conv.h` called `compute_multiplier_limit` that computes the multiplier limit to take advantage of padding and pruning.

Discussed in #56 

Perhaps the one other thing to note is that zero weights are detected with
```
if( weights[index_weight] > 1e-20 || weights[index_weight] < -1e-20 ){
```
This is different from how we have done it with the fully connected layers.   In that case we check if the weight `==0` at full precision in python [here](https://github.com/hls-fpga-machine-learning/hls4ml/blob/master/keras-to-hls/keras-to-hls.py#L33).  Here we don't use the boolean expression, and we look at the weight at ap_fixed precision.

### DIFF
--- a/nnet_utils/nnet_conv.h
+++ b/nnet_utils/nnet_conv.h
@@ -49,7 +49,7 @@ struct conv_config
 
 
 //Computes number of multiplications required 
-//This function is not synthesized into firmware
+//This function should not be synthesized into firmware
 template<typename CONFIG_T>
 int compute_n_mult(
     typename CONFIG_T::weight_t  weights[CONFIG_T::y_filt * CONFIG_T::n_chan * CONFIG_T::n_filt]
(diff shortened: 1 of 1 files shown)


jmduarte:
Do we want to use `np.isclose()` (need numpy) or `math.isclose()` (maybe only available python 3) instead of this comparison?

https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python

nnet_utils/nnet_conv.h
```diff
@@ -47,6 +47,43 @@ struct conv_config
...
+                    int index_mult   = ii*CONFIG_T::n_filt*CONFIG_T::n_chan*CONFIG_T::y_filt + ff*CONFIG_T::n_chan*CONFIG_T::y_filt + cc*CONFIG_T::y_filt + jj;
+                    int index_weight = jj*CONFIG_T::n_chan*CONFIG_T::n_filt + cc*CONFIG_T::n_filt + ff;
+                    
+                    if((ii*CONFIG_T::stride+jj) < CONFIG_T::pad_left || (ii*CONFIG_T::stride+jj) >= (CONFIG_T::pad_left + CONFIG_T::y_in)){
+			//padded -- do nothing
+			continue;
+                    }
+                    else {
+			//need to tune this cut?
+                        if( weights[index_weight] > 1e-20 || weights[index_weight] < -1e-20 ){
```
jmduarte: Do we want to use `np.isclose()` or `math.isclose()` ?
https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python

benjaminkreis: It looks like isclose is doing the same thing, just with a configurable tolerance.  From a little googling, I don't know of a function for this in C++ (have you found one?).  However, in your link and the examples I saw, the tolerance was more like 1e-8, not 1e-20.  Maybe we should make ours bigger?

jmduarte: this is fine for now

PR #58, opened 2018-04-15
Update Pytorch support

zhenbinwu:
1. Moving print to array from keras to global hls writer
2. Support pytorch with cpu or gpu model


benjaminkreis:
Confirmed that it doesn't change results for keras 1 hidden layer example.  Also ran the default PyTorch example through IP export, and it ran fine.  Going to merge

zhenbinwu:
Thanks for merging! In our next update of the gitbook, we can change the "PyTorch, in alpha testing" to "Pytorch, support Dense network" :-)

PR #60, opened 2018-05-11
conv2d

benjaminkreis:
library and example for 2D convolutional implemented in similar way as 1D 

The example is trained on the famous MNIST handwritten digit dataset, but just using a subset of the image to keep the network small [(see this repo)](https://github.com/benjaminkreis/keras-conv2d/tree/small_for_fpga).  The architecture is 8x8 conv2d with 1 channel and 2 filters -> relu -> dense with 10 outputs -> softmax.  With 32-bit fixed point numbers, the relative agreement with Keras is at the level of 10E-5 for the 10 outputs of the first sample *before* the softmax.  After the softmax, it degrades quite a bit, but this is not specific to conv2d and can be followed up on separately.  

With 18-bit fixed point numbers and reuse=1, it uses 1803 DSPs and 16 200 MHz clocks.

Next steps:
- the multiplier limit is ready for a compressed model, but this hasn't been tested
- the flatten function has been tested, but unflatten has not, which would be useful for architectures with multiple conv2d layers
- serial mode not implemented, similar to conv1d
- max pooling 
- update hls-writer to do the translation


benjaminkreis:
Added conv2d support similar to conv1d support to hls_writer.py

Looking at this a bit closer while updating for Conv2D, I note that not all orderings of Dense, Conv1D, and Conv2D are currently supported.  It expects Dense->Dense, Conv1D->Conv1D, Conv1D->Dense, Conv2D->Conv2D, or Conv2D->Dense.   Also it expects the final layer to be Dense.  I would think these are the typical cases, but we should probably print some errors for the other cases at the very least.

nhanvtran:
@benjaminkreis going through the PR now... in the meantime, what do you want to add to close this PR?  Nothing?  Or at least errors on which Conv1D/2D layers are not supported (we can also do in another PR)?

benjaminkreis:
I think this one is ready to go.  Can add more in future PRs

nhanvtran:
performed unit test, and had a look at the code -- looks good.

PR #61, opened 2018-05-18
BDT

thesps:
Implementation of BDT for HLS4ML.


nhanvtran:
Thanks @thesps  -- this looks really nice.  

Some first comments on the structure/style
- for testing, can you add a testbench code?  I'm working through the code, but all of our projects come with a testbench for debug/testing
- this is a larger coment, in the documentation (I'll work on this) we should add the supported tools and structures including dependencies (here I think it's `sklearn` and `scipy`)
- in the project, again for consistency, I would combine the `parameters.h` and `bdt_config.h` files
- in the `bdt_utils` directory, I see another `parameters.h` -- is this misplaced?  
- for consistency, is it possible to have the BDT call be more something like: 
`bdt_decision_function( x, bdt, score)` or something like that?  This would be consistent with the nnet_utils API

I'm still going through, but I thought I'd get your thoughts on at least these points.  For the "for consistency" points, we can discuss.  There's something to be said for a consistent user interface, but we never decided how strictly to enforce this.

nhanvtran:
Doing a little more digging, the testbench would really help (at least!) me :)...

In the HLS resource usage...

Interestingly, if you do 10 trees, the resource usage is like 4% LUT usage, however it doesn't scale to 100 trees naively -- I get 600% LUT usage.  Somewhere around 50 trees, things start to blow up.  

Also 1 tree has a latency of 2 clocks, so I'm not sure how the latency is mapping to the number of trees.  Again naively I would have thought that everything could have been run in parallel but 100 trees is 20 clocks.  

Anyway, let's discuss here.  Once we're happy with the structure (see above comments) -- I think we could merge that in and then think about how we would handle dataflow and resource usage.

thesps:
@nhanvtran @jngadiub I've added a testbench code that gets generated with the project, hopefully that helps a bit, although it doesn't do very much.
I added support for multi-class BDTs - which really just means more trees and one score per class.
The API changed a little: requiring an array of scores means that the array is passed in the arguments rather than returning a single value, which is consistent with your suggestion above I think?
I tried to make the API similar to the sklearn one, using the same names and structure (having decision_function as a member of the BDT class), but I'm open to changing it.

Regarding evaluating the HLS estimates, it seems odd. Looking at the latency view, I see that some decision functions start on different cycles, which is not expected, but would explain why the latency increases with increases number of trees... I wonder if it's because using a non-power-of-2 number of trees makes the adder tree a bit wonky? Or it relates to the number of adds it can do per cycle: if I change the part to xcku115-flva1517-2-i the split across cycles occurs differently.
Looking at the resource pane of the analysis perspective I see a lot of 'add' opcodes within the decision_function, from cycle 1. The only place they appear is to sum all the scores returned by the Tree::decision_function, and incrementing the loop index. Is it performing the loop index operations in the chip?

bdt_utils/BDT.h
```diff
@@ -0,0 +1,130 @@
...
+		#pragma HLS ARRAY_PARTITION variable=comparison
+		#pragma HLS ARRAY_PARTITION variable=activation
+		#pragma HLS ARRAY_PARTITION variable=activation_leaf
+		#pragma HLS ARRAY_PARTITION variable=value_leaf
+
+		// Execute all comparisons
+		Compare: for(int i = 0; i < n_nodes; i++){
+			#pragma HLS unroll
+			// Only non-leaf nodes do comparisons
+			if(x[feature[i]] != -2){ // -2 means is a leaf (at least for sklearn)
```
nhanvtran: I guess that this should be:
`if (feature[i] != 2)`?

nhanvtran:
@thesps I managed to get the resource usage down to 20% (FFs) and 37% (LUTs)

See below the full table.  Digging through the code, the main improvements (as far as I can tell -- I didn't do a detailed line-by-line comparison) was:
   - simplifying the logic walking through the decision tree instead of walking through each node to determine if it is active
   - partitioning all the tree arrays (thresholds, values, children, parents, etc.)

The result of my code experimentation is on the `nvt/bdt` branch where most everything is here:
https://github.com/hls-fpga-machine-learning/hls4ml/blob/nvt/bdt/bdt_utils/BDT.h#L38-L82

Thinking about this more, because you have to walk through the tree, this is a somewhat serialized operation so the latency should go as `maxdepth`.  Therefore, I think running with 1 tree and getting a latency of 2 clocks is not really realistic -- HLS must have done some super optimization for the single tree case.  It should take several clocks that is somehow proportional to the `maxdepth`.  We should study this with more BDT architectures varying that.  

For the resource usage, I thought about this a bit more too.  I'm not sure I understand how you can get 1% resource usage for a BDT of this size.  Naively, you have to store at least the thresholds and values of each node.  This is 15 numbers per tree for `maxdepth == 3`.  If you set the precision at 18-bits (`ap_fixed<18,8>`), for 100 trees, the FPGA has to store _at least_ 15*2*100*18 bits worth of data, which is about 50,000 FFs.  My best guess is that right now in the example you have the sparse trees that often repeat themselves and Vivado is optimizing that part out.  Unless, of course, I'm missing something silly -- which is very possible.  

I think there are a few more optimization games to experiment with, not sure if it will amount to anything significant...

```
|FIFO             |        -|      -|       -|       -|
|Instance         |        -|      -|  178688|  161653|
|Multiplexer      |        -|      -|       -|      31|
|Register         |        -|      -|       9|       -|
|Total            |        0|      0|  178697|  161688|
|Utilization (%)  |        0|      0|      20|      37|
```

violatingcp:
@thesps Can you add the option of doing a regression? It would nice to see a regression in general. With regressions, we typically use deep trees (as much as depth 30) with a few trees. Its a very different type of approach.

thesps:
n_estimators_scan
max_depth_scan
Here are the plots in place.

nhanvtran:
Thanks a lot @thesps, sorry for the late reply, on holiday last week.

Let me try to break up all this good information.  

1. A naive question: what's the difference between an estimator and a tree?  is the ratio of trees/estimators = 5 fixed?

It's really nice to see these trend plots vs. max depth and n estimators.   

2. About the crazy behavior at ~175 trees, do I understand correctly that this is an HLS synthesis issue because the implementation results still gives a reasonable result?  Maybe something like this thread might help?  I need to understand more.
https://forums.xilinx.com/t5/Vivado-High-Level-Synthesis-HLS/Array-Partition-Completely/td-p/534937

I'm still missing some very basic story other than predicting trends.  We can talk about it tomorrow.  I'd like to revisit this point of just counting the number of registers you need in a single tree for a given precision.  And if trees become sparse then HLS optimizes things out.  I think there is an elegant picture we can paint here still that will help with understanding (maybe I'm just being hopeful).  This way we could from "first principles" estimate the resources of a BDT.  

3. I think there are some fun single tree studies to do with "full" and "sparse" trees.  I'll try to explain this better with some code but let me know if I'm making any sense.

thesps:
# BDT Update
I've recently added a few things to the BDT development. The first of these is `tmva-to-hls`, which creates an hls4ml BDT project from a TMVA xml file. There were a few quirks due to differences between TMVA and scikit-learn, but nothing major. The `sklearn-to-hls` implementation was already reading the model file into an internal format before passing to the hls-writer, so I just had to implement the model-reading for TMVA. This is analogous to the HLSModel, so when it's time for a rebased PR it shouldn't be too horrible.

The second, more major, addition is a VHDL implementation of the BDT. I had been thinking about this for a while as a way to get around the limitations of the HLS implementation. These limitations are documented in the posts above, but the short summary is that the HLS version fails for 'too large' BDTs (too large in the number of trees or depth). 

The VHDL implementation does seem to get around this limitation, as seen in the plots below. These show the LUT usage, latency and synthesis time. The first set is a scan over `n_estimators` with `max_depth = 3`, the second set is a scan over `max_depth` with `n_estimators = 10`. The results from the HLS implementation, already shown above, are overlaid for comparison. 

The most noteworthy aspect is that the VHDL implementation doesn't seem to stumble for 'too large' ensembles, at least with the sizes I tried. In the `n_estimators` scan an ensemble with 1000 estimators (which for 5 classes means 5000 total trees) was no problem, whereas ~175 estimators was already an issue with the HLS version. In the `max_depth` scan, a depth of 10 worked using the VHDL version compared to a maximum of 5 for the HLS. (A depth of 6 compiled with the HLS but the LUTs, latency, and compile time exploded). 

The scaling of LUTs, compile time, and latency is also more reasonable with the VHDL implementation. As a function of `n_estimators`, the LUTs and compile time scale linearly. As a function of `max_depth`, they both scale as `2 ^ depth`. Interestingly, it seems that after the HLS output has been synthesised with Vivado (the step after CSynth) the HLS resources are slightly lower than the VHDL version. The post-synthesis-HLS and synthesised-VHDL resources are very similar though, which is encouraging. 

One nice feature of the HLS which is lost in the VHDL version is the latency (in clock cycles) being dependent on the chosen target clock period. In the VHDL the latency is always fixed, and the design is pipelined aiming to meet high clock frequencies.
The exact latency of a BDT in the VHDL is `1 + depth + ceil(log2(n_estimators))`: 1 for the comparisons, 1 cycle for the 'tree' logic which has as many stages as the depth, and the latency of the 'add-tree' which goes logarithmically with the number of trees. This add-tree is the part which seems to be over-pipelined compared to the HLS, which seems happy to sum 10s of variables in a single cycle, whereas the VHDL only does one pair per cycle, in a tree. 
In the plots below, the latency of the VHDL implementation is generally higher than HLS, although the latency is always very low for both. 

n_estimators_scan

max_depth_scan

# Other Notes
## II
The Initiation Interval is always 1. For now, reuse is not implemented, although I started to put in the infrastructure to add it. From what we've seen with the normal Dense, and Binary Dense, the LUTs don't change much with reuse, so it's not too problematic not to have it, I think. 

## Precision 
The choice of precision in the `config.yml` file still influences the firmware. The user still specifies the type in the HLS manner, eg `ap_fixed<18, 8>` (18 bits, 8 integer). Internally the VHDL implementation only uses `signed` (integer) data types. The width of the type is derived from the type specified in the config file, so this example would become `signed(17 downto 0)`. Any constants, such as the thresholds used in the decision nodes, and the scores output, are then scaled according to the number of fractional bits. In the example `ap_fixed<18, 8>`, there are 10 fractional bits. So every value is multiplied by `2^10` before being written to the firmware.

## Usability
To try to keep the VHDL version as user friendly as possible, and as similar to the regular hls4ml as possible, some scripts and test input vectors are provided when the project is created.
First of all, the use of VHDL is chosen in the config file with `Backend: VHDL` (or the default `Backend: HLS`). 
A script is provided to run behavioral simulation on the design. This uses the VHDL File IO features, reading from a .txt file, and writing to a .txt file. When the project is first generated, an input txt file is created with `0` for every input feature, just like regular hls4ml. The BDT output is written into a txt file on every clock cycle. So the user never needs to open a waveform viewer, and can easily test the numerical performance by reading the output file. For now the script targets Modelsim, but I think adding a script for the Vivado simulator is also needed since we can expect hls4ml users to have Vivado software but not necessarily Modelsim. 

Another script is generated to run synthesis and produce a utilisation report. It is executed like: `vivado -mode batch -source synth.tcl.

## Summary
So, there's a new implementation of the BDT. It seems to solve the major problem with the first implementation, that the size of BDTs that could be handled was quite limited, and the compile times were very long. Obviously there are some new limitations, and maybe some 'philosophical' questions, but I think it makes sense to provide both HLS and VHDL implementations of the BDT in hls4ml, with the `Backend` option in the config file. If others agree, we could then work to update the hls4ml python side of things to the current master.

PR #63, opened 2018-05-30
setting up environment with miniconda for python2

nhanvtran:
helping the user to set up the required dependencies


jmduarte:
I think we can merge now

PR #64, opened 2018-06-06
WIP: Add a conv2d implementation based on Tensorflow tf.nn.conv2d

GiuseppeDiGuglielmo:
See: https://www.tensorflow.org/api_docs/python/tf/nn/conv2d

Computes a 2-D convolution given 3-D input and 4-D filter tensors.

Given an input tensor of shape [in_height, in_width, in_channels] and a filter
tensor of shape [filter_height, filter_width, in_channels, out_channels], this
accelerator performs the following:

- Flattens the filter to a 2-D matrix with shape
  [filter_height * filter_width * in_channels, output_channels].

- Please note that the filter matrix has to be transposed in advance.

- Extracts image patches from the input tensor to form a virtual tensor of
  shape [out_height, out_width, filter_height * filter_width * in_channels].

- For each patch, right-multiplies the filter matrix and the image patch
  vector.


GiuseppeDiGuglielmo:
Let me highlight the differences between [conv2d_tf](https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L142) and [conv2d](https://github.com/hls-fpga-machine-learning/hls4ml/blob/e3cec22ce76af1e5f8b06abcfaab5f69f2cf78c6/nnet_utils/nnet_conv2d.h#L102):

- **conv2d** is a two phases approach that iterates directly over the input features. It multiplies features and weights before passing them to the sum-reduce phase.  **conv2d_tf** has three phases: feature-patch extraction, multiplcation and sum-reduction. Extracting all of the patches in advance may allow to increase the parallelism of the following multiply-accumulate steps. I never formally proved this, so I will appreciate any comment.

- Please note that the input weights are [transposed](https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/example-prjs/conv2d-tf-1layer/firmware/weights/w1.h#L6) in **conv2d_tf**. The reason is that this is my original baseline for a conv2d accelerator that handles bigger feature / weight hypercubes. Such accelerator requires to transfer portions of data from the main memory (DRAM) to the local BRAMs. The access to the main memory through DMA is efficient only if the data is stored contiguously in main memory.

- I just noticed that I do not take into account the bias data, but that can be easily added to the implementation.

Let me know if you have any comments. If you are interested, I can prepare a Latex documentation for the this implementation.

violatingcp:
I am trying to understand the differences with respect to the other conv2d. I see that the latency and pipeline interval get a little worse. For the old conv2d I see: 

+ Latency (clock cycles):
    * Summary:
    |   16|   16|    1|    1| function |

Whereas for this PR conv2d-tf. I see: 
+ Latency (clock cycles):
    * Summary:
    |   18|   18|    6|    6| function |

I can reduce the pipeline to 3 by removing these lines and filling the array directly above: 
https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L127-L131

I still haven't tried this for much larger conv2Ds. I will try that now. Perhaps it would be good to add a much larger conv2d, or at least something where we can see the gain.  

The resource usage is also a little different. Basically more FF and LUTs are used, which I guess has to do with the additional array you are filling.  For completeness here they are: 

conv2d:
* Summary:
|Expression       |        -|      -|       0|       6|
|FIFO             |        -|      -|       -|       -|
|Instance         |       50|   1803|  138788|   75148|
|Multiplexer      |        -|      -|       -|      36|
|Register         |        -|      -|    5959|       -|
|Total            |       50|   1803|  144747|   75190|
|Utilization (%)  |        1|     50|      16|      17|
conv2d-tf
* Summary:
|FIFO             |        -|      -|       -|       -|
|Instance         |       50|   1803|  180087|   87543|
|Multiplexer      |        -|      -|       -|      92|
|Register         |        -|      -|    3648|       -|
|Total            |       50|   1803|  183735|   87639|
|Utilization (%)  |        1|     50|      21|      20|

GiuseppeDiGuglielmo:
@violatingcp can you post the hash of the implementation of *conv2d*  that you are comparing with?

The additional resources may come from the patch buffering:
https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L157-L158
Let me first try to push a little more the QoR and let you know :-) Then we can try to scale it to larger conv2d.

GiuseppeDiGuglielmo:
Few more details about the GEMM algorithm behind the implementation that I uploaded:
https://petewarden.com/2015/04/20/why-gemm-is-at-the-heart-of-deep-learning/
https://arxiv.org/abs/1806.01683

This implementation is to be considered _non-optimized-yet_ but it was the starting point for the development of our accelerator that uses BRAMS and AXI4 channels for data movement.

In first instance, 

- we mapped the following arrays as BRAMS
  https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L158
  https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L89

- we assigned an AXI-master interface for each of the top-module inputs
https://github.com/hls-fpga-machine-learning/hls4ml/blob/d09d8cc5353f3fcc3598fe42372de6d68752413d/nnet_utils/nnet_conv2d_tf.h#L143-L146

- because _potentially_ we cannot fit all of the weights and features into the BRAMs _at the same time_ we bring in portions of the weights that we convolve with portions of the features; you can find a cartoon in the top right corner of this image tha may give an idea
http://www.cs.columbia.edu/~giuseppe/conv2d.jpg

- additional buffers (as BRAMs) are visible on the left of the image

- finally we leverage the wide AXI4 bus on our ZynqMP to move in/out 128 bit per tranfer (4 consecutives 32-bit words).

violatingcp:
Here is a full copy of the logs for the the conv2d and the conv2d-tf: 
https://www.dropbox.com/sh/4kmxds5qrcpf5pi/AABxEC4lMZCNk2USPvjmlgqSa?dl=0

ejk43:
Hey, GDG, fantastic work! Really exciting to see this coming together. Just a few thoughts so far...

> we assigned an AXI-master interface for each of the top-module inputs

I really like this approach! Should remove the BRAM size limitation. It does somewhat change the computing paradigm though, since we have previously been assuming the inference accelerator was implemented as an axi slave, with all weights in either fabric or BRAM to reduce latency and keep things simple. 

Overall this is a really good change towards improving capabilities-- we'll just want to also think about how this translates into the hls-writer and what software we need on the zynq to support weight-writing, since weights may not be built into the image. But, in a big picture view, this opens up a new range of possibilities and I think it's an exciting mode to support as an option.

> finally we leverage the wide AXI4 bus on our ZynqMP to move in/out 128 bit per tranfer (4 consecutives 32-bit words)

What do you think about supporting both 64 and 128 bit widths, for those of us currently still stuck on gen 1 zynqs?? :)

PR #66, opened 2018-06-25
Making the LUT table static

zhenbinwu:
For the activation function, the initiation of LUT is taking too much CPU time
during csim. To fix this issue, change the table to be a static varibles, which will
remain in memory for the running lifetime. Also only call the initiation function by
checking the table value, making the initiation function got called only once.


nhanvtran:
@zhenbinwu Awesome!

PR #67, opened 2018-06-27
Use boolean instead of value for checking initialization

zhenbinwu:
Notice the -999 value could be cast differently once the table type
changed, like width/iteger value of ap_fixed. Use boolean instead for
safety

### DIFF
--- a/nnet_utils/nnet_activation.h
+++ b/nnet_utils/nnet_activation.h
@@ -137,9 +137,15 @@ template<class data_T, class res_T, typename CONFIG_T>
 void  sigmoid(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 {
     // Initialize the lookup table
-    static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size] ={-999.};
-    if(sigmoid_table[0] == -999.)
+    static bool initialized = false;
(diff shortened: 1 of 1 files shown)


nnet_utils/nnet_activation.h
```diff
@@ -137,9 +137,15 @@ template<class data_T, class res_T, typename CONFIG_T>
...
-    if(sigmoid_table[0] == -999.)
+    static bool initialized = false;
+    static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
+    if (!initialized)
+    {
       init_sigmoid_table<CONFIG_T, CONFIG_T::table_size>(sigmoid_table);
+      initialized = true;
+    }
+
+    if(sigmoid_table[0] == CONFIG_T::table_dummy)
```
vloncar: What is the purpose of this line?

zhenbinwu: Ops, crap, this line should be removed. Let me fix it! Thanks

nhanvtran:
hi @zhenbinwu @vloncar thanks for adding this optimization and the validation!  from the review by @vloncar it looks ready to merge, do you agree?

vloncar:
Looks good to me. Once you merge it, I will rebase my code and (finally) make the PR with the additional activation functions.

nhanvtran:
ok thanks @vloncar ! merging...

PR #68, opened 2018-06-29
Ap fixed omode

zhenbinwu:
The default ap_fixed has the overflow mode as AP_WRAP, which will
cause the number to change significantly from original input.

For example, with the initial value of 8.99, the default ap_fixed<10,4> will
cast it to  -7.01563, while the new ap_fixed<10, 4, AP_TRN_ZERO, AP_SAT> will
cast it to 7.98438.

The change of ap_fixed type only apply to the node output of each layer for now,
to prevent overflow during the run time. The weights and bias are still stored as
default input type. The resource usage remain the same from this type change.

### DIFF
--- a/hls-writer/hls_writer.py
+++ b/hls-writer/hls_writer.py
@@ -314,7 +314,7 @@ def hls_writer(layer_list, yamlConfig):
             for i in range(1,len(layer_list)):
             #    if layer_list[i-1]['class_name']=='Dense':
             #        newline += 'typedef {precision} layer{index}_t;\n'.format(precision=yamlConfig["DefaultPrecision"], index=i)
-                newline += 'typedef {precision} layer{index}_t;\n'.format(precision=yamlConfig["DefaultPrecision"], index=i)
+                newline += 'typedef {precision} layer{index}_t;\n'.format(precision=yamlConfig["DefaultPrecision"][:-1]+", AP_TRN_ZERO, AP_SAT>", index=i)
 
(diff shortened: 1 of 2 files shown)


nhanvtran:
hi @zhenbinwu can we also add these ap_fixed options to the weights, accumulators, bias, inputs and results?

https://github.com/hls-fpga-machine-learning/hls4ml/blob/ApFixedOmode/hls-writer/hls_writer.py#L284-L288

zhenbinwu:
Since I didn't see any change in the resource, I think it should be fine. Let me run a test. For the weights/bias, I think the users should be careful to choose the right width.

nhanvtran:
Ok so the rationale is just to add this option for the layer outputs but not the weights/biases?  in this case I would also add the option to the inputs and results, no?

zhenbinwu:
Typically, the overflow happens within the layer where additions apply. That is why I started with only the layer values.
screen shot 2018-07-10 at 10 01 06 am

We could propagate the overflow mode to all of the ap_fixed type. That could be easier, we just need to modify the ap_fixed to ap_fixed<x, x, tmode, omode> in the yml file.

nhanvtran:
yeah let's do that, it's cleaner, just add a comment to the yml to say what those options mean, thanks!

zhenbinwu:
Yeah, agree. Let us close this PR then. I will make a new PR

nhanvtran:
👍

PR #72, opened 2018-07-05
batch normalization layer

jngadiub:
Start here with adding the class for computing batch normalization given some trained parameters as implemented here:

https://github.com/keras-team/keras/blob/master/keras/backend/tensorflow_backend.py#L1857

The gamma / sqrt(var + epsilon) factor is loaded directly as one value called "scale".

The hls-writer still needs to be updated to support this.


nhanvtran:
thanks @jngadiub! to formalize this, can we add an example model that uses this and can be parsed by the keras hls-writer?

jngadiub:
@nhanvtran I updated the hls-writer and keras-to-hls and also added an example model: 

KERAS_3layer_batch_norm.json
KERAS_3layer_batch_norm_weights.h5

However, I am having a problems in compiling when all layer_t types are declared as ap_fixed. In particular, the softmax seems to give the problem as below. While if I declare everything as float the problem does not show up. I am investigating.

csim.mk:79: recipe for target 'obj/myproject.o' failed
In file included from ../../../../firmware/parameters.h:11:0,
                 from ../../../../firmware/myproject.cpp:21:
/afs/cern.ch/work/j/jngadiub/fpga-ml/hls4ml/nnet_utils/nnet_activation.h: In function 'void nnet::softmax(data_T*, res_T*) [with data_T = ap_fixed<18, 8>, res_T = ap_fixed<18, 8>, CONFIG_T = softmax_config12]':
/afs/cern.ch/work/j/jngadiub/fpga-ml/hls4ml/nnet_utils/nnet_activation.h:237:14: internal compiler error: Segmentation fault
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://gcc.gnu.org/bugs.html> for instructions.
make: *** [obj/myproject.o] Error 1
4
    while executing
"source [lindex $::argv 1] "
    ("uplevel" body line 1)
    invoked from within

zhenbinwu:
What is in line 237 in your nnet_activation.h? From the master branch, it seems just assignment: https://github.com/hls-fpga-machine-learning/hls4ml/blob/master/nnet_utils/nnet_activation.h#L237

jngadiub:
But mine does not seem different:

https://github.com/jngadiub/hls4ml/blob/master/nnet_utils/nnet_activation.h#L237

or what do you mean?

It does not give problems for the default 3 layer model. Only after I added the batch normalization + activation layers.

zhenbinwu:
```/afs/cern.ch/work/j/jngadiub/fpga-ml/hls4ml/nnet_utils/nnet_activation.h:237:14: internal compiler error: ```

So it pointed to this line of code. Not sure why it could cause Segmentation fault.

**Comment by jngadiub** [comment]
yes, right!

at some point @benjaminkreis was telling me that he noticed there are problems when having too many typedefs. not sure if this is the case and how to go over it.

**Comment by benjaminkreis** [comment]
The error @jngadiub is referring to had different messages.  See https://github.com/hls-fpga-machine-learning/hls4ml/issues/53

It looks like @jmduarte mentioned a seg fault like this once: https://github.com/hls-fpga-machine-learning/HLS4ML/pull/40.  It's not 100% clear to me from this thread what fixed it.  Maybe reducing the number of dimensions of the array?  But Line 237 only has 2D arrays, which we haven't had problems with.

is it really due to line 237?  what happens if you comment out that line and make the `else` always run? (is that `if` really needed?)

**Comment by jngadiub** [comment]
@benjaminkreis if it would be an issue with the dimension of the array, it would show up for the default 3 layer model too, and this does not happen. I tried to comment line 237, but then the error simply moves down to line 243

https://github.com/jngadiub/hls4ml/blob/master/nnet_utils/nnet_activation.h#L243

I keep digging into it.

**Comment by benjaminkreis** [comment]
ah right, I had assumed this line was in the batch norm layer!

**Comment by vloncar** [comment]
We looked into this, and I am pretty certain this is a GCC bug, since it works with clang compiler. GCC that comes with Vivado HLS (GCC 4.6.3 in 2017.2) seems to have a problem with the type of `exp_diff_res`, and screws up somehow. Declaring `exp_diff_res` with `data_T` (which we shouldn't do but for the sake of testing) allows it to work, as does using `ap_fixed<18,8>` directly. Also, as Jen found out, changing the call of softmax to
```
nnet::softmax<result_t, result_t, softmax_config12>(layer11_out, res)
```
makes it work.

BTW, why is `exp_diff_res` a 2d array? Its sole purpose is to populate exp_res array in a loop, and is not used afterwards. Strangely enough, changing it to be a scalar allows GCC to compile it, the code synthetises, and has the same resource usage/latency as when it is defined as 2d array. In other words, HLS seems to know what to do when parallelizing the loop. I was thinking maybe it is related to serial code, however none of the models with softmax synthetize when configured as serial (by configured I mean run `keras-to-hls` with `io_serial` in .yml file). I will open up a separate issue for that.

**Comment by benjaminkreis** [comment]
Indeed it also looks to me like exp_diff_res could just be a scalar.

Is the plan to just avoid the compiler bug by doing 
```
nnet::softmax<result_t, result_t, softmax_config12>(layer11_out, res)
```
?

Is this in the PR?

**Comment by jngadiub** [comment]
@benjaminkreis yes, that's the plan. PR on its way. I am still doing few tests on the final output values and on the results of the synthesis.

**Comment by vloncar** [comment]
Do we know why is `exp_diff_res` a 2d array? According to [git blame](https://github.com/hls-fpga-machine-learning/hls4ml/blame/master/nnet_utils/nnet_activation.h#L235), @jmduarte can shed some light?

I feel that it would be a simpler solution to declare `exp_diff_res` as scalar and avoid the problem that way. Changing `keras-to-hls.py` and/or `hls_writer.py` to work around the problem every time softmax is used is more work.

**Comment by jngadiub** [comment]
@vloncar I think the change when loading the softmax has to be added anyways so that it is used consistently through all the models considered in the hls-writer. Furthermore, there is no reason why the last layer should have a different type name than result_t. I already implemented and I'll push it. I am adding the change to the exp_res_diff too.

**Comment by jngadiub** [comment]
The main problem I see at the moment is that the output values do not fully match with what I expect from keras, even when using floating point instead of ap_fixed. I would like to understand that before making a PR.

**Comment by jngadiub** [comment]
Updated the PR with the fixes to avoid the crash in the csim as discussed above.

The model architecture has been added to the keras-training repo in this PR https://github.com/hls-fpga-machine-learning/keras-training/pull/10

and the corresponding json and h5 file for hls4ml are currently in the example folder for testing
https://github.com/hls-fpga-machine-learning/hls4ml/pull/72/files#diff-210c7bd45c0f64016b0ec1464d91b085R1

I tested the following array of features taken from the trainining dataset:

```
1.20769 -1.01877 -1.1673 -0.844171 -1.06598 -0.623767 -0.0862555 0.0107256 -0.0862555 0.456239 0.633049 0.975798 0.456674 0.683313 -1.21286 -0.843307
```

The prefiction from keras for the features above is

```
0.112673 0.635937 0.124081 0.12241 0.00489956
```

while from the csim using floating point is

```
0.114286 0.64 0.125 0.124031 0.0156403
```

where residual differences are very likely due to approximations we are doing in the softmax (I tested the full model without softmax at the end and I get same values as keras so the BN calculations should be correct)

Bad news when looking at the results of the synthesis, as the latency for the compute_layer increased from 1 to 2 when adding the batch normalization layers:

**Latency with BN layers (total latency = 34)**


screen shot 2018-08-07 at 22 00 17

**Latency w/o BN layers (total latency = 22)**

screen shot 2018-08-07 at 22 27 52

**Comment by benjaminkreis** [comment]
```Bad news when looking at the results of the synthesis, as the latency for the compute_layer increased from 1 to 2 when adding the batch normalization layers:```

I'm confused by this statement.  In your tables it looks like compute_layer always has latency=2.  Can you please clarify?

File: keras-to-hls/keras-to-hls.py [hunk 1]
```diff
@@ -111,17 +111,31 @@ def main():
...
+        if layer['class_name'] != 'BatchNormalization' and layer['class_name'] != 'Activation':
+         found_weights = h5File[layer['name']].visit(find_kernel_in_h5)
+         weights = h5File['/{}/{}'.format(layer['name'],found_weights)][()]
+         found_bias = h5File[layer['name']].visit(find_bias_in_h5)
+         biases = h5File['/{}/{}'.format(layer['name'],found_bias)][()]
+         cur_n_zeros = print_array_to_cpp("w{}".format(layer_counter), weights, yamlConfig['OutputDir'])
+         print_array_to_cpp("b{}".format(layer_counter), biases, yamlConfig['OutputDir'])
+         layer['weights_n_zeros'] = cur_n_zeros 
+        elif layer['class_name'] == 'BatchNormalization':
+         beta = h5File['/{}/{}/beta:0'.format(layer['name'],layer['name'])][()]
```
benjaminkreis: For extracting beta and gamma, can you update this to the approach of the weights and bias above?  I found that this code does not work for this example `/afs/cern.ch/user/k/kreis/public/hls4ml/walter.tgz`

nnet_utils/nnet_batchnorm.h
```diff
@@ -0,0 +1,139 @@
...
+{
+    // Internal data type definitions
+    typedef float beta_t;
+    typedef float scale_t;
+    typedef float mean_t;
+    typedef float accum_t;
+
+    // Layer Sizes
+    static const unsigned n_in = 10;
+    static const unsigned n_out = 10;
```
benjaminkreis: Do we need an n_out parameter?  Are the number of outputs ever different from the number of inputs?

hls-writer/hls_writer.py
```diff
@@ -35,8 +36,14 @@ def hls_writer(layer_list, yamlConfig):
         elif '//hls-fpga-machine-learning insert weights' in line:
             newline = line
             for i in range(1,len(layer_list)+1):
-                newline += '#include "weights/w{}.h"\n'.format(i)
-                newline += '#include "weights/b{}.h"\n'.format(i)
+                if layer_list[i-1]['class_name'] != 'BatchNormalization' and layer_list[i-1]['class_name'] != 'Activation':
```
benjaminkreis: I see Activation several places below.  What is this for?  I think we haven't come across this before.

hls-writer/hls_writer.py
```diff
@@ -225,22 +237,42 @@ def hls_writer(layer_list, yamlConfig):
...
+                    newline += '    nnet::sigmoid<{}, {}, {}>({}, {});\n'.format(input_type, output_type, activation_name, input_object, output_object)
+                 elif layer_list[i-1]['activation'] =="tanh":
+                    newline += '    nnet::tanh<{}, {}, {}>({}, {});\n'.format(input_type, output_type, activation_name, input_object, output_object)
+                 elif layer_list[i-1]['activation'] == "binary_tanh":	
+                    newline += '    nnet::binary_tanh<{}, {}, {}>({}, {});\n'.format(input_type, output_type, activation_name, input_object, output_object)
+                 elif layer_list[i-1]['activation'] =="linear": 
                     #github Issue 53
-                    newline += '    nnet::linear<{}, {}, {}>(logits{}, {});\n'.format(output_type, output_type, activation_name, i, output_object)
-                else:
+                    newline += '    nnet::linear<{}, {}, {}>(logits{}, {});\n'.format(input_type, output_type, activation_name, input_object, output_object)
```
benjaminkreis: Is the fact that this `linear` looks different from the one above related to what Activation is?

nnet_utils/nnet_batchnorm.h
```diff
@@ -0,0 +1,139 @@
...
+        #pragma HLS ARRAY_PARTITION variable=shift complete
+
+        int multiplier_limit  = ceil(float(CONFIG_T::n_in*CONFIG_T::n_out) / float(CONFIG_T::reuse_factor)) - floor(float(CONFIG_T::n_zeros) / float(CONFIG_T::reuse_factor));
+        #pragma HLS ALLOCATION instances=mul limit=multiplier_limit operation
+
+    } else if (CONFIG_T::io_type == io_serial){
+        #pragma HLS ARRAY_RESHAPE variable=scale complete dim=2
+        #pragma HLS ARRAY_RESHAPE variable=beta complete dim=2
+        #pragma HLS ARRAY_RESHAPE variable=mean complete dim=2
+        #pragma HLS ARRAY_PARTITION variable=mult complete dim=2
```
benjaminkreis: are these array_partition settings tested or based on something (ie compute_layer)?

nnet_utils/nnet_batchnorm.h
```diff
@@ -0,0 +1,139 @@
...
+        
+    // Do the multiplication
+    Product: for(int i = 0; i < CONFIG_T::n_in; i++) {
+        if (CONFIG_T::io_type == io_serial){
+            #pragma HLS PIPELINE
+        }
+        mult[i] = scale[i] * shift[i];
+    }    
+    
+    // Initialize accumulator with beta values
```
benjaminkreis: This code looks good, but i wouldn't really call it an accumulator because each entry is just taking one multiplication result.

Do you get different resource usage if you collapse all of these loops into just one loop with `output = (data-mean)*scale + beta`?

jngadiub:
@benjaminkreis regarding the latency you are right. I got confused because before merging the changes that have been pushed for the compute_layer and sublayer, I was getting a latency of 1 clock cycle for the compute_layer function as I showed in slide 11 in this past presentation

https://indico.cern.ch/event/742016/contributions/3066083/attachments/1681759/2702147/binary-dense-hls4ml-5jul18.pdf

We now get a total latency of 22 clock cycles while in the paper we had 15 (as in the slide). Have you observed this? The latency for the softmax also changed (from 4 to 7).

nnet_utils/nnet_batchnorm.h
jngadiub: I changed to calculate the result in one line as you suggested. It does not seem to affect latency nor resources.

benjaminkreis:
@jngadiub what version of Vivado and which Xilinx part number are you using?

With 2017.2 and xcku115-flvb2104-2-i, I get this for the `KERAS_3layer_batch_norm` model :
```
+ Latency (clock cycles): 
    * Summary: 
    |   23|   23|    1|    1| function |

    + Detail: 
        * Instance: 
        +----------------------------------+-----------------------+-----+-----+-----+-----+----------+
        |                                  |                       |  Latency  |  Interval | Pipeline |
        |             Instance             |         Module        | min | max | min | max |   Type   |
        +----------------------------------+-----------------------+-----+-----+-----+-----+----------+
        |grp_compute_layer_0_0_0_1_fu_103  |compute_layer_0_0_0_1  |    1|    1|    1|    1| function |
        |grp_compute_layer_0_0_0_s_fu_171  |compute_layer_0_0_0_s  |    1|    1|    1|    1| function |
        |grp_compute_layer_0_0_0_2_fu_207  |compute_layer_0_0_0_2  |    1|    1|    1|    1| function |
        |grp_normalize_0_0_0_0_0_2_fu_213  |normalize_0_0_0_0_0_2  |    1|    1|    1|    1| function |
        |grp_compute_layer_0_0_0_3_fu_281  |compute_layer_0_0_0_3  |    1|    1|    1|    1| function |
        |grp_softmax_fu_317                |softmax                |    4|    4|    1|    1| function |
        |grp_normalize_0_0_0_0_0_1_fu_330  |normalize_0_0_0_0_0_1  |    1|    1|    1|    1| function |
        |grp_normalize_0_0_0_0_0_s_fu_366  |normalize_0_0_0_0_0_s  |    1|    1|    1|    1| function |
        |call_ret3_relu_2_fu_402           |relu_2                 |    0|    0|    1|    1| function |
        |grp_normalize_0_0_0_0_0_3_fu_470  |normalize_0_0_0_0_0_3  |    1|    1|    1|    1| function |
        |call_ret7_relu_1_fu_479           |relu_1                 |    0|    0|    1|    1| function |
        |call_ret11_relu_fu_515            |relu                   |    0|    0|    1|    1| function |
        |call_ret1_linear_2_fu_551         |linear_2               |    0|    0|    1|    1| function |
        |call_ret5_linear_1_fu_619         |linear_1               |    0|    0|    1|    1| function |
        |call_ret9_linear_fu_655           |linear                 |    0|    0|    1|    1| function |
        |call_ret13_linear_3_fu_691        |linear_3               |    0|    0|    1|    1| function |
        +----------------------------------+-----------------------+-----+-----+-----+-----+----------+
```

I've seen softmax take 7 clocks with xc7vx690tffg1927-2.  Are you comparing the "paper" dense network on xc7vx690tffg1927-2 with your "KERAS_3layer_batch_norm" network on xcku115-flvb2104-2-i?

nnet_utils/nnet_batchnorm.h
```diff
@@ -0,0 +1,94 @@
...
+        //   - completely partition arrays -- target fabric
+        //   - if we have an unroll factor, limit number of multipliers
+        #pragma HLS PIPELINE II=CONFIG_T::reuse_factor
+
+        // #pragma HLS ARRAY_PARTITION variable=weights complete // remove this line for now, it breaks compression sometimes
+	#pragma HLS ARRAY_PARTITION variable=scale complete
+        #pragma HLS ARRAY_PARTITION variable=beta complete
+	#pragma HLS ARRAY_PARTITION variable=mean complete
+
+        int multiplier_limit  = ceil(float(CONFIG_T::n_in*CONFIG_T::n_in) / float(CONFIG_T::reuse_factor)) - floor(float(CONFIG_T::n_zeros) / float(CONFIG_T::reuse_factor));
```
benjaminkreis: I think we want to change this to 
`ceil(float(CONFIG_T::n_in) / float(CONFIG_T::reuse_factor))`

(assuming we don't do any kind of pruning of batch norm parameters)

do you agree?

jngadiub:
@benjaminkreis you are right, the resources that I posted above are for the xc7vx690tffg1927-2, while for the xcku115-flvb2104-2-i I get same as you and consistent with the paper. So each normalization layer adds one clock cycle and DSPs.

nnet_utils/nnet_batchnorm.h
jngadiub: I agree and changed as you suggested.

jngadiub:
I updated the PR to support different configurations of fully-connected models with batch normalization layers:

1) the batch normalization is performed after activation (so no need for the Activation layers)
https://github.com/jngadiub/keras-training/blob/master/models/models.py#L105

2) batch normalization is the first or last layer
https://github.com/jngadiub/keras-training/blob/master/models/models.py#L126

jngadiub:
@benjaminkreis I have to correct my statement above. So for a model like this one 
https://github.com/jngadiub/keras-training/blob/master/models/models.py#L77
where the batch normalization is performed before activation and one needs the Activation layers, the increase in total latency is of 8 clock cycles (15 --> 23), although the batch normalization layer only takes 1 clock cycle as the dense layer.

If one performs batch normalization before the activation as in this model
https://github.com/jngadiub/keras-training/blob/master/models/models.py#L105
where one does not need the additional Activation layers, the total latency increases of only 3 clock cycles as expected considering the addition of 3 batch normalization layers.

Not sure if and how we should handle the two cases differently.

I just added the support for Conv2D with batch normalization as in the model from @pierinim where he performs batch normalization before activation. Other configurations have to be added though.

benjaminkreis:
@jngadiub and I looked into the latency issue she described immediately above, where Dense->Linear->Batch Norm->Relu takes longer than Dense->Relu->Batch Norm.  In the resource report, the latency of Linear, Batch Norm, and Relu all have Latency=0.  We find in the Analysis view of the GUI that the latency difference comes from where HLS chooses to place clocks between these blocks as it connects them together.  It doesn't seem to be a problem with our code -- the individual pieces have the latency we expect -- and just seems to be something about the HLS algorithm.  We can keep it in mind as a curiosity perhaps to be followed up if it matters for someone, but I don't think we need to delay merging this.

I've run this PR on a tiny conv2d model I trained on MNIST and find that we can csynth it and export it without problems.  I also checked the csim output for a couple of samples with very high precision fixed point and a nonsensical final activation (relu) just to avoid the reduced precision of the lut-based activations:

```
keras:  0.10620116 0. 0. 2.0469031  0. 0.08349947 0.  0.  0.8883401  0.17893735
hls4ml: 0.106202   0  0  2.0469     0 0.0834994   0   0   0.88834    0.178937

keras:  4.9425244 0. 0. 0.16815907 0. 1.6537665 0. 0.7933433 0. 0.
hls4ml: 4.94252   0  0  0.168159   0  1.65377   0  0.793343  0  0 
```
It looks good!

will go ahead and merge :+1:

PR #78, opened 2018-07-17
Use C++ math functions

vloncar:
This solves the synthesis process hanging when we use static tables.

### DIFF
--- a/nnet_utils/nnet_activation.h
+++ b/nnet_utils/nnet_activation.h
@@ -20,7 +20,7 @@
 #ifndef NNET_ACTIVATION_H_
 #define NNET_ACTIVATION_H_
 
-#include <math.h>
+#include <cmath>
 #include "ap_fixed.h"
 #include "nnet_common.h"
 
@@ -115,7 +115,7 @@ void  relu6(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 //       Sigmoid Activation
 // *************************************************
 inline float sigmoid_fcn_float(float input) {
-    return 1.0 / (1 + exp(-input));
+    return 1.0 / (1 + std::exp(-input));
 }
 
 template<typename CONFIG_T, int N_TABLE>
@@ -168,7 +168,7 @@ void  sigmoid(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 //       Softmax Activation
 // *************************************************
 inline float exp_fcn_float(float input) {
-    return exp(input);
+    return std::exp(input);
 }
 
 
@@ -268,7 +268,7 @@ void init_tanh_table(typename CONFIG_T::table_t table_out[N_TABLE])
         // First, convert from table index to X-value (signed 8-bit, range -4 to +4)
         float in_val = 2*4.0*(ii-float(N_TABLE)/2.0)/float(N_TABLE);
         // Next, compute lookup table function
-        typename CONFIG_T::table_t real_val = tanh(in_val);
+        typename CONFIG_T::table_t real_val = std::tanh(in_val);
         //std::cout << "Tanh:  Lookup table Index: " <<  ii<< " In Value: " << in_val << " Result: " << real_val << std::endl;
         table_out[ii] = real_val;
     }


benjaminkreis:
Unfortunately this doesn't look like a fix.  I'm getting bad latency and interval results for the sigmoid:

```
        +---------------------------------+-----------------------+-----+-----+-----+-----+----------+
        |                                 |                       |  Latency  |  Interval | Pipeline |
        |             Instance            |         Module        | min | max | min | max |   Type   |
        +---------------------------------+-----------------------+-----+-----+-----+-----+----------+
        |grp_compute_layer_0_0_0_s_fu_79  |compute_layer_0_0_0_s  |    2|    2|    1|    1| function |
        |grp_sigmoid_fu_85                |sigmoid                |    4|  515|    4|  515|   none   |
        |grp_compute_layer_0_0_fu_94      |compute_layer_0_0      |    2|    2|    1|    1| function |
        |call_ret1_relu_fu_131            |relu                   |    0|    0|    1|    1| function |
        +---------------------------------+-----------------------+-----+-----+-----+-----+----------+

```

vloncar:
I get the same numbers since the introduction of static tables. If I remove the static qualifier but keep the cmath changes the latency goes back to ~10

benjaminkreis:
Ok I think we need to take a closer look at that static PR

nhanvtran:
Thanks all, ok can we roll back on the static PR, @zhenbinwu @vloncar?  My fault for merging it too soon.  We're working on some unit tests as the project grows...

zhenbinwu:
The static problem and the csim issue can be fixed like below:
```
#ifdef __HLS_SYN__
    bool initialized = false;
    typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
#else
    static bool initialized = false;
    static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
#endif
```

No my favorite style, but it solved both issues. @vloncar @benjaminkreis Can you double check?

vloncar:
@zhenbinwu I am not sure how this would work. If the table is not static during synthesis (`#ifdef __HLS_SYN__`), the code essentially behaves in the same way as before the change was introduced, right? It would simply be reinitialized every time. Or am I missing something?

zhenbinwu:
@vloncar There was an issue with csim in vivado_hls v2018.2. The initiation of the LUT takes a very long time. We introduced the static variables to avoid multiple initiation. Now I just split out the csim and syn steps in the code. BTW, can you check the csim speed after you change to std::exp()?  Maybe it got fixed?

vloncar:
@zhenbinwu The speed is the same, as C++ compiler essentially does the same thing and only a workaround for HLS. There is another option, using `hls_math.h` and using its functions (e.g., `hls::exp()`), but this significantly increases the compilation time of csim (like 4-5 minutes), so I didn't use it. It also doesn't solve the problem with latency.

benjaminkreis:
Regarding the HLS Math Library -- this would be to evaluate the exp() in real time with FPGA logic rather than by precomputing a lookup table.  I think we want to stick with the lookup table approach.

zhenbinwu:
What do you mean the speed is the same? With std::exp(), the original code csim with 1layer is still slow.

vloncar:
There should be no difference between `std:exp()` from `cmath` and `exp()` from `math.h`. The latter one exists for the compatibility with C, and the former is the C++ way, but the underlying math is the same. My guess is that HLS synthesis gets confused somehow, it sees `exp()`, which accepts `double`, yet we give it `float`, does the type promotion, blows up memory usage, and hangs. 

Since this PR doesn't fix the problem with latency, I would suggest we close it without merging.

nhanvtran:
hi all,  ok, i tried to get a summary from @benjaminkreis and @zhenbinwu 

In short, adding the static blows up the csynth time and the resource usage, but without the static, the csim time is super slow.  So we will close this PR and @zhenbinwu is going to make a new PR which has the solution along the lines of what he posted above:

```
#ifdef __HLS_SYN__
    bool initialized = false;
    typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
#else
    static bool initialized = false;
    static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
#endif
```

PR #81, opened 2018-07-23
Condition the LUT initialtion

zhenbinwu:
This is to fix the slow csim for Vivado_HLS >2017.4. Also try to avoid
the slow syn issue with the introduction of static variables.

Tested with 1Layer DNN.  The current master code won't work with 3Layer. But test with 3Layer
in older version of the code seems OK

### DIFF
--- a/nnet_utils/nnet_activation.h
+++ b/nnet_utils/nnet_activation.h
@@ -20,7 +20,7 @@
 #ifndef NNET_ACTIVATION_H_
 #define NNET_ACTIVATION_H_
 
-#include <math.h>
+#include <cmath>
 #include "ap_fixed.h"
 #include "nnet_common.h"
 
@@ -115,7 +115,7 @@ void  relu6(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 //       Sigmoid Activation
 // *************************************************
 inline float sigmoid_fcn_float(float input) {
-    return 1.0 / (1 + exp(-input));
+    return 1.0 / (1 + std::exp(-input));
 }
 
 template<typename CONFIG_T, int N_TABLE>
@@ -137,8 +137,13 @@ template<class data_T, class res_T, typename CONFIG_T>
 void  sigmoid(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
 {
     // Initialize the lookup table
+#ifdef __HLS_SYN__
+    bool initialized = false;
+    typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
+#else
     static bool initialized = false;
     static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
+#endif
     if (!initialized)
     {
       init_sigmoid_table<CONFIG_T, CONFIG_T::table_size>(sigmoid_table);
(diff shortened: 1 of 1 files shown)


nhanvtran:
Thanks @zhenbinwu, I'm going to request that @vloncar reviews this PR.  While we wait for the 3layer pruned to get sorted out, we can test the others.  @vloncar if it checks out, then we can merge with PR #69

benjaminkreis:
Why doesn't the 3 layer work?

vloncar:
If `bool initialized = false` is set without a `static` modifier, doesn't that mean that every time a function is called, `initialized` will be set to `false`, essentially making `initialized` redundant and behaving the same as before, but now only during synthesis? For example, is this an equivalent code, just without defining `initialized` at all during synthesis? I find it to be slightly less confusing about the intended behavior.
```
#ifdef __HLS_SYN__
    typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
    init_sigmoid_table<CONFIG_T, CONFIG_T::table_size>(sigmoid_table);
#else
    static bool initialized = false;
    static typename CONFIG_T::table_t sigmoid_table[CONFIG_T::table_size];
    if (!initialized {
      init_sigmoid_table<CONFIG_T, CONFIG_T::table_size>(sigmoid_table);
      initialized = true;
    }
#endif
```
Also, where is `__HLS_SYN__` macro documented? I know of `__SYNTHESIS__` macro which is mentioned in UG902, which I guess is the same thing, however the documentation only suggests using `__SYNTHESIS__`. `__HLS_SYN__` seems to be only be used in HLS header files and is intended for internal use.

zhenbinwu:
@benjaminkreis The master branch of the hls4ml is having problem with issue. See issue 79

@vloncar Ha, you are right. Meant to copy the older version. Let me fix them. Thanks!

nhanvtran:
@zhenbin actually, can you test the 3 layer in 2017.2 or 2017.4 as long as the macro __HLS_SYN__ is available?

vloncar:
@zhenbinwu And `__SYNTHESIS__`? I guess it is just a matter of style, since it is unlikely that Xilinx will change that macro, so if you don't want to do it I won't be nitpicking anymore :-)

zhenbinwu:
@vloncar The __HLS_SYN__ is used in the ap_fixed.h source code. 
```
#ifdef __HLS_SYN__
#error ap_fixed simulation header file is not applicable for synthesis
#else

#include "ap_int.h"

// Specialization of std containers, so that ap_fixed behave similarly to
// float or double.
#include "ap_fixed_special.h"

#endif
```

Is the __SYNTHESIS__ recommended? From https://www.xilinx.com/support/answers/54160.html, it suggested not to use?

vloncar:
@zhenbinwu UG902 only mentions the existence of `__SYNTHESIS__`, and not `__HLS_SYN__`, so we don't really know what `__HLS_SYN__` is doing (but I believe that they are the same thing). I don't think we mask top level ports with this code, so the suggestion from the link probably doesn't apply.
Anyway, since you have already tested with `__HLS_SYN__` and everything behaves correctly, I suggest we merge it without further changes.

zhenbinwu:
@nhanvtran "can you test the 3 layer in 2017.2 or 2017.4 as long as the macro HLS_SYN is available?" 

Yes, the 3layer passed the test with v2017.2.

nhanvtran:
thanks @vloncar for the review and @zhenbinwu for the additional tests!

PR #85, opened 2018-08-06
Change weight arrays in example projects to only 1 dimension.  Fixes crash in csim

benjaminkreis:
Found that dense layers of old example projects crash in csim due to dimension of weight arrays.  Change from `weight[x][y]` to `weight[x*y]`.


nhanvtran:
Tested -- looks good, merging

Issue #79, opened 2018-07-17
Syn failed for 3Layer with sublayer


zhenbinwu:
The current master branch of the hls4ml can't synthesize  the 3layer Model. 
```
ERROR: [XFORM 203-103] Array 'mult.V' (/data/benwu/HLS4ML_2018/hls4ml/nnet_utils/nnet_layer.h:56): partitioned elements number (2048) has exeeded the threshold (1024), which may cause long run-time.
ERROR: [HLS 200-70] Pre-synthesis failed.
command 'ap_source' returned error code
    while executing
"source [lindex $::argv 1] "
    ("uplevel" body line 1)
    invoked from within
"uplevel \#0 { source [lindex $::argv 1] } "

INFO: [Common 17-206] Exiting vivado_hls at Fri Jul 13 12:00:25 2018...
```
The error can be trace to https://github.com/hls-fpga-machine-learning/hls4ml/pull/62/files#diff-b37c065f136460b015788b96b5c25102L52

jmduarte:
Somehow the issue comes from re-writing the `mult` array as a 1D array instead of a 2D array: 

https://github.com/hls-fpga-machine-learning/hls4ml/commit/d90e74a7c92eb63e675e27396177cb05b9286f00#diff-b37c065f136460b015788b96b5c25102L96 
https://github.com/hls-fpga-machine-learning/hls4ml/commit/d90e74a7c92eb63e675e27396177cb05b9286f00#diff-b37c065f136460b015788b96b5c25102L96 

Conversely if I comment out the `mult` array partitioning (much like the `weight` array is *not* partitioned), it also works again (although I see some other warnings)
https://github.com/hls-fpga-machine-learning/hls4ml/blob/d90e74a7c92eb63e675e27396177cb05b9286f00/nnet_utils/nnet_layer.h#L70

Maybe it would help if I knew why the `weight` array partitioning was commented out. It looks like @benjaminkreis did it?

nhanvtran:
just to record this for everyone...

So this error actually looks like a difference in Vivado version.  This does not occur with 2017.2.  It looks like Xilinx added an array size limit for fully partitioning in 2018.2 -- probably because of our complaints about RAM  limits in the compiler 😱

SteCla0:
Error still exists with 16x100x100x100x100x5 example in hls4ml v0.6.0 and vivado 2020.1

PR #86, opened 2018-08-08
Fix sublayer to use layer code (+ pruning support + default unroll limit of 4096)

jmduarte:
This address issue #79, where synthesis fails in Vivado 2018.2

In addition, we now count the number of zeros in sublayer call to appropriately set the multiplier limit. 

However, there's still an open issue, in that the initiation interval is now 2 instead of 1 with this code. It seems to be related to the multiplier limit code because without the limit the initiation interval goes back to 1.

Tested with the following `keras-config.yml`:
```
KerasJson: example-keras-model-files/KERAS_3layer.json
KerasH5:   example-keras-model-files/KERAS_3layer_70pruned_retrained_weights.h5 
OutputDir: my-hls-test
ProjectName: myproject
XilinxPart: xcku115-flvb2104-2-i
ClockPeriod: 5

IOType: io_parallel # options: io_serial/io_parallel
ReuseFactor: 1
DefaultPrecision: ap_fixed<16,6> 
```
and got latency of 22 clocks for the 3-layer model:
```
================================================================
== Performance Estimates
================================================================
+ Timing (ns): 
    * Summary: 
    +--------+-------+----------+------------+
    |  Clock | Target| Estimated| Uncertainty|
    +--------+-------+----------+------------+
    |ap_clk  |   5.00|     4.366|        0.62|
    +--------+-------+----------+------------+

+ Latency (clock cycles): 
    * Summary: 
    |   22|   22|    2|    2| function |

================================================================
== Utilization Estimates
================================================================
* Summary: 
+---------------------+---------+-------+---------+--------+
|         Name        | BRAM_18K| DSP48E|    FF   |   LUT  |
+---------------------+---------+-------+---------+--------+
|DSP                  |        -|      -|        -|       -|
|Expression           |        -|      -|        0|       6|
|FIFO                 |        -|      -|        -|       -|
|Instance             |       13|    954|    23564|   42434|
|Memory               |        -|      -|        -|       -|
|Multiplexer          |        -|      -|        -|      69|
|Register             |        -|      -|     3201|       -|
+---------------------+---------+-------+---------+--------+
|Total                |       13|    954|    26765|   42509|
+---------------------+---------+-------+---------+--------+
|Available SLR        |     2160|   2760|   663360|  331680|
+---------------------+---------+-------+---------+--------+
|Utilization SLR (%)  |    ~0   |     34|        4|      12|
+---------------------+---------+-------+---------+--------+
|Available            |     4320|   5520|  1326720|  663360|
+---------------------+---------+-------+---------+--------+
|Utilization (%)      |    ~0   |     17|        2|       6|
+---------------------+---------+-------+---------+--------+
```


jmduarte:
Sublayer is now working (with pruning support) and reuses existing layer code.

To test with the standard 3-layer model, you can simply change the `MAX_MULT` in `keras-to-hls.py` to a smaller number, like `1024`... Or you can use the new 16x100x100x100x5 dense model example (which takes hours to synthesize).

zhenbinwu:
Does this split the weights? Does it help with the syn time?

jmduarte:
@zhenbinwu I don't think it helps with synth time, but it is cleaner now, since we can reuse `compute_layer` rather than having a specialized function.

zhenbinwu:
Yeah, I think it is a great idea to reuse nnet_layer only.

jmduarte:
@nhanvtran this should be ready to be merged!

jmduarte:
@nhanvtran, that's a good point. this won't work for `pytorch-to-hls.py` yet, but it should be easy to add the analogous lines of code there.

zhenbinwu:
What are the requisite hooks? I was expecting the pytorch-to-hls.py works automatically with this change.

nhanvtran:
I saw that MAXMULT only lives in `keras-to-hls.py` -- doesn't that mean it won't be picked up by pytorch?

keras-to-hls/keras-to-hls.py
```diff
@@ -191,8 +193,15 @@ def main():
...
-                    layer['n_part'] += 1   
+                    layer['n_part'] += 1
+                for i_part in range(0,layer['n_part']):
+                    i_subout = 0
+                    if i_part>0:
+                        i_subout = sum(layer['n_subout'][0:i_part])
+                    cur_n_zeros = print_array_to_cpp("w{}".format(layer_counter), weights, yamlConfig['OutputDir'], i_part, layer['n_part'], i_subout, layer['n_subout'][i_part])
+                    print_array_to_cpp("b{}".format(layer_counter), biases, yamlConfig['OutputDir'], i_part, layer['n_part'], i_subout, layer['n_subout'][i_part])
+                    layer['weights_n_subzeros'].append(cur_n_zeros)
+
```
zhenbinwu: Can this move to the central hls_writer.py? Then this can be picked up by Pytorch-to-hls.py as well

nhanvtran:
up to you @jmduarte @zhenbinwu if you want to put it in this PR or a separate one.  if it's a quick fix, might as well put it in this one

zhenbinwu:
@nhanvtran @jmduarte I am fine with merging this PR. But this get me thinking we probably should move more functionality to the central hls_writer.py so that both keras/pytorch can pick up the code change easily. 

After the pumping, I will follow up on this and get the pytorch-to-hls.py up to date.

nhanvtran:
Ok i'm going to merge this then and we'll add the pytorch support in another PR because this involves some refactoring of hls-writer

simplelins:
hi, 
   for vitis_hls 2020, we checked the document "[UG1391 - Vitis HLS Migration Guide (v2020.2)](https://www.xilinx.com/support/documentation/sw_manuals/xilinx2020_2/ug1391-vitis-hls-migration-guide.pdf)" , the command ' config_array_partition -maximum_size 4096' is unsupported. so the issue should be continue to check, thanks

PR #89, opened 2018-08-23
Recursive2

violatingcp:
This is the full implementation of LSTM. RNN is also there, but cannot be converted from Keras yet.


violatingcp:
Yes, its recurrent. this PR contains GRU and LSTM.

vloncar:
Continued in #193 and #329

PR #93, opened 2018-08-31
Fix dense layer instantiation for io_serial mode

ejk43:
A few things needed to make serial mode work well for dense layers:
1. Partition the "flattened" weight vector correctly
2. Edit the softmax operation in serial mode so it accesses the input in sequential order
3. Add a RESOURCE directive to instantiate weights into BRAMs if requested

As best as I can tell, the 1layer and 3layer example projects are working well with these updates. Item 2 *should* be compatible with parallel mode -- but I have not gotten a chance to test this yet so I'll take an action to exercise softmax with io_parallel before the PR gets merged. 

The end result appears to be working as expected and completes both the synthesis and RTL simulation successfully... Honestly, the performance vs resource tradeoffs we can evaluate using the reuse_factor is WAY better than anything I had in the old RFNOC repo :)


nhanvtran:
I tested serial and parallel mode and it seems fine.  

For parallel mode, I had to add this line to the tcl which is part of a larger issue we need to discuss with Xilinx 
`config_array_partition -maximum_size 4096`

But this is a different issue.  As far as serial mode goes, I think this is ready to merge.  

@ejk43 do you want test anything else?

ejk43:
@nhanvtran  Nothing else on my end-- thanks for checking out parallel mode! 

Agree this is good to go for now.

PR #94, opened 2018-08-31
Automated testing framework

vloncar:
This adds the testing framework. Most of the changes are in the test directory, which contains the bash scripts used to contruct the Jenkins pipeline, as well as the Dockerfile which creates a development environment in which all testing takes place. In order to connect hls4ml github project to Jenkins instance this needs to be merged.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,127 @@
+pipeline {
+  agent {
+    docker {
+      image 'hls4ml-with-vivado:latest'
+    }
+  }
+  options {
+    timeout(time: 3, unit: 'HOURS')
+  }
+
+  stages {
+    stage('Keras to HLS') {
+      parallel {
+        stage('py36') {
+          steps {
+            dir(path: 'test') {
+              sh '''#!/bin/bash
+                 . activate hls4ml-py36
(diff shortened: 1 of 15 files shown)


vloncar:
Brief overview of the changes outside `test/` directory:
- `Jenkinfile` - File describing Jenkins pipeline (build process).
- `build_prj.tcl` - These changes allow project building to be parametrized, i.e., you can just build the Vivado HLS project (for GUI) and do csim, but not synth and cosim. Just doing `vivado_hls -f build_prj.tcl` still does the same thing as it did before. This is useful as it is used to break the build process into several smaller steps, as opposed to it being one giant step.
- `hls_writer.py` - Required to make the files generated by Python 2 and Python 3 identical. Otherwise, values in `w*.h` files are written with different in the number of decimal places. Due to a rather weird bug in numpy < 1.15, `format()` function errors out, so I use the old `%` syntax.
- `keras-to-hls.py` - This small change is required when different processes access the file in parallel, during Jenkins build. By default, a write lock is obtained, and other processes report an error, which stops the Jenkins build. Opening in read-only mode uses a shared lock and the problem goes away.
- `nnet_conv.h` - Fixes a problem with csim in one of the example models. This was mentioned on one of the previous biweekly meetings.

PR #95, opened 2018-08-31
Conv csim fix

vloncar:
C simulation of KERAS_conv1d example model fails due to the size of mult array, which get allocated on stack. Synthesis however, works as expected. So instead stack allocation, allocate using `malloc` during csim, but still use standard `[]` allocation during synthesis.

### DIFF
--- a/nnet_utils/nnet_conv.h
+++ b/nnet_utils/nnet_conv.h
@@ -91,32 +91,36 @@ void conv_1d(
 	     typename CONFIG_T::bias_t    biases[CONFIG_T::n_filt])
 {
 
+#ifdef __SYNTHESIS__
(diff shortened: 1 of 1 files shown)


nhanvtran:
what happens if we keep the fix for c-sim in synthesis?  @benjaminkreis can also comment but he's traveling at the moment

vloncar:
You can't use `malloc`/`free` or `new`/`delete` or any other dynamic memory allocation functions in synthesis, meaning it would fail when it encounters these functions. This is documented in "Dynamic memory usage" section of UG902 (on page 302), and the proposed fix is taken from the example on page 325 (assuming guide for version 2017.2, probably slightly different for later versions of the document).

I have jumped the gun when I said "synthesis works as expected". It appears to be working (doesn't fail or get stuck), but it takes enormous amount of time. I have tried running it for a few hours, there is progress, but I didn't wait for it to complete. I don't think that is related to this change, as synthesis takes the same code path as before.

PR #98, opened 2018-09-06
Automated testing: Add io_serial tests for 1layer and 3layer networks

ejk43:
A modest change that simply adds a few serial mode tests. I first tried to include the options as a new line in the keras-models.txt file (like, "KERAS_1layer -s") but this was not compatible with the Jenkins xargs and bash scripts... I ended up using a new text file for serial mode and adding a new command to the Jenkinsfile to read in the new file. 

I needed to append io type onto the "vivado_prj" sub-folder in order to not conflict with the parallel mode test. I was able to run the serial mode tests correctly without failures. 

One thing to mention: when I first tried to run through all test cases, my run did not finish after 24 hours-- is this expected behavior right now? Seems to be slowing down on some of the larger serial mode tests. I'm trying again tonight to see if it was some kind of mistake and I'll save the logs to review this time...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,8 @@ pipeline {
             dir(path: 'test') {
               sh '''#!/bin/bash
                  . activate hls4ml-py36
-                 cat keras-models.txt | xargs ./keras-to-hls.sh -p 3'''
+                 cat keras-models.txt | xargs ./keras-to-hls.sh -p 3
+                 cat keras-models-serial.txt | xargs ./keras-to-hls.sh -p 3 -s'''
             }
           }
         }
@@ -25,7 +26,8 @@ pipeline {
             dir(path: 'test') {
               sh '''#!/bin/bash
                  . activate hls4ml-py27
-                 cat keras-models.txt | xargs ./keras-to-hls.sh -p 2'''
+                 cat keras-models.txt | xargs ./keras-to-hls.sh -p 2
+                 cat keras-models-serial.txt | xargs ./keras-to-hls.sh -p 2 -s'''
             }
           }
         }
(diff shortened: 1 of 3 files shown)


vloncar:
LGTM.

I plan on enhancing the way the tests are defined once we get the Jenkins up and running. The idea is to either move the test configuration to Jenkins or to implement a mechanism that would allow Jenkins to override default configuration. In addition to testing both parallel and serial mode for the same models, I can see that we will need the same thing for the reuse factor, and maybe even for the xilinxpart. In the meantime your changes should suffice.

Regarding the long running time, the offender here is KERAS_conv1d model. I have the same issue as you do, and I think others have also encountered it. We should remove it from keras-models.txt until the solution is found.

nhanvtran:
hi, talking with @jmduarte and @benjaminkreis conv1d is an old and big model that hasn't been fully tested.  let's remove it from the list.  do we want to add a model like the one from Grzegorz as a nice serial mode case that starts to use the BRAM?

vloncar:
@ejk43 I have expanded the testing scripts, so now we can define what we want to test in `keras-models.txt` (or any other file). With the new `convert-keras-models.sh` we can read the models defined in a file or an environment variable (the latter one being useful for changing the tested models in Jenkins without new commits). It is much more flexible, essentially we can define a test of any combination of parameters e.g., we can define different models for serial and parallel IO, different models for different boards, different types for different models etc. Current WIP is in my [branch](https://github.com/vloncar/hls4ml/tree/jenkins-ci). A few bits are still missing, but eventually I plan on making a PR with that work that would supersede this one. I would appreciate any feedback and ideas on how to expand it further.

ejk43:
Closing in favor of #100

PR #100, opened 2018-09-19
Expand automated test scripts

vloncar:
This expands the scripts used to start the test and allows multiple tests to be defined with different properties (e.g., some with parallel IO, some with serial, yet some other with different reuse factor, clock, type and board...). Also included in the tests are Jen's models for batch normalization. Supersedes #98


nhanvtran:
thanks @vloncar, thanks!  

naive question, how should we review these PRs?  we have to merge them to do an automated testin right?  could we include that standalone script which runs the machinery as a way to validate CI-related PRs?

vloncar:
The standalone script `hls4ml-keras-test.sh` is included. You can run it to see that every model defined in `keras-models.txt` is working, thus validating the scripts. Jenkins does the same thing, but its script is `Jenkinsfile`, and that is a bit more difficult to test until we have Jenkins instance up and running (and integrated with GitHub). I test it by using my own local Jenkins instance and it works, however that is not a ideal way to test since I don't have the GitHub hooks, meaning it is not automated in the way that our GitHub-integrated Jenkins instance at Fermilab will be.

Jenkinsfile
```diff
@@ -16,7 +16,7 @@ pipeline {
             dir(path: 'test') {
               sh '''#!/bin/bash
                  . activate hls4ml-py36
-                 cat keras-models.txt | xargs ./keras-to-hls.sh -p 3'''
+                 ./convert-keras-models.sh -p 3 -f keras-models.txt'''
```
ejk43: Does this need to have `-x` as a command line arg? Just running this command did not generate converted output for me

test/keras-models.txt
```diff
@@ -1,7 +1,32 @@
...
 KERAS_1layer
 KERAS_3layer
-KERAS_3layer:KERAS_3layer_70pruned_retrained_weights
-KERAS_conv1d
+#KERAS_3layer:KERAS_3layer_70pruned_retrained_weights
+#KERAS_conv1d
 KERAS_conv1d_small
 KERAS_conv2d_model
 KERAS_dense_big
+KERAS_3layer_batch_norm
```
ejk43: The 3layer_batch norm failed for me with the typical failure case:

```
ERROR: [XFORM 203-103] Array 'mult.V' (/home/ejk/work/HLS4ML/nnet_utils/nnet_layer.h:56): partitioned elements number (2048) has exeeded the threshold (1024), which may cause long run-time.
ERROR: [HLS 200-70] Pre-synthesis failed.
command 'ap_source' returned error code
    while executing
"source [lindex $::argv 1] "
    ("uplevel" body line 1)
    invoked from within
```

I used Vivado 2017.4. Is this maybe a version issue? Or, does something need to include the directive in the test process that bumps up the max number of elements?

nhanvtran: Which branches are being compared?  

I thought we had that directive in the tcl:
https://github.com/hls-fpga-machine-learning/hls4ml/blob/master/hls-template/build_prj.tcl#L24

Jenkinsfile
vloncar: Good catch! I have fixed it in the latest commit.

test/keras-models.txt
nhanvtran: Yeah, I think this directive went in after this branch was made so that makes sense.

@benjaminkreis was just telling me yesterday that `config_array_partition -maximum_size 4096` doesn't work in 2017.2 so I guess we are "supporting" 2018.1 and above...

benjaminkreis: I like the idea of supporting older versions if it's only a matter of changing that line to that!

vloncar: With this change and with 2017.2 the 3layer_batch norm model synthesizes. I can submit a PR, but this change only prevents the build script from failing on that particular line. It doesn't solve the underlying problem that we cannot increase the maximum partition size on older versions of HLS and the issue will probably resurface with some future models. Still want it?

benjaminkreis: Yes, I think we still want it.  The only goal of that line is to set the max above our own threshold for our "sublayer" approach.  With older versions, we won't be setting a max, but the "sublayer" approach will kick in at 4096 and prevent memory issues.  With newer versions, the max will increase to 4096, and again the sublayer approach kicks in at 4096 to prevent memory issues.  Chatted with @nhanvtran, and he agrees.

PR #102, opened 2018-09-28
Initial gitignore file to ignore pycache and a few other folders

ejk43:
Pretty trivial change -- just adds a gitignore to clean up some local folders from git status.


nhanvtran:
sounds good!

PR #109, opened 2018-10-15
[WIP] Compare synthesis reports with previous builds

vloncar:
Adds script (`compare-reports.sh`) that compares reports generated from previous stage (`gather-reports.sh`). On Jenkins side, this will work by saving the report as the build "artifact", and then use the artifact of last successful build of master branch as the baseline for comparison. Not yet complete, currently only records artifacts. The comparison part will be added once I see how the artifacts are exposed through Jenkins API.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,7 @@ pipeline {
     }
   }
   options {
-    timeout(time: 3, unit: 'HOURS')
+    timeout(time: 6, unit: 'HOURS')
   }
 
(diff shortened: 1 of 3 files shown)


vloncar:
I have the second part (comparison in Jenkins pipeline) ready, however, in order for it to work, master branch needs to produce artifacts first, otherwise tests will always fail. Therefore this PR needs to be in master before I can submit the second part as a separate PR.

PR #110, opened 2018-10-17
Compare reports in Jenkins pipeline (part 2 of PR109)

vloncar:
Now that master branch saves reports, call the script to compare reports in Jenkins pipeline


nhanvtran:
@vloncar what's the state of this? ready for merging?

vloncar:
@nhanvtran It can be merged as is, but I wanted to discuss if this solution is adequate, since it is a very naive check, i.e., it will fail if there are any differences (increase or decrease) in resource consumption between reports.

PR #111, opened 2018-10-19
softmax: Move the HLS PIPELINE command into the outer loop

ejk43:
I ran into a few issues with PIPELINE in the inner loop. Basically,
I could not meet timing for the device I was simulating (xc7z020clg484-1)
for the keras_3layer example, and the critical path traced back to
the exp_diff_res lookup and accumulate. With some empiprical testing,
I found moving HLS PIPELINE to the outer loop gets much better results:

1. Timing improved from a min of 6.6 ns to 4.32
2. Latency reduced in the softmax from 106 to 70 (likely driven by size of the softmax input)
3. Other resource usage stayed pretty similar. A big more usage in softmax, but OK overall

This should only impact serial mode

### DIFF
--- a/nnet_utils/nnet_activation.h
+++ b/nnet_utils/nnet_activation.h
@@ -238,10 +238,10 @@ void  softmax(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
       exp_res[ii] = 0;
     }
     for (int ii=0; ii<CONFIG_T::n_in; ii++) {
+      if (CONFIG_T::io_type == io_serial){
+          #pragma HLS PIPELINE
+      }
       for (int jj=0; jj<CONFIG_T::n_in; jj++) {
-        if (CONFIG_T::io_type == io_serial){
-            #pragma HLS PIPELINE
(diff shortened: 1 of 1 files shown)


nhanvtran:
@ejk  a bit ago there was some issue where aerial mode would not finish cosim. Does this also help with that?  Or can you remind me what was the resolution for that?

ejk43:
Hm, I think the serial mode hanging during csim was fixed for dense layers in the PR here: https://github.com/hls-fpga-machine-learning/hls4ml/pull/93

... Specifically fixed by this change in nnet_activation.h softmax function to cache data before operating in the double for-loop: https://github.com/hls-fpga-machine-learning/hls4ml/pull/93/files#diff-603f8d121a5a8f419b572162a6538708

I havent yet progressed to conv mode yet, so that might still be failing csim -- TBD. 

BTW, someone else has the ejk tag -- I'm ejk43 here...  I got ejk on gitlab though :)

ejk43:
Should be good to merge, thanks!

PR #114, opened 2018-10-25
Binary/ternary dense network

jngadiub:
Add support for binary dense network.

Tested with this model (equivalent to default 3 layer model + batch normalization)

https://user-images.githubusercontent.com/5583089/40817230-f029045e-6516-11e8-8447-fe48c65dcf2c.png
https://github.com/hls-fpga-machine-learning/keras-training/blob/master/models/models.py#L105

Test with <16,6> precision and these inputs

```
0.63117, 0.676428, 0.0257138, 0.00082122, 0.0243723, 0.000579686, 0.336411, 0.163256, 0.336411, 0.610527, 0.593471, 0.534274, 0.706917, 0.566858, 0.0145026, 0.136126
```
the results are

```
hls: -0.734375 -0.0185547 -0.549805 -0.886719 -1.17383
keras: -0.73436 -0.00826645 -0.547652 -0.889665 -1.18085 
```

Compare latency and resources below.

* Latency for binary dense model

latency_binary_dense

* Latency for dense model

latency_dense

* Resources for binary dense model

resources_binary_dense

* Resources for dense model

resources_dense

The physics performance for this binary dense model are worse by about 10% wrt the default dense model. Better performance can be obtained by increasing the number of parameters. I have a bigger binary dense model with 10x more parameters but it takes long time to synthetize it and it might crash at the end. Perhaps 2-5x more parameters it is sufficient to recover the performance.

The ternary dense model should be straightforward to add as well. I'll update the PR with it next.


jngadiub:
Updated the PR to support also ternary dense models.

Tested on the same set of inputs as above, the results are:

```
keras: -0.886669 0.368325 -0.719488 -0.668888 -1.15423
hls: -0.84082 0.160156 -0.662109 -0.704102 -0.998047
```

* Latency:

latency_ternary_dense

* Resources:

resources_ternary_dense

nhanvtran:
@jngadiub 
cool!  what are the model details?  do you want someone to start reviewing?

jngadiub:
The binary and ternary NN models I tested so far have the same number of layers and node per layers as the usual 3 layer model + batch normalization:

https://user-images.githubusercontent.com/5583089/40817230-f029045e-6516-11e8-8447-fe48c65dcf2c.png
https://user-images.githubusercontent.com/5583089/40817238-f93eff26-6516-11e8-9786-835542da9b20.png

The corresponding weights+model files are added under the example-keras-model folder. The weights come out from the training non binarized/ternarized, however they are typically quantized in the inference process. To avoid doing this operation inside hls, I do it outside in the hls-writer so that the weights loaded in hls are already +/-1,0:

https://github.com/jngadiub/hls4ml/blob/bnn/hls-writer/hls_writer.py#L813-L821

I think the review can start if we are interested in supporting these particular cases in the package. After that, I guess we can move to binary/ternary conv2d.

keras-to-hls/keras-config.yml
```diff
@@ -1,10 +1,10 @@
-KerasJson: example-keras-model-files/KERAS_3layer.json
-KerasH5:   example-keras-model-files/KERAS_3layer_70pruned_retrained_weights.h5
+KerasJson: example-keras-model-files/KERAS_3layer_ternary_small.json
```
vloncar: Do we really want to change the default example (intended for new users to get into hls4ml) to be ternary dense model?

Also, adding this to the automated tests would be nice.

jngadiub: I agree. I changed the default example to the usual 3-layer model. And also added bnn/tnn to the list of models for automated test.

jngadiub:
Can we merge this?

drankincms:
I don't know what the protocol is for this. Should I merge or is @nhanvtran supposed handle/authorize the merge?

nhanvtran:
this is good -- in a new PR we should add the binary/ternary to the standard jenkins tests...

vloncar:
@nhanvtran Jen added the tests in 556ba08353009e3fd5c9dc94b00bfe7c7d412b4d. No need for a new PR.

Issue #75, opened 2018-07-10
AveragePooling1D/2D/3D Layer


nhanvtran:
1D: https://github.com/keras-team/keras/blob/master/keras/layers/pooling.py#L87
2D: https://github.com/keras-team/keras/blob/master/keras/layers/pooling.py#L225
3D: https://github.com/keras-team/keras/blob/master/keras/layers/pooling.py#L389

Issue #74, opened 2018-07-10
Max Pooling layer


nhanvtran:

benjaminkreis:
Reminder that @ejk43 has a version of this here: https://github.com/Xilinx/RFNoC-HLS-NeuralNet/blob/master/rfnoc/hls/nnet_lib/nnet_pooling.h

Also Alexx has a "find max" example: https://github.com/p2l1pfp/GlobalCorrelator/pull/18

PR #117, opened 2018-11-01
Max/Average pooling

thesps:
This PR contains the addition of max and average pooling to hls4ml.
The HLS implementation is in ```nnet_utils/nnet_pooling.h```
Changes to `keras-to-hls.py` and `hls-writer.py` add the pooling layer to a project.
Related to issues #74, #75


thesps:
I've been testing the pooling layer in isolation for some small test cases in Keras, created like:
```
model = keras.models.Sequential()
model.add(keras.layers.InputLayer(input_shape = ... ))
model.add(keras.layers.AveragePooling2D(input_shape = ..., padding = 'same' ))
model.predict(x)
```
For some desired input shape and padding, with a randomised test vector `x`.
I've used input shapes 4x4x1 and 5x5x1, the pool size at its default of 2x2(x1) with stride 2x2(x1).
These inputs and the corresponding output shapes are depicted in the images below .
The group of hatched pixels in the input images (left) are pooled into the pixel with the same hatching in the output image (right). 
For the 5x5x1 case, the 'same' padding adds a row and column of '0's. The pooling operation ignores these values however, the padding just allows the pixels that would otherwise be clipped away to contribute to the output.
All these tests have been with `ap_int<16>`, IO parallel, reuse 1.

pooling_4x4x1img_2x2pool
pooling_5x5x1img_2x2pool

Here's a summary of the different scenarios tested so far:

| Input shape | Pool size | Pooling operation | CSim Test performance | LUT | Latency | II |
| --- | --- | --- | --- | --- | --- | --- |
| 4x4x1 | 2x2 | Max | Exact agreement | 418 | 1 | 1 |
| 4x4x1 | 2x2 | Average | Differences up to LSB value | 594 | 1 | 1 |
| 5x5x1 | 2x2 | Max | Exact agreement | 612 | 1 | 1 |
| 5x5x1 | 2x2 | Average | Differences up to LSB value, or up to 2xLSB value in padded region | 1003 | 1 | 1 |
 | 32x32x1 | 2x2 | Max | Not tested | 61,000 | 1 | 1 | 
 | 64x64x1 | 2x2 | Max | Not tested | 244,000 | 1 | 1 | 

There are two classes of discrepancies I see with the AveragePooling. In the 4x4x1 case, where there is no padding, some rounding means that the result with `ap_int` or `ap_fixed` can differ from the floating point. For example averaging [1, 2, 3, 4] using integers gives 2 rather than 2.5. This is probably fine, but could be avoided by using a wider type for the output than the input. For pixels within the padding some bigger differences can be seen due to the way I've implemented the padding. For example, for the 5x5x1 image, the only pixel that contributes to the lower right output pixel is the lower right input pixel. If this has a value of e.g. 7, the output should be 7. To try to keep the code simple, I take the average of [7, 0, 0, 0] and multiply, so the result is 7 / 4 * 4 = 4 using integers. I'll fix this issue.

I'll try out resource reuse and different II next.

**Note,** using a 64x61x1 (= 4096 pixels) input I get an error during CSynth: 
`ERROR: [XFORM 203-133] Bitwidth of reshaped elements (65536 bits) exceeds the maximum bitwidth (65536 bits) for array 'data.V' (firmware/myproject.cpp:34).`
I do not get the same error if I change the pragma from `#pragma HLS ARRAY_RESHAPE variable=data complete dim=0`, to `#pragma HLS ARRAY_PARTITION variable=data complete dim=0`.
Would there be any consequences to changing this pragma everywhere in the hls_writer?

nhanvtran:
> Note, using a 64x64x1 image (= 4096 pixels) input I get an error during CSynth:
ERROR: [XFORM 203-133] Bitwidth of reshaped elements (65536 bits) exceeds the maximum bitwidth (65536 bits) for array 'data.V' (firmware/myproject.cpp:34).
I do not get the same error if I change the pragma from #pragma HLS ARRAY_RESHAPE variable=data complete dim=0, to #pragma HLS ARRAY_PARTITION variable=data complete dim=0.
Would there be any consequences to changing this pragma everywhere in the hls_writer?

We went with reshape so that the ports of the IP aren't out of control.  If there is a maximum allowed set of bits allowed into the IP I wonder if this is something that we should get around, or stream the inputs in like in serial mode...

thesps:
I've been looking at the impact of array partitioning on the timing and resource utilisation of the pooling.
The brief version is that the partitioning makes a much bigger impact than the function limit.

In all cases, an allocation limit is made on the function which does the pooling operation, calculated from the specified reuse factor with:
`#pragma HLS ALLOCATION instances=pool_op limit=limit function`
In the myproject.cpp I then use `#pragma HLS ARRAY_PARTITION ...` for the input and output images.

The model I'm using is just a MaxPooling layer which takes a 32x32x1 input, has pool size 2x2 and stride 2x2, giving a 16x16x1 output.
So there are 16 * 16 * 1 = 256 calls to the pooling function. 
The limit is set to 16 * 16 * 1 / reuse.

My intuition around the array partitioning is that the partition factor should allow as many parallel reads as there are pooling function instances. Any more then I have more ports than I can use (extra resources), and less then I can't fill all instances of the function (extra latency). So with smaller reuse I need more blocks (higher partition factor). I thought that cyclic partitioning made more sense than block, such that adjacent pixels would be in different arrays and therefore able to be read in parallel. I also think that it makes sense to have twice as many blocks in each partitioned dimension in the input image than output, since 2 input pixels from each dimension map to 1 output pixel.

It seems my intuition isn't completely borne out in the results, presented in the tables below. Block partitioning appear to be better than cyclic here. With one reuse factor, quite a variety of partition factors give reasonable results. For example with `reuse=16`, I see the best resource usage for either `factor=16` or `factor=1` on both inputs and outputs, with worse performance for factors between these. My intuition was that `factor=2` and `factor=1` for the input and output respectively were appropriate for `reuse=16`. 

I do see that if I don't allocate enough blocks the latency goes up, for example `reuse=4` and `factor=1` gives a latency of 18 where it is 4 with complete partitioning .

So I'm not sure what conclusions to draw yet, other than that partitioning gives a big gain. 
I'll update the tables with other results as I make them, particularly with some `reuse > 16` such that I will have to partition more than one dimension of the image.

With block/cyclic/complete partitioning on dim=1, complete partitioning on dim=2, dim=3:

| Reuse | Partition Type | In Partition Factor | Out Partition Factor | Latency | II | HLS LUT (x10³) | HLS FF (x10³) | Vivado LUT (x10³) | Vivado FF (x10³) |
| -------- | ------------------ | ----------------------- | ------------------------- | ---------- | -- |------|-----| --- | --- |
| 1 | Complete | NA | NA | 1 | 1 | 61 | - |
| 16 | Complete | NA | NA | 16 | 16 | 44.0 | 37.6 | 20.3 | 37.8 |
| **16** | **Block** | **16** | **8** | **18** | **16** | **9.1** | **1.3** | **9.4** | **1.3** |
| 16 | Cyclic | 16 | 8 | 18 | 16 | 14.0 | 1.3 | - | - |
| 16 | Block | 8 | 4 | 18 | 16 | 15.2 | 15.7 | 8.6 | 15.7 |
| 16 | Block | 4 | 2 | 18 | 16 | 13.8 | 13.6 | 7.5 | 13.6 |
| 16 | Block | 2 | 1 | 18 | 16 | 12.0 | 9.5 | 5.5 | 9.5 |
| 16 | Cyclic | 2 | 1 | 18 | 16 | 14.0 | 9.5 | 7.5 | 9.5 |
| **16** | **Block** | **16** | **16** | **17** | **16** | **7.2** | **4.9** | - | - |
| 16 | Block | 8 | 8 | 18 | 16 |  15.5 | 15.7 | - | - |
| 16 | Block | 4 | 4 | 18 | 16 | 14.1 | 13.6 | - | - |
| 16 | Block | 2 | 2 | 18 | 16 | 11.7 | 9.5 | - | - |
| **16** | **Block** | **1** | **1** | **18** | **16** | **8.5** | **1.3** | - | - |
| 4 | Complete | NA | NA | 4 | 4 | 50 | 37 |
| 4 | Block | 16 | 16 | 9 | 8 | 8.7 | 5.7 |
| 4 | Block | 8 | 8 | 10 | 8 | 16.5 | 14.9 |
| 4 | Block | 4 | 4 | 10 | 8 | 16.1 | 10.8 | 
| 4 | Block | 2 | 2 | 10 | 8 | 10.0 | 3.0 | 
| 4 | Block | 1 | 1 | 18 | 16 | 10.2 | 2.6 |
| 4 | Block | 16 | 8 | 10 | 8 | 8.8 | 2.6 |
| 4 | Block | 8 | 4 | 10 | 8 | 18.0 | 14.9 |
| 4 | Block | 4 | 2 | 10 | 8 | 16.0 | 10.8 |

With block/complete partitioning on dim=2, complete partitioning on dim=2, dim=3:

| Reuse | In Partition Factor | Out Partition Factor | Latency | II | LUT (x10³) | FF (x10³) |
| -------- | ----------------------- | ------------------------- | ---------- | -- |------|-----|
| 1 | Complete | Complete | 1 | 1 | 61 | X |
| 4 | Complete | Complete | 4 | 4 | 50 | 37 |
| 4 | 16 | 16 | 5 | 4 | 13 | 7 |
| **4** | **16** | **8** | **6** | **4** | **13** | **5** |
| 4 | 8 | 4 | 7 | 5 | 22 | 14 |
| 8 | Complete | Complete | 8 | 8 | 46 | 37 |
| **8** | **16** | **8** | **10** | **8** | **9** | **2.6** |
| 8 | 8 | 4 | 11 | 9 | 19 | 15 |
| 8 | 4 | 2 | 13 | 11 | 17 | 11 |
| 16 | Complete | Complete | 16 | 16 |  44 | 38 |
| **16** | **16** | **8** | **18** | **16** | **7** | **1** |
| 16 | 8 | 4 | 19 | 17 | 17 | 16 |
|16 | 4 | 2 | 21 | 19 | 15 | 14 |

nhanvtran:
@thesps this is really interesting to see

Some unordered set of comments:
   - at leading order, a main part of the tool is that you define a reuse factor and can implicitly control the II.  do you have a feeling for why this doesn't always work for reuse = 4?  It seems to be pretty successful for reuse = 16.  i was thinking that if reuse, and in/out block partition factor are all the same, it would guarantee II is the same, but it doesn't seem to be the case.  
   - in the conv case, the N-D array is flattened into a big 1-D array.  this might help to control better the behavior in a more intuitive way.  is it possible in the pooling case?  (in my admittedly naive understanding)
   - how's the synthesis time?  do you notice that for larger partition factors, is the HLS synthesis faster?

benjaminkreis:
This is pretty puzzling to me.  I think I also would have expected cyclic to be better, and also for there to not be much of a difference between partitioning dimension 1 versus 2.

For the case of strides that lead to no overlap in the pool, I wonder if we can create our own partitioning by dumping the input into a reordered array which we then block partition.  It's micromanaging more than we would hope, so I'm not sure we want to resort to this yet.  But it wouldn't be as horrible as in the conv2d case because we don't have to keep track of the order/address within each pool.

I just wanted to note that we are trying to figure this out also for the conv2d layer.  The approach we have been investigating is different in a couple of ways:
1) Our intermediate objects, the multiplier results ("mult") and the accumulator ("acc"), are both single arrays for the whole layer.  In your case, you create one pool array per output.  
2) We have been looking into partitioning the intermediate objects and seeing if Vivado figures out the rest before and after, while you are partitioning the input and output objects and seeing if Vivado figures out the middle.  

In both cases, I don't know if there is a right answer, so maybe it's good that we are attacking this from a few different angles.  Though for the second item, I wonder if it complicates the interface to the previous and following functions at the top level.

hls-template/build_prj.tcl
```diff
@@ -16,8 +16,8 @@ foreach arg $::argv {
 
 open_project -reset myproject_prj
 set_top myproject
-add_files firmware/myproject.cpp -cflags "-I[file normalize nnet_utils]"
-add_files -tb myproject_test.cpp -cflags "-I[file normalize nnet_utils]"
+add_files firmware/myproject.cpp -cflags "-I[file normalize nnet_utils] -std=c++0x"
```
vloncar: Since we now rely on C++11, we should be using `-std=c++11`. c++0x is not recommended.

thesps: My understanding was that Vivado HLS does not support -std=c++11, but c++0x is okay [1]. Maybe this is no longer true in most recent versions?

[1] https://forums.xilinx.com/t5/SDSoC-Environment-and-reVISION/Error-when-using-flag-quot-std-c-11-quot/td-p/704847

vloncar: I tried it on 2017.2, and it works.

thesps: For CSIM or CSYNTH?

vloncar: csim. Didn't try synth.

nhanvtran: sorry if i missed it, is there a specific this flag is needed in the specific PR?

vloncar: I was wrong, it appears c++11 is not really supported and we should use c++0x [1].

[1] https://forums.xilinx.com/t5/Vivado-High-Level-Synthesis-HLS/Can-HLS-2017-02-support-std-c-11/td-p/896120

vloncar: Hmm... on the machine where we have Vivado 2017.2, gcc is 4.8.5 and it accepts `-std=c++11`. But the default one (4.6.3) does not. It appears we have manually changed the compiler at some point :-). Anyway, no change is needed.

@nhanvtran `constexpr` is C++11, used in `nnet_helpers.h`

thesps:
@nhanvtran 
> do you have a feeling for why this doesn't always work for reuse = 4?

It seems the numbers for reuse 4 were just wrong, I automated the scan and produce the table below.
Now it makes more sense, you can get Latency = 4, II = 4 using many blocks, or Latency = 18, II = 16 with just 1 block, and sensible results in between. The 'most sensible' configurations with factor=4 on the input, and factor=2, 4, or 8 on the output are not exactly where I expected though. 
For the PR I think it makes sense to avoid the partitioning for now (it's not in the code), until it's better understood. The `#pragma HLS ALLOCATION` is in there, which gives a sensible, controllable Latency and II for complete partitioning, but with extra resources.

| Reuse Factor | Reuse Type | In Partition Factor | Out Partition Factor | Latency | II | HLS LUT | HLS FF |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 4 | Block | 1 | 1 | 18 | 16 | 13462 | 5220 |
| 4 | Block | 1 | 2 | 18 | 16 | 13942 | 5220 |
| 4 | Block | 1 | 4 | 18 | 16 | 15126 | 5220 |
| 4 | Block | 1 | 8 | 18 | 16 | 15510 | 5220 |
| 4 | Block | 1 | 16 | 17 | 16 | 11670 | 8036 |
| 4 | Block | 2 | 1 | 10 | 8 | 13668 | 5204 |
| 4 | Block | 2 | 2 | 10 | 8 | 13668 | 5204 |
| 4 | Block | 2 | 4 | 10 | 8 | 14468 | 5204 |
| 4 | Block | 2 | 8 | 10 | 8 | 15620 | 5204 |
| 4 | Block | 2 | 16 | 9 | 8 | 11780 | 7764 |
| 4 | Block | 4 | 1 | 10 | 8 | 15524 | 6740 |
| 4 | Block | 4 | 2 | 6 | 4 | 14754 | 5196 |
| 4 | Block | 4 | 4 | 6 | 4 | 14754 | 5196 |
| 4 | Block | 4 | 8 | 6 | 4 | 14946 | 5196 |
| 4 | Block | 4 | 16 | 5 | 4 | 13026 | 7244 |
| 4 | Block | 8 | 1 | 10 | 8 | 21748 | 15188 |
| 4 | Block | 8 | 2 | 8 | 6 | 22264 | 14672 |
| 4 | Block | 8 | 4 | 7 | 5 | 22062 | 13902 |
| 4 | Block | 8 | 8 | 6 | 4 | 20130 | 13388 |
| 4 | Block | 8 | 16 | 5 | 4 | 19170 | 15436 |
| 4 | Block | 16 | 1 | 10 | 8 | 15524 | 6740 |
| 4 | Block | 16 | 2 | 8 | 6 | 16312 | 6736 |
| 4 | Block | 16 | 4 | 7 | 5 | 16878 | 5710 |
| 4 | Block | 16 | 8 | 6 | 4 | 13026 | 5196 |
| 4 | Block | 16 | 16 | 5 | 4 | 13026 | 7244 |
| 4 | Block | 32 | 1 | 9 | 8 | 52392 | 36436 |
| 4 | Block | 32 | 2 | 7 | 6 | 53180 | 36432 |
| 4 | Block | 32 | 4 | 6 | 5 | 53746 | 35406 |
| 4 | Block | 32 | 8 | 5 | 4 | 49894 | 34892 |
| 4 | Block | 32 | 16 | 4 | 4 | 49894 | 36940 |

nhanvtran:
@thesps 

let me make sure i understand.  the goal (as I understand, as far as controlling the interval is concerned) is that if the user specifies `reuse = N`, then the synthesis should return `interval <= N` also.  it seems this is achieved by setting the `in` and `out` partition factors to `block` and also the block factor to be the same as `reuse`.

thesps:
Maybe I'm misunderstanding the partitioning, but my expectation was that with a reuse factor '`reuse`', the array should be partitioned such that there are `reuse` elements in each block of the output, and `reuse / pool_dimension` in each block of the input, so `factor = array_size / reuse` blocks for the output, and `factor = array_size / (reuse / pool_dimension)` for the input, rather than `factor = reuse` blocks. Then you can read one element from all blocks in parallel, and you can read the elements within each block sequentially, giving you the `II=reuse` we're looking for.
Okay, the example above simplifies things because the array size of the output is 16x16. But regardless, the behaviour is not always meeting my expectation. 

The correct `II=reuse` behaviour can be obtained with `complete` partitioning, but it seems like here, you only save significant resources if an array partition block factor is used in addition to the allocation limit. At this stage I'm not sure how to set it programmatically from the writer. I've put all the numbers for all block partition factors in [this gist](https://gist.github.com/thesps/6cefca78805cc49d4a32f82c57b6fa51), only going up to 16 because that's the size of one dimension of the output, and it seems like I should get that behaving before making it more complicated.

thesps:
I've updated [the gist](https://gist.github.com/thesps/6cefca78805cc49d4a32f82c57b6fa51), having scanned up to `reuse=256` now (so just 1 instance of the pooling function). With this, and after speaking with @nhanvtran, it seems that good performance can be achieved setting the partition factor programmatically. A special treatment might have to be made when going into the second dimension of the image, so I will still check whether this can be avoided (perhaps with a different partitioning in the 1st dimension).

nhanvtran:
@benjaminkreis @thesps i think this is ready for review now right?

benjaminkreis:
Yes, I will look soon!

benjaminkreis:
Looks good.  I personally tested the Max and Average 2D pooling on a little conv2d.  Merging!

PR #125, opened 2018-12-13
Expand Docker image and instructions

vloncar:
Adds additional user-oriented functionality to the docker image with the goal being that the image can be used as a development/testing platform for hls4ml.

The existence of this as an alternative installation method should be referenced somewhere in the hls4ml GitBook docs.


benjaminkreis:
Hi @vloncar, I'm trying this out.  A couple of quick questions:  

1) In the first steps, what is meant by "Put the Vivado installation folder here".

2) It seems I need to be root to run the first docker command on correlator2.  Is this expected?

benjaminkreis:
Yeah, that sounds good.

I'm now able to build the docker container.  I did it for 2017.2 because I had that installation archive on hand.

I believe the next step is `docker run -it hls4ml-with-vivado bash`?  Can you add that to the README, too?  After I do that, I can get through exporting an hls4ml IP 👍

nhanvtran:
before totally closing it, can we put this in the gitbook too?  i reserved a spot for it

PR #127, opened 2018-12-29
Add timing to the build step

vloncar:
Mentioned in the last meeting, discussed with @nhanvtran. This times sim, synth, cosim and export phase of the `build_prj.tcl` script and displays elapsed time in human-readable form (hours, minutes and seconds).


nhanvtran:
@vloncar do you want to add the regex updates that @asvetek suggested before I merge this?

PR #128, opened 2019-01-04
Large mlp

violatingcp:
Example of code that compiles large MLPs. Models shown are the Hcal regression model with 1,10 and 40 inputs so you can test the scalability note that setup.sh is needed for 40 or larger models. A 75 input model is there to look at models with more than 1 Million weights. There is a lot more work to get this to work optimally.


ejk43:
@violatingcp Incredible! Finally had a chance to review your slide deck and code changes. This is really really great! 

I'm trying to jog my memory on a few of the issues we faced in the past and understand what your changes are doing here to fix these problems. I'd be happy to hear any of your insights into these items based on your experiments...
1. I thought there was some point where we had a comparable architecture (accumulator inside the multiply loop), but ended up pulling it out because it gave us better resource usage... I looked through git history and it appears we were always using unrolls at both of the loops, with different unroll factors. I assumed that "unroll" at the top level loop would be equivalent to what you're doing here, but I suspect one of the major improvements here is that you're changing the for-loop bounds and dynamically calculating the matrix indices. Thoughts?
2. I'm *really* surprised that you're not seeing a large amount of fabric/memory increase when you use dataflow at the top level. When I've tested serial mode, dataflow inserts a big ping-pong buffer between all the dataflow-ified functions, and it takes up a drastic amount of memory without inserting `HLS STREAM` pragmas which switch to NOT ping pong buffers. I'll have to play with it a bit. I wonder if it's because you've completely partitioned all arrays at the top level functions?
3. I think I may recall why we yanked out the accumulator: when the synthesis clock period drops down to 5 ns or so, I seem to remember we'd have trouble meeting timing doing the multiply and accumulate inside the for-loop... Any ideas what fixed that?? Are you still using aggressive clock periods? I wonder if including the pipeline "rewind" pragma indicates to the compiler that it can keep the II low while continuing work on previous iterations? Very curious!

A couple other PR-related thoughts: I'm having a hard time understanding the code as-is right now. I suggest the divergence between the "low_latency" version and the improved version is a good candidate to split into two sub-functions (I expect this should be allowed based on the CONFIG_T input without penalty; see my original version of the nnet_layer.h file for an example) -- it could make the code a lot more intelligible if we intend to support both operating modes. Second, with these improvements, I'm very amenable to removing serial mode as we know it today -- I'd still like to support a top-level interface that can stream data in and out rather than requiring a batching operation outside the logic core, but I dont think I see a reason why we shouldnt just accumulate a chunk of data inside the HLS core and then use the parallel mode with appropriate reuse factors.

Overall, super excited to see this progress!! I consider these improvements to be "HLS best practices" I'll try to learn from for future work.

nhanvtran:
ok, @violatingcp and I have cleaned up the idea with a near-final proposal.  See the slides linked to this agenda:
https://indico.cern.ch/event/794756/

The latest is in the branch `nvt/large_mlp_wDep_v4`

This works generically for large networks and large reuse factors.  Feedback welcome.  We'll keep working to get an implementation for a fully integrated PR

vloncar:
Merged long time ago as part of hack6 branch

PR #131, opened 2019-02-11
Use / as a separator on Windows, to avoid problems with tcl

vloncar:
It appears the issue on Windows is due to the way tcl interprets paths, where `..\..\nnet_utils` gets broken because `\n` becomes a newline.

### DIFF
--- a/hls-writer/hls_writer.py
+++ b/hls-writer/hls_writer.py
@@ -882,6 +882,8 @@ def hls_writer(layer_list, yamlConfig):
 
     nnetdir = os.path.abspath(os.path.join(filedir, "../nnet_utils"))
     relpath = os.path.relpath(nnetdir, start=yamlConfig['OutputDir'])
+    if os.name == 'nt':
+        relpath = relpath.replace('\\','/')
 
     f = open(os.path.join(filedir,'../hls-template/build_prj.tcl'),'r')
     fout = open('{}/build_prj.tcl'.format(yamlConfig['OutputDir']),'w')


nhanvtran:
@richarao any progress here?

A related note: since you have set up Windows, can we also document on the gitbook if any changes are needed for setting up the conda environment?

richarao:
@nhanvtran I've been working on this to check if a separate change is required for windows or if this change would be compatible with other systems as well. Based on this, I would be updating the gitbook.

nhanvtran:
@richarao perfect, thanks! keep us posted

Sally-Islem:
I tried to integrate the path of the nnet_utils folder into the environment variables of Windows. Also I added the nnet_utils folder in the vivado_hls project and I put each time its location in each header (exp: #include "nnet_utils / nnet_layer.h" #include "nnet_utils / nnet_conv.h" #include "nnet_utils / nnet_conv2d.h" #include "nnet_utils / nnet_batchnorm.h" .....)

But I do not have the same error "gcc.exe: error: net_utils: No such file or directory"
sans nom recupere

Can you help me with more details !!

richarao:
@Miss-Sally Can you tell me which example file you're running? I tried few keras examples and its working alright. I'll try to reproduce your error and work on it.

Miss-Sally:
**@Richaro** I worked on keras to hls  folder   runing the examples "KERAS_1layer","KERAS_conv2d_model" ,"jetTagger_Conv2D_Small" and others but always I have the same error in HLS  gcc.exe :net_utils:no such file or directory
( I installed miniconda on windows and the dependencies requiered and I just run the project bellow)

vloncar:
@Miss-Sally Does the proposed fix of this PR not work for you? You are essentially getting the same error, but now instead of files, it can't find the folder. What environment variables did you try to set?

Miss-Sally:
@**vloncar** I tried to modified the file hls_writer.py as it defined bellow and I fixed the path of nnets_utils to the environment variables of windows  and also I integrate the nnet_utils to the hls  project

Miss-Sally:
I set the location of the nnet_utils folder to the environment variables of
my windows system

<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Garanti
sans virus. www.avg.com
<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

richarao:
@vloncar I was trying to re-produce the error being mentioned above, but now i'm into a different one. 
image

vloncar:
Hi @richarao,

There are two issues here:
```exception_ptr.h:132:13: error: unknown type name 'type_info'```
and
```nested_exception.h:122:61: error: redefinition of default argument```

Both look like bugs in the C++ standard library that ships with Vivado, but they only happen when we use C++11 features (for pooling layers) in clang compiler (which is used for synthesis). I think this is solved in newer versions of the library, but even the latest Vivado ships with old versions.

We can't solve both issues by changing our code, but we can address the issues by patching the header files of the standard library, located at in the Vivado installation directory, e.g.,:
```
C:\Xilinx\Vivado_HLS\2017.2\win64\tools\clang\include\c++\4.5.2
``` 
We can solve the first one by adding the missing declaration in `exception_ptr.h` (line 69) like:
```
#ifdef __clang__
    class type_info;
#endif
```
The second one is to remove `= 0` from line 122 in `nested_exception.h` so that the line reads:
```
__throw_with_nested(_Ex&& __ex, const nested_exception*)
```

With these changes both simulation and synthesis work for me.

@nhanvtran Maybe we can raise this issue with Xilinx people and hope this gets solved in a future version of Vivado?

nhanvtran:
sounds good -- let's discuss in our meeting tomorrow and raise it with Xilinx people.  we can ask the right person to send this to

Miss-Sally:
@nhanvtran 
@richarao
Hope that it will be solved

yosukeueda33:
My version is 2019.1.
At this version, it seems [this](https://github.com/hls-fpga-machine-learning/hls4ml/pull/131#issuecomment-467722142) modification doesn't need.  
With only [this](https://github.com/hls-fpga-machine-learning/hls4ml/issues/121#issuecomment-512710828) modification, both simulation and synthesis finished with 0 errors.
FYI.

Miss-Sally:
Hi ,great but is there an advancement compared to the application of a deep
neuron network like u-net where you already have the file json and h5


Le lun. 22 juil. 2019 à 02:14, yosukeueda33 <notifications@github.com> a
écrit :

Issue #133, opened 2019-02-23
Precision in batch norm


benjaminkreis:
@sergojin, @jmduarte, and I found that doing the batch norm computation and casting to `res_T` in one line leads to poorer agreement with floating point than expected.

Replacing this:
```
res[ires] = (res_T) (data[ires]-mean[norm_index])*scale[norm_index]+beta[norm_index]
```

with 
```
ap_fixed<32,14> temp = (data[ires]-mean[ires])*scale[ires]+beta[ires];
res[ires] = (res_T) temp;
```
where ap_fixed<32,14> is some arbitrarily chosen high precision leads to good agreement between `res` and the floating point expectation.  In other words, the casting is not just changing precision at the end of the calculation but is instead influencing the precision of the calculation itself (maybe not a surprise).

It's a small change, but we need to think of something better than just hardcoding a random high precision type for `temp`.

holzman:
What types were data_T and res_T in this particular example?

benjaminkreis:
IIRC, we were using ap_fixed<18,8> or <16,6>.

sergojin:
Correct. <18,8> in that particular case.

holzman:
Your original code isn't quite doing what you think it is doing.

```
res[ires] = (res_T) (data[ires]-mean[norm_index])*scale[norm_index]+beta[norm_index]
```

C-style casts have higher precedence than arithmetic operations, so this is casting (data-mean) to res_T but not scale or beta. If you want to cast the whole expression, add another set of parens, or use static_cast.

benjaminkreis:
@sergojin, do you remember if we tried adding the parenthesis? Do you have the numbers on hand that showed the problem?  That would make it quick to test.

PR #135, opened 2019-03-02
batch norm precision fix

benjaminkreis:
Fixed Issue #133 caused by casting having higher precedence than expected.  This PR removes the casting (adding parentheses would also work).

### DIFF
--- a/nnet_utils/nnet_batchnorm.h
+++ b/nnet_utils/nnet_batchnorm.h
@@ -85,10 +85,10 @@ void normalize(
             #pragma HLS UNROLL
             #pragma HLS PIPELINE
         }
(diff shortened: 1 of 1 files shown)


jngadiub:
@benjaminkreis I was following issue #133 but I haven't fully understood how this problem came up as all those variables are typically all set to the same precision as res_T. Anyways, I will merge the PR as I understand that this is in general the correct way to do what we want: perform the cast on the result of the arithmetic instead of doing a partial casting as we were doing before.

benjaminkreis:
@jngadiub, I haven't tried with @sergojin's numbers, but it could be this -- if `data` and `mean` have the same precision, the result of `(data[ires]-mean[ires])` needs one more bit to represent the full range of possible values.  So you could imagine a case where you overflow in the early truncation, where you wouldn't if you truncated after multiplying by a scale number with magnitude less than 1.

Issue #120, opened 2018-11-20
Restructuring hls writer


benjaminkreis:
Here we can discuss restructuring [hls_writer.py](https://github.com/hls-fpga-machine-learning/hls4ml/blob/master/hls-writer/hls_writer.py).  Now that we are supporting so many layer types this code has gotten rather inelegant.  

I chatted with @jmduarte and @thesps recently, and here are a few initial thoughts:

1) hls_writer expects a list of python dictionaries that hold the layer info extracted in keras-to-hls or pytorch-to-hls.  For @sergojin's model with the output of one layer feeding into two following layers (#104), we should add to this dictionary where the inputs are coming from rather than assume the ordering in the json.  @jmduarte made the nice suggestion of making the list of dictionaries a dictionary of dictionaries so that we can access layers by a key such as an index or name rather than an index in a list.  This will make it easier to find the input to a layer.

2) Currently hls_writer's outer loop is through the lines of template C++ files.  It has an inner loop over layers at each place it needs (e.g. when including the weight files, when adding the layers in the top level, etc.).  Perhaps it would be better to make the outer loop over the layers (ie the dictionary of layer dictionaries) and append python objects holding the C++ code (which we would dump to files at the end).  @jmduarte points out that this would make adding new layer types easier because development would be more confined to one section of code.    

3) A lot of the confusing logic in hls_writer is in the sharing between layers of the `#define` parameters in parameters.h.  This makes the generated C++ code easier for a human to read, but is making hls_writer harder to write because we have to write out every possible combination.  For example, when making a dense layer, we check which layer came before to decide if we want to write n_in as a function of a previous Conv2D output (`n_in = OUT_HEIGHT*OUT_WIDTH*N_FILT`), a previous Conv1D output (`n_in = Y_OUTPUTS*N_FILT`), or a previous dense output (`n_in = N_LAYER_1`).   Maybe we can still use `#define` parameters set by info only from the present layer dictionary (assuming this can be extracted from the json and h5),  sometimes leading to redundant ones created in the next.  This would keep some of the readability but avoid having to code the logic to explicitly handle every possible combination in hls_writer.

zhenbinwu:
For 1, this is partially started for MLP， support Pytorch,  at https://github.com/hls-fpga-machine-learning/hls4ml/tree/zw/addPytorch

vloncar:
[...]

Issue #104, opened 2018-10-02
multiple layers running in parallel on same input


benjaminkreis:
@sergojin has a use-case for multiple dense layers running in parallel on the same input and producing multiple outputs. 

Two example keras models are here:
https://github.com/hls-fpga-machine-learning/hls4ml/tree/multiple_layers/keras-to-hls/fromSergo

And a working HLS project made by hand from the ".5" model is here:
https://github.com/hls-fpga-machine-learning/hls4ml/tree/multiple_layers/keras-to-hls/my-hls-test-modified
The final layers run in parallel on the output of the previous layer, and their output is merged to form the result.

What is needed is the hls4ml translation part.  Right now we assume the output of each layer is input to only one layer, with the order taken from the order in the json file.  @sergojin and @nhanvtran found we can use the `inbound_nodes` of the json to map the layers to each other.

benjaminkreis:
Note `inbound_nodes` doesn't exist for Sequential models

Issue #70, opened 2018-07-05
Concatenate layer


jmduarte:
Add support for `Concatenate` layer type:

https://github.com/keras-team/keras/blob/master/keras/layers/merge.py#L320

vloncar:
I have been looking at the the possible ways to implement this, however, other than the simple approach demonstrated in `nnet::merge` function which is used in sublayer project the other approaches don't work nicely with HLS. Ideally, we would have a `nnet::concatenate` function which takes a list of arguments, but when creating a list of existing arrays, e.g., for 1D arrays:

```
layer_t logits_0[N_LAYER/2];
layer_t logits_1[N_LAYER/2];
layer_t (*logits_list[2])[N_LAYER/2] = {&logits_0, &logits_1};
```
HLS  synthesis complains that using `&` results in an unsupported access pattern and thus the arrays cannot be partitioned (which is bad). Using pure pointer approach has the same problems.

Another approach I tried is to use function with variadic arguments (similar to C `printf`), but this crashes the synthesis process (it segfaults on a line where I access a pointer through `va_arg`).

This leaves only the simple approach with multiple functions with different number of arguments. So we would have one `nnet::concatenate` which concatenates 2 arrays, one for 3 arguments etc. The same for other Keras merge layers and (eventually) the same 2D, 3D...

Handling this in keras-to-hls is not a problem, so I am thinking about doing it this way even though it is not perfect. Thoughts?

benjaminkreis:
Hi @vloncar, thanks for investigating this!  I think going with the `nnet::merge` is good.   Maybe not the most beautiful, but simple is good for HLS.

PR #136, opened 2019-03-08
HLS writer (& others) rewrite

vloncar:
This includes a major rewrite of most components of the translation, discussed in #120. It introduces a new object model in `hls_model.py` which will (hopefully) simplify adding support for new nets/layers. As a result, `hls_writer.py` is now a significantly simpler.

The changes include support for multiple inputs/outputs (#104), move to flattened arrays everywhere, treating of all activations as layers and I also removed sublayer support (in anticipation of a better alternative). As a result, `keras-to-hls.py` has also seen some cleanups.

I have tested it and it produces the same results in C simulation and same usage/latency numbers in synthesis.

I am sorry this is all one big changeset, but it was very hard to track all upstream changes and rebase every time, so I have mostly developed this as separate files which were then just renamed and placed over existing ones in the end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 __pycache__
 vivado_prj
+.vscode
--- a/example-prjs/conv-1layer/firmware/myproject.cpp
+++ b/example-prjs/conv-1layer/firmware/myproject.cpp
@@ -76,7 +76,7 @@ void myproject(
     //Dense
     result_t logits2[N_OUTPUTS];
     #pragma HLS ARRAY_PARTITION variable=logits2 complete dim=0
-    nnet::compute_layer<input_t, result_t, config2>(layer1_out_flat_relu, logits2, w2, b2);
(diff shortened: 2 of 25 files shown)


vloncar:
Added initial support for Keras [merge](https://keras.io/layers/merge/) layers (#70). Limited to layers with 2 inputs. `Concatenate` layer supports only up to rank-3 tensors.

benjaminkreis:
This looks great to me!  I think we can merge, but I will also let @holzman comment as the other reviewer.

vloncar:
Thanks @benjaminkreis!

@zhenbinwu you also expressed interest in this. Do you have time to take a look?

holzman:
I personally haven't had time to review (caught the flu at the HOW meeting and am still not 100%) -- so let's not wait for me on this one

vagos06:
In lines 247-249 in concatenate3d, there are calls to n_elem1[0]/n_elem2[0] in error (should be n_elem1_0/n_elem2_0, respectively).

PR #137, opened 2019-04-08
fixed keras-to-hls.py to support newer keras versions

jngadiub:
Found differences in the model json file for different keras versions (2.0.5 vs 2.2.4). The current keras-to-hls.py supports the older version with which the json files in the example folder were created. This PR adds a simple fix to keras-to-hls.py proposed by @vloncar  to support both versions.


vloncar:
Looks good! Thanks Jen.

PR #182, opened 2020-03-04
Adding GarNet and introducing some changes to the core to make GarNet possible

yiiyama:
Adding support to a graph neural network (GarNet). The compatible keras implementation of GarNet is not the [original](https://github.com/jkiesele/caloGraphNN/blob/master/caloGraphNN_keras.py) but this [temporary-ish code](https://github.com/yiiyama/graph-hls-paper/blob/master/keras/layers/simple.py). We will discuss how the new implementation can be included in the "official" repository.

hls4ml/templates/vivado/nnet_utils/nnet_garnet.h is the source code for HLS GarNet. Aside from adding this file and inserting lines to convert / configure in keras_to_hls.py, hls_model.py, vivado_template.py, and vivado_writer.py, I introduced some changes to the hls4ml core software in order to accommodate the needs of graph networks:

- keras_to_hls.py was written with a linear model (single input, single flow of data through layers) in mind. Particularly, layers being parsed were communicating the shapes of the input and output arrays through the `current_shape` variable. Since GarNet takes two inputs (array of vertex features and the number of vertices in the given sample), I replaced this communication mechanism with more explicit `input_shapes` and `output_shapes` variables. There is still an assumption that each layer outputs a single array, but it is easy to generalize that if a need arises.
- In hls_model.py and vivado_writer.py, I added some more handles to the models and layers that can be steered in the config yml file:
    - GlobalPipelining (model property): By default, the full model is either PIPELINEd or DATAFLOWed, depending on the IOType and Strategy parameters. GlobalPipelining config allows directly setting or turning off (set to None) pipelining.
    - OutputPartitioning (layer property): By default, layer outputs are partitioned completely. Graph network outputs are large arrays, so this would be a problem. To specify a partitioning, add a line like `OutputPartitioning: partition,cyclic,4` (translates to `ARRAY_PARTITION cyclic factor=4`) under the LayerType or LayerName config to set the partitioning type. (GarNet layers automatically set their output partitioning using the reuse_factor, so users actually don't need to use this config for anything at the moment.)
    - CustomizeIO, InterfaceMode (model properties): Input and last layers can have custom partitioning using the OutputPartitioning config. Set CustomizeIO to 1 to have the model I/O actually be partitioned accordingly. InterfaceMode allows setting the value of `HLS INTERFACE` pragma by hand.
- In vivado_templates.py and templates.py, I added the "include list" feature so that only the header files that are actually used in the given model are included. Whenever we add a layer, we need to specify which headers are to be included in vivado_templates.py.
- C++ template reorganization: I wanted parameters.h to include the weights for GarNet (so that the config structs can refer to the weight arrays). To make this happen, header inclusions in the C++ templates have been shuffled:
    - Added a new file defines.h that has the `hls-fpga-machine-learning insert numbers` and `hls-fpga-machine-learning insert layer-precision` lines (originally in parameters.h)
    - Line `hls-fpga-machine-learning insert weights` is now in parameters.h
    - myproject.h includes defines.h
    - myproject.cpp includes parameters.h
    - myproject_test.cpp does not include parameters.h any more (it only needed the number macros and typedefs, which are now in defines.h and are included through myproject.h)

It's a long list of feature updates - I can split it into multiple PRs if that's more convenient.


vloncar:
Many thanks!

I haven't looked into the whole set of new changes, but here are some thoughts on the PR that was made in my fork:
I don't see what `GlobalPipelineing` config property adds to the project. `Strategy: Resource` is DATAFLOW, `Strategy: Latency` is PIPELINE, so with `GlobalPipelining` there are two ways to configure the same thing, and one is only for GarNet and does nothing for every other layer.
`OutputPartitioning` has undesired consequences. We cannot just expose that and allow arbitrary stuff. If you change from complete partitioning to anything else you'll very likely mess up pipelining and end up with large latency and II larger than RF. I think we should explore every possibility of figuring the best possible partitioning automatically based on `ReuseFactor` and `Strategy` parameters before giving up and exposing it to users (because they will be even more clueless then us).
We are thinking of removing `io_serial`, so the `InterfaceMode` would be a good replacement for one functionality that `io_serial` used.
`templates.py` existed by mistake, forget about it. Only `vivado_templates.py` is relevant.
The changes in `keras_to_hls.py` look ok, but so far we have somehow managed to support multiple flows/inputs/outputs. Keras orders the layers in such a way that when parsing JSON sequentially, you are always working with one flow, so you will always encounter an `InputLayer` or a layer with 'inbound_nodes` when switching flows, thus `current_shape` ends up being good. Perhaps this is something you can improve in serialization of GarNet layer itself.

The restructuring of `parameters.h` wasn't there before, so I haven't yet peeked into that, but I agree with these changes conceptually.

jmduarte:
Hi @yiiyama,

Can you add an example GarNet keras model and json to the [example-models/keras](https://github.com/yiiyama/hls4ml/tree/graph-hls-paper/example-models/keras) directory to make it easier to test? 

Eventually, we can also add it to the default tests too.

yiiyama:
@jmduarte Will do! @vloncar wanted separate PRs for different features I introduced here, so I'm closing this one and will have several feature proposal PRs plus one GarNet PR on top, and will include an example in that one.

PR #184, opened 2020-03-16
making input and output shapes explicit to support nonlinear models

yiiyama:
Replacing the variable `current_shape` with more explicit `input_shapes` and `output_shapes` variables. This adds the flexibility needed for some layers that take multiple input tensors. The `output_shape` of each layer is inserted to the `output_shapes` dict at line 432 (for skip layers in line 167-170) with the name of the layer as the key. The `inbound_nodes` parameter of the layer config in JSON is given in terms of the layer names, so we can look up the shapes from this dict to set the `input_shapes` list (lines 154-157).


yiiyama:
Ah! Didn't think about that. Sure, I'll figure out how to support both.

PR #185, opened 2020-03-16
Vivado template restructure

yiiyama:
- C++ template reorganization: I wanted parameters.h to include the weights for GarNet (so that the config structs can refer to the weight arrays). To make this happen, header inclusions in the C++ templates have been shuffled:
    - Added a new file defines.h that has the `hls-fpga-machine-learning insert numbers` and `hls-fpga-machine-learning insert layer-precision` lines (originally in parameters.h)
    - Line `hls-fpga-machine-learning insert weights` is now in parameters.h
    - myproject.h includes defines.h
    - myproject.cpp includes parameters.h
    - myproject_test.cpp does not include parameters.h any more (it only needed the number macros and typedefs, which are now in defines.h and are included through myproject.h)
- "include list": Making the list of included nnet_utils files model-specific.
    - Added an attribute `include_list` to the class `Layer` in hls_model.py
    - The layer include list is set in templates/vivado_template.py (with functions defined in templates/templates.py)
    - Added a template line `hls-fpga-machine-learning insert includes` to templates/vivado/firmware/parameters.h
    - keras_writer.py collects the layer include lists, uniqify them, and inserts the list to this template line in parameters.h
- Array partitioning and reshape: Allowing layers in hls_model.py to specify how they want to partition / reshape the input and output arrays.
    - I suppose we don't want the vivado-specific `#pragma HLS ...` lines in hls_model.py any more (lines 375-393).
    - Instead we should just set the `pragma` attribute of each ArrayVariable object to simple keywords of `'partition'`, `'reshape'`, or `'stream'`, or to a tuple (`'partition/reshape'`, `'complete/cyclic/block'`[, factor]) in case of incomplete partitioning or reshaping.
    - vivado_writer.py then should pick this config up and form the `#pragma HLS ...` lines (added a helper function `_make_array_pragma`).
- hls_model.py made compatible with multiple input layers: lines 679-685. Type for _i_th input array (_i_ != 1) is named `input[i]_t`
- Cosmetics in myproject_test.cpp: Using `std::copy` and `std::fill_n` to set the input and output array values

### DIFF
--- a/hls4ml/writer/vivado_writer.py
+++ b/hls4ml/writer/vivado_writer.py
@@ -60,6 +60,13 @@ class VivadoWriter(Writer):
 
     @staticmethod
     def _make_array_pragma(variable):
+        """
+        Layers in hls_model.py can specify output array partitioning through the `pragma` attribute.
+        If `pragma` is a string: options are 'partition', 'reshape', or 'stream'.
+        If `pragma` is a tuple: (mode, type, factor) where mode is 'partition' or 'reshape', type is
+        'complete', 'cyclic', or 'block', and factor is an integer only used when the type is not 'complete'.
+        """
+        
         config = variable.pragma
         if type(config) is tuple:
             mode = config[0]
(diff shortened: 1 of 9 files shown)


hls4ml/writer/vivado_writer.py
```diff
@@ -293,24 +348,26 @@ def write_test_bench(self, model):
                 newline = line.replace('myproject', model.config.get_project_name())
             elif '//hls-fpga-machine-learning insert data' in line:
                 newline = line
+                newline += '      std::vector<float>::const_iterator in_begin = in.begin();\n'
```
vloncar: if you want `const_iterator` shouldn't you use `cbegin()`?

(I know I'm nitpicking here, but I have to comment on something or it will look like I didn't even go through the code :) )

yiiyama: Haha sure, [begin() also returns a `const_iterator`](https://en.cppreference.com/w/cpp/container/vector/begin) but indeed `cbegin` would be more explicit.

ngiambla:
This PR ensures that we are including only _used_ headers:
Shouldn't we only copy over the used headers as well?
```
    def write_nnet_utils(self, model):
        ###################
        ## nnet_utils
        ###################

        filedir = os.path.dirname(os.path.abspath(__file__))

        srcpath = os.path.join(filedir,'../templates/vivado/nnet_utils/')
        dstpath = '{}/firmware/nnet_utils/'.format(model.config.get_output_dir())

        reqheaders = ['nnet_common.h','nnet_array.h','nnet_helpers.h']

        if not os.path.exists(dstpath):
            os.mkdir(dstpath)

        headers = [os.path.basename(h) for h in glob.glob(srcpath + '*.h')]

        for h in headers:
            if "nnet_utils/"+h in sorted(set(sum((layer.include_list for layer in model.get_layers()), []))):
                copyfile(srcpath + h, dstpath + h)
        for reqh in reqheaders:
            copyfile(srcpath+reqh, dstpath+reqh)
```

vloncar:
@ngiambla headers in `nnet_utils` may may include each other.

ngiambla:
Ah! Do you think it's worth it to isolate each header (or list out it's dependencies) and then we can include things we _only_ need?

vloncar:
Isolating is a bad idea for many reasons (most obvious one being unnecessary duplication), and tracking the dependencies of each would add complexity that I don't think is worth it. I don't see what we would save, the files that are not used have no impact on synthesis and they don't take up noticeable amount of storage space.

ngiambla:
Ah okay! I figured out this could be added for sake-of-completeness (since this PR introduces layer-specific includes) Thanks for your input! :smile:  If you do wish to add this:

You can add the dependencies here (hls4ml/templates/vivado_template.py):

```
dense_include_list = ['nnet_utils/nnet_dense.h', 'nnet_utils/nnet_dense_compressed.h', 'nnet_utils/nnet_dense_large.h']
batchnorm_include_list = ['nnet_utils/nnet_batchnorm.h']
conv1d_include_list = ['nnet_utils/nnet_conv.h', 'nnet_utils/nnet_conv_large.h']
conv2d_include_list = ['nnet_utils/nnet_conv2d.h', 'nnet_utils/nnet_conv2d_large.h']
activ_include_list = ['nnet_utils/nnet_activation.h']
pooling_include_list = ['nnet_utils/nnet_pooling.h']
merge_include_list = ['nnet_utils/nnet_merge.h']
resize_include_list = ['nnet_utils/nnet_image.h']
transpose_include_list = ['nnet_utils/nnet_array.h']
```

PR #188, opened 2020-04-10
Adding the GarNet layer and an example project

yiiyama:
Original keras layer: https://github.com/jkiesele/caloGraphNN/blob/master/caloGraphNN_keras.py

Aside from GarNet-specific templates, converter codes, and hls_model.py classes, this PR introduces the following changes:
- In `hls_model.py`, a new variable type name is introduced when `Layer.add_weights_variable` is called with a specified `precision`, even when the `type_name` argument is None. Otherwise repeated calls with `precision` not None but `type_name` None will keep using the default type name and overwriting the precision specified in the previous call.
- In `templates.py` and `vivado_template.py`, added a method `get_precision` to the `Backend` class. The method is called with the general type of the variable ('f' or 'i') and the number of binary digits, and returns the string corresponding to the precision for the given backend. `GarNet` and `GarNetStack` classes in hls_model.py use this method to avoid explicitly mentioning `ap_fixed` and `ap_int` types.


vloncar:
Awesome work @yiiyama ! Can you also add a json+h5 and yml to the PR so that the conversion can be tested?
I'll do the proper review next week (hopefully)

yiiyama:
Sure! Should I make a directory in example-models to put the files in or can I just tuck them in the example project directory?

Also I forgot to mention one important thing - nnet_garnet.h doesn't synthesize with at least Vivado 2018.1. I've been developing on my personal machine which has 2019.2 installed so I didn't notice this until today. Apparently the older version of Vivado doesn't like having pointer / reference static const class members, which is used a lot in garnet_config (see vivado_templates.py).

vloncar:
If the files are not too large you can put them there. If the result ends up being the same as the project in `example-prj` maybe the example could be removed.

In 2019.1 they finally updated the compiler, so maybe that was the issue.

yiiyama:
I `reset --hard` to remove the example project and instead added the weights and json files to example-models. Because I wanted example `tb_input` files to be there too, those are included in the example model directory.

jmduarte:
@yiiyama could you add an example model (ideally from the paper?) in a PR to the new [example-models](https://github.com/hls-fpga-machine-learning/example-models) repo?

also, could you rebase this PR so it's mergeable?

@vloncar any other comments for this to be merged?

I'm also happy to test the example model to confirm everything is working as expected.

yiiyama:
Superseded by #213.

PR #193, opened 2020-05-26
Keras LSTM and GRU implementation

richarao:
This is the full implementation of LSTM and GRU models converted from Keras.


vloncar:
Continued in #329

PR #203, opened 2020-07-02
Fix two issues for the concatenate layer

melko:


vloncar:
Awesome!
Do you have an example model which demonstrates the improvement? Like a minimal example...

melko:
Hi,
I saw the issue on a model I'm trying, which implements several convolutional layer, then concatenating outputs before a dense layer. I can share that one or, if you prefer, I can make a minimal example with just a concat layer.

vloncar:
I would prefer a minimal example because we can include it in the future test suite (currently being revamped). You can just share a code gist somewhere, that would be fine

melko:
Here's a minimal example. which concatenates an input with itself:
https://gist.github.com/melko/51411e1f952038dfaadd4669ad960388

PR #194, opened 2020-05-27
Update for PyPI

jmduarte:
See `hls4ml` on the test server here: https://test.pypi.org/project/hls4ml/

(Note some of the badges don't work because it's not on the official PyPI yet)

PR #205, opened 2020-07-12
Documentation for new API functionalities, and possibly PyPI and example models

Duchstf:
New doc for PR #195 And possibly #194 too. There have been a lot of changes, so the documentation might be restructured a bit. 

You can preview the new doc here (still incomplete), any comments/suggestions are welcomed:

https://duchstf.github.io/hls4ml/

**HELP NEEDED!!** If anyone can help with updating the doc it would be great 😬. There are a lot of small moving pieces here.


thesps:
Thanks a lot @Duchstf! It looks really nice.
I have a few small changes I would make, and also I'll contribute the 'Quantization' section. So I'll make a PR to your branch soon.

ngpaladi:
Looks great! Have we considered a move to a pure Jekyll theme? That way this could move the documentation into a `doc/` folder in the main branch (telling Github pages to pull from that) and bundle it with source code downloads while omitting extra files (other than a `_config.yml`)?
I'm also happy to help out with updating.

Duchstf:
@ngpaladi Hey, I don't really understand what you're trying to say. So you mean move all the docs to `doc/` in `master`? and then tell Github pages to pull from that? 

I think generally we shouldn't try to complicate things (so that everyone can easily contribute to the doc). I think the way we're doing it right now is very straightforward (just write the markdown file) and user-friendly. 

My only problem with the current doc is the code & text formatting is not as nice as, say, Keras or Tensorflow's docs. So if you know how to improve on that in Gitbook it would be great.

ngpaladi:
Jekyll is basically the same, it renders markdown files just as you describe. In fact I wouldn't be surprised if Gitbook uses it in their backend. For a quick demo see [this](https://github.com/ngpaladi/py2700) python package I maintain. The code for the [documentation](https://www.noahpaladino.com/py2700/) is in the `docs/` folder in `master`. The Jekyll theme can be customized as needed and written independently of the documentation, as it is in this case.

Duchstf:
@ngpaladi Hi, thanks for your interest in helping out. 2 things:

1. I don't see the urgent need to switch to a pure Jekyll theme right now, but it's an interesting option to keep in mind as the project gets larger. 

2. The [concepts page](https://fastmachinelearning.org/hls4ml/CONCEPTS.html) is still under construction. If it's possible for you, it would be interesting to re-write the [original paper](https://iopscience.iop.org/article/10.1088/1748-0221/13/07/P07027) in a blog-post form, so that `hls4ml` is more approachable to a wider audience. It'll also help you to understand `hls4ml` better to help out with code development. 

Duc.

ngpaladi:
1. Sure, that's fine. I will probably be reworking the theme I'm using now anyway, so if there are any requests I can implement those.

2. I will work on this. Would you prefer it be completed before this is merged?

Duchstf:
@ngpaladi Hey, you can just work on it and make a PR to `hls-fpga-machine-learning:gh-pages-2`. It doesn't need to be included in this PR.

Issue #204, opened 2020-07-11
CSIM Inference Output Discrepancy


veyron8800:
Hello,

I am running into an issue where HLS4ML model inference output from CSIM does not match what is produced by using tf.keras. I have compiled the model into [this repository](https://github.com/veyron8800/DigitClassification), so that the bug can be easily reproduced.

The model in question is a simplistic CNN with the following architecture:

```
model = models.Sequential()
model.add(layers.Conv2D(4, (3, 3), activation='relu', input_shape=(28, 28, 1)))
model.add(layers.MaxPooling2D((2, 2)))
model.add(layers.Conv2D(8, (3, 3), activation='relu'))
model.add(layers.MaxPooling2D((2, 2)))
model.add(layers.Conv2D(16, (3, 3), activation='relu'))
model.add(layers.Flatten())
model.add(layers.Dense(16, activation='relu'))
model.add(layers.Dense(10, activation='softmax'))
```

Unfortunately, the inference output on the test image set for the [csim of the hls4ml](https://raw.githubusercontent.com/veyron8800/DigitClassification/master/InferenceComp/HLS4ML_Out.dat) model is nowhere close to what is produced by [tf.keras](https://raw.githubusercontent.com/veyron8800/DigitClassification/master/InferenceComp/TF_OUT.dat). The thing that really caught my eye is the [hls4ml output](https://raw.githubusercontent.com/veyron8800/DigitClassification/master/InferenceComp/HLS4ML_Out.dat) is that there are values that exceed the bounds of [0,1], when the final layer has a softmax activation. This suggests to me that the problem is most likely with the [cpp definition of the softmax function](https://github.com/hls-fpga-machine-learning/hls4ml/blob/e9d057618d875f0e75015786b315d5fef63e6e84/hls4ml/templates/vivado/nnet_utils/nnet_activation.h#L174).

Here are some more links to the example repo for convenience:
Model creation and training: [ModelTraining.py](https://github.com/veyron8800/DigitClassification/blob/master/ModelTraining.py)
Config file: [DC.yml](https://github.com/veyron8800/DigitClassification/blob/master/DC.yml)
Inference output generation: [InferenceComp.py](https://github.com/veyron8800/DigitClassification/blob/master/InferenceComp.py)
Inference output: [InferenceComp](https://github.com/veyron8800/DigitClassification/tree/master/InferenceComp)

Thanks

Duchstf:
Hey, 

[...]

PR #207, opened 2020-07-24
Stable softmax

thesps:
I've a bit more for the Softmax implementation. You may recall there was a new version introduced with PR #195.
In issue #204, it was observed to not be working that well, possibly resulting in unphysical values (out of the range {0,1}) and giving incorrect predictions.

After some more experiments, I've done a few things:
- Changed the default type (when using `hls4ml.util.config`) to include the `AP_RND`, `AP_SAT` rounding and saturation modes. This bounds the values to the correct range
- Added a new 'stable' implementation (documented below)
- Added a config parameter to select between implementations
- Based on the above parameter, added back the implementation used until v0.2.0 (now called 'legacy')

The new 'stable' version uses the same type-agnostic lookup tables as the recent new implementation. For the vector `x`, Softmax does: `y = exp(x) / sum(exp(x))`. In the code, lookup tables are used for `exp(x)` and `1 / exp(x)`. The new version does `y = exp(x - max(x)) / sum(exp(x - max(x)))`. This construction means that all the `x - max(x)` values are either `0` (when `x = max(x)`) or negative. This makes the lookup tables much nicer. The exponential table effectively saturates at the ~0 end rather than the +infinity end like just looking up `exp(x)`.

On the jet tagging dataset, the version in master already performs well, and the new stable implementation performs similarly (just showing 2 classes for clarity):
roc

The problem was seen in MNIST, where the accuracy is typically close to 100%. This means one class is typically predicted with a much greater probability than any other. Most of the time the implementation in `master` works fine, but due to the big range in input values, sometimes class predictions which are actually quite different get clipped to the same probability (~1). Then when taking the `argmax` to predict and evaluate accuracy, the wrong prediction can be made. The new 'stable' implementation resolves these nicely.
Here's the ROC just for one class:

ROC_mnist

It may look like it's still not doing that well, but the accuracy tells the story:

| Inference | Accuracy |
| --- | --- |
| Keras |   0.9774 |
| Master |   0.9720 |
| Legacy |  0.9689 |
| Stable |    0.9774 |

So the `argmax` prediction matches Keras every time, in this case. The ROC differs slightly because there's not quite so much precision on the hls4ml prediction as the Keras one. 

In terms of latency & resources, the new stable version is very slightly slower than either the current master, or 'legacy' versions due to having to find the `max(x)`: 6 cycles for those vs 8 for the stable in a standalone Softmax project for 10 classes (MNIST).

The resources are most similar to the current master implementation, so `~0%` for all types, but slightly more LUTs and FFs are used. Both these are smaller than the legacy implementation.

So based on the latency & resources I've set the default implementation to be the one in current master (called `latency`). 
The selection is done in the config like `cfg['LayerName']['softmax']['implementation'] = 'stable'` (options are `latency, stable, legacy`). We might want to make the default `stable` to 'just work' in more cases, but for low latency applications the `latency` version is probably best.

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -698,6 +698,8 @@ class Softmax(Activation):
                 self.set_attr('exp_table_t', self.get_attr('table_t'))
             if 'inv_table_t' not in self.attributes:
                 self.set_attr('inv_table_t', self.get_attr('table_t'))
+            if 'implementation' not in self.attributes:
+                self.set_attr('implementation', 'latency')
 
 class BatchNormalization(Layer):
     def initialize(self):
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -698,8 +698,11 @@ class Softmax(Activation):
                 self.set_attr('exp_table_t', self.get_attr('table_t'))
             if 'inv_table_t' not in self.attributes:
                 self.set_attr('inv_table_t', self.get_attr('table_t'))
-            if 'implementation' not in self.attributes:
+            if self.model.config.is_resource_strategy(self):
+                # 'resource' strategy = 'latency' for Softmax
                 self.set_attr('implementation', 'latency')
+            else:
+                self.set_attr('implementation', self.model.config.get_strategy(self).lower())
 
 class BatchNormalization(Layer):
     def initialize(self):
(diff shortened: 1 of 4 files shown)


thesps:
I think the 'legacy' (v0.2.0) above looks a bit worse than expected just because it's a log-log plot. Here's the logy plot we usually look at.
roc

PR #209, opened 2020-08-04
Update to the "Concepts" Page in Documentation

ngpaladi:
I've started work on summarizing the original hls4ml paper in the concepts page, mostly speaking to the motivation for the project thus far. I plan on adding more detailed information about the jet substructure application.


Duchstf:
Thanks, let me know if you need help. I think we should aim to make it very introductory & concise.

ngpaladi:
Okay. I'll just wrap up with a paragraph or two explaining the neural net implementation described in Figure 2 of the paper.

Duchstf:
Hey, I think what I have in mind is something like [this](https://towardsdatascience.com/illustrated-guide-to-lstms-and-gru-s-a-step-by-step-explanation-44e9eb85bf21) -- A step-by-step introduction with concise & intuitive sections.

ngpaladi:
I'll make some changes to get something more along those lines.

Duchstf:
Yes, this is what I was thinking of!! We might want to fix some parts but the outline is there!😃

thesps:
Thanks for contributing this! It looks like the math formatting isn't working to me?

CONCEPTS.md
```diff
@@ -1,19 +1,41 @@
...
+
+To ensure optimal performance, the user can control aspects of their model, principally:
+
+   * **Size/Compression** - Though not explicitly part of the hls4ml package, this is an important optimization to efficiently use the FPGA resources
+   * **Precision** - Define the precision of the calculations in your model
+   * **Dataflow/Resource Reuse** - Control parallel or serial model implementations with varying levels of pipelining
+
```
thesps, line 34: I think here we could add a link from the Precision bullet point to the PROFILING.md page. 
And maybe add another bullet afterwards like:

- Quantization Aware Training - achieve best performance at low precision with tools like QKeras, and benefit automatically during inference with hls4ml parsing of QKeras models

(It's kind of in the same vein as the Size/Compression bullet: something outside of hls4ml that benefits the inference performance)

CONCEPTS.md
```diff
@@ -1,19 +1,41 @@
...
+<br>
+<div style="text-align: center;"><img src="img/reuse_factor_paper_fig_8.png" width="500" /></div>
+<br>
+
+Often, these decisions will be hardware dependent to maximize performance. Of note is that simplifying the input network must be done before using hls4ml to generate HLS code, for optimal compression to provide a sizable speedup. Also important to note is the use of fixed point arithmetic in hls4ml. This improves processing speed relative to floating point implementations. The hls4ml package also offers the functionality of configuring binning and output bit width of the precomputed activation functions as necessary. With respect to parallelization and resource reuse, hls4ml offers a "reuse factor" parameter that determines the number of times each multiplier is used in order to compute a layer of neuron's values. Therefore, a reuse factor of one would split the computation so each multiplier had to only perform one multiplication in the computation of the output values of a layer, as shown above.
```
thesps, line 39: And here I think we could also link to the CONFIGURATION.md page which explains how to set `ReuseFactor`

thesps: And, after the explanation of "reuse factor of one", how about another sentence like "Conversely, a reuse factor of four (in this case) uses a single multiplier, four times sequentially. Low reuse factor achieves the lowest latency and highest throughput, using the most resources, while high reuse factor save resources at the expense of longer latency and lower throughput." ? What do you think?

ngpaladi: Sounds good, will include both suggestions

CONCEPTS.md
ngpaladi: Thanks for the suggestion, I'll add that

ngpaladi:
I made the requested changes. It seems that the LaTeX plugin for GitBook no longer works, and I tried a few other ways of loading the MathJax script that did not work. As such, I just redid the equations in html, which isn't ideal but doesn't look too terrible.

vloncar:
What is the status of this? Was it included in the new sphinx docs?

Duchstf:
Yes, sorry forgot to close this.

PR #210, opened 2020-08-05
Read InputLayer from Sequential models (TF 2.3)

vloncar:
Tensorflow 2.3 (Keras 2.4.0) now adds `InputLayer` to `Sequential` models, so we have to make sure we read it, otherwise two inputs will be created and HLS will mess up.


thesps:
Looks good to me, and I can confirm that without this a model trained with TF2.3 didn't work properly (inputs not propagated to layers), and this fixes it. So when the tests are done, we merge...

PR #215, opened 2020-08-22
Move pointer to include garnet test, update example fetching scheme

Duchstf:
Two main updates in this PR:

1. Move pointer to submodule to include the new garnet model test. 
2. Update fetching model scheme, now also support loading example data and configuration.


vloncar (approved):
Are we waiting for something else here or are we ready?

Duchstf:
@vloncar everything working now! I think it's ready to be merged!

PR #219, opened 2020-09-07
Fixes

thesps:
Two small fixes, spotted preparing notebooks for FPL tutorial.
Although, the fix to the config dictionary generator is quite important.

### DIFF
--- a/hls4ml/utils/config.py
+++ b/hls4ml/utils/config.py
@@ -70,7 +70,7 @@ def config_from_keras_model(model, granularity='model', default_precision='ap_fi
     conv_layers = ['Conv1D', 'Conv2D', 'BinaryConv2D']
     pooling_layers = ['MaxPooling1D', 'MaxPooling2D', 'AveragePooling1D', 'AveragePooling2D']
     norm_layers = ['BatchNormalization']
-    activation_layers = ['Activation', 'LeakyReLU', 'ThresholdedReLU', 'ELU', 'PReLU']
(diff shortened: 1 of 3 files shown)


hls4ml/utils/config.py
```diff
@@ -144,7 +144,7 @@ def make_layer_config(layer):
             layer_config['Precision'] = default_precision
             layer_config['ReuseFactor'] = default_reuse_factor
             layer_config['table_size'] = 1024
-            if layer['config']['activation'] == 'softmax':
+            if layer_config['activation'] == 'softmax':
```
vloncar, line 147: With the latest changes in JSON format, `Softmax` and `ReLU` are layers themselves (i.e., they can appear outside of other layers not just as `Activation` layers). They should be added to `activation_layers`

thesps: I've added that in the latest commit.

vloncar (approved):
Looks good!

thesps:
I added the new TF2.3 `Softmax` and `ReLU` layers to the supported list, and made sure the config generating utility treats both `Softmax` and `Activation(activation='softmax')` layers correctly. I did a quick test like this:
```
m = tf.keras.models.Sequential()
m.add(tf.keras.layers.Input((10,)))
m.add(tf.keras.layers.Activation(activation='softmax'))
m.add(tf.keras.layers.Activation(activation='relu'))
m.add(tf.keras.layers.Softmax())
m.add(tf.keras.layers.ReLU())
cfg = hls4ml.utils.config_from_keras_model(m, granularity='name')
hls_model = hls4ml.converters.convert_from_keras_model(m, hls_config=cfg)
hls_model.write()
```
And it all looks good.

PR #220, opened 2020-09-14
Streaming implementations of Conv2D and friends

vloncar:
This PR introduces stream-based implementations of most of the layers. The main idea behind the CNN layers implementation is that instead of bookkeeping a position within the kernel window that comes from sequential processing using streams we encode the instructions for all pixels of the input into an array. To keep the array size in check (so it is not as large as the input image itself), we encode the instructions for a minimal input that exhibits the same access pattern. For example, every convolution with a `3x3` kernel with `stride=1` regardless of input size can be represented with 25 unique instructions, corresponding to the minimal input size of `5x5`. See the following animation for a demonstration
conv2d_animation
(Input is at the top, output at the bottom. the top left corner contains the "pattern" for each pixel, while on the right we is the internal state during processing.)

To use it, just switch `IOType` to `io_stream`. This can be done through API (`convert_from_keras_model`) or in yml configuration.

Implemented layers:
- `Conv1D`, `Conv2D`
  - `same` padding is implemented by inserting a `ZeroPadding2D` layer before `Conv2D` (same for 1D)
  - 1x1 `Conv2D` will be handled in a special way, same for pointwise `Conv1D`
- `SeparableConv1D`, `SeparableConv2D`, `DepthwiseConv2D`
- `MaxPooling1D`, `MaxPooling2D`, `AveragePooling1D`, `AveragePooling2D`
- `GlobalMaxPooling2D`, `GlobalAveragePooling1D`
- `Dense`
  - Including QKeras variants
- `BatchNormalization`
- `ReLU` and `Softmax` activations
- `UpSampling2D`
  - Only `nearest` interpolation is supported
- `Add`, `Subtract`, `Multiply`, `Average`, `Maximum` and `Minimum`

`Conv1D/2D` and `Dense` leverage the existing `dense` HLS implementation, so `ReuseFactor` and `Strategy` work.

Also included are dozens of small improvements to the internal model and the overall codebase.

Current limitations:
- Square kernels (e.g., 2x2, 3x3...)
- Maximum stride is the width of the kernel (i.e., you can't skip over some pixels) except for 1x1 (pointwise) convolution
- No grouped or dilated convolutions

Earlier implementation introduced `PackingFactor` to cram more pixels into a single element of the stream. While this helps process the input faster it caused many issues when input and output packing were mismatched and increased the LUT/FF usage significantly. So it has been abandoned (for now). We are considering tiling as an alternative.

This is feature-complete and has been extensively tested by the CERN team so I consider it fairly stable and ready for a review. Some very minor changes and bugfixes may still come.

Questions, comments, praises and criticisms are welcome!


Duchstf:
I have some free time to take on the 1D stuffs if everyone agrees on this streaming implementation. 

Also, I know we kind of discussed this individually, but just an FYI for everyone, perhaps we should gather some common large CNN & datasets for testing and benchmarking as well. Since there is also another branch for streaming implementation:

https://github.com/violatingcp/hls4ml/tree/sc20_test  

For Conv1D I'll suggest [these models](http://kastner.ucsd.edu/blog/real-time-automatic-modulation-classification/). I also have another model that I've been playing with as well.

pierinim:
Hello 
I would move this discussion to slack/meeting since it concerns an already agreed strategy towards publishing our cnn results

Maurizio

thesps:
Could we get that latest commit 46edf5a "Handle no-bias fusing of Dense and BN" as a separate PR to master?

vloncar:
I've updated the text above to reflect the current state. I believe it is ready for review.

The tests should pass after #249 is merged.

hls4ml/__init__.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+__version__ = '0.5.0'
```
thesps, line 3: Bold 😉

hls4ml/converters/keras/core.py
```diff
@@ -102,6 +90,8 @@ def parse_activation_layer(keras_layer, input_names, input_shapes, data_reader,
         layer['activ_param'] = keras_layer["config"].get('theta', 1.)
     elif layer['class_name'] == 'ELU':
         layer['activ_param'] = keras_layer["config"].get('alpha', 1.)
+    elif layer['class_name'] == 'ReLU':
+        layer['class_name'] = 'Activation'
```
thesps, line 94: These are repeated a few lines below

hls4ml/templates/vivado/build_prj.tcl
```diff
@@ -65,6 +65,7 @@ if {$opt(reset)} {
   open_solution "solution1"
 }
 catch {config_array_partition -maximum_size 4096}
+config_compile -name_max_length 60
```
thesps, line 68: What's this for?

vloncar: It is the length of the function name. It is very useful when looking at reports. You can figure out exactly which layer you are looking at. Instead of `dense_U0` you would get full template info, like `dense_array_array_ap_fixed_16_7_5_3_0_8u_config6_U0`

thesps: Nice!

hls4ml/templates/vivado/nnet_utils/nnet_mult.h
```diff
@@ -0,0 +1,70 @@
...
+
+template<class data_T, class res_T, typename CONFIG_T>
+inline typename std::enable_if<(not std::is_same<data_T, ap_uint<1>>::value), res_T>::type
+cast(typename CONFIG_T::accum_t x){
+  return (res_T) x;
+}
+
+}
+
+#endif
```
thesps, line 70: Good call separating this into a new file.

vloncar: Gonna be even better once #231 turns these into classes.

thesps (approved):

thesps:
I'm happy to merge at this point. Passing tests (and other checks) show that "old" code base still works fine with `io_parallel`. We've seen a few issues with large CNN synthesis recently resolved by using this branch and switching to `io_stream`. Recent results, shown at FastML Workshop for example, show it behaves well. 

@nhanvtran & @jmduarte I requested reviews to have some extra oversight, since it's such a large PR.

nhanvtran:
adding @violatingcp here too.  

@thesps -- my recollection is that we would discuss it in the Dec 11 hls4ml meeting, but do you think we should merge now or wait?  we can discuss it either way at the meeting.

PR #223, opened 2020-10-02
ApplyAlpha layer inherits Trace setting from parent

thesps:
The QKeras optimization pass may insert an `ApplyAlpha` (`BatchNormalization`) layer for QKeras Dense or Conv layers with `alpha != 1`. Currently, these can only be `trace`d by adding the layer name to the config dictionary. Since the layer name is generated by the optimizer pass, it's a bit unwieldy for users (who may not anyway be aware that it's being added). 

This PR copies the `Trace` setting of the parent layer into the `ApplyAlpha` layer. 
I had to adjust some of the handling of the `Trace` settings to make it usable, specifically: rather than retrieving a layer's trace setting from the config, it's now retrieved from the layer attributes. Layer config entries were already made layer attributes, so this doesn't interfere with the existing functionality. 

I also added a 'convert_precision_string' method, continuing the push to remove precision strings everywhere, which is then used in the `profiling` to query the type width/integer bits.

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -38,6 +38,33 @@ class FixedPrecisionType(object):
         typestring = 'ap_{signed}fixed<{args}>'.format(signed='u' if not self.signed else '', args=args)
         return typestring
 
+def convert_precision_string(precision):
+    assert isinstance(precision, str), "This method converts precision strings to PrecisionTypes. You provided a {}".format(type(precision))
+    bits = re.search('.+<(.+?)>', precision).group(1).split(',')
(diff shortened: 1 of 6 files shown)


hls4ml/model/hls_layers.py
```diff
@@ -38,6 +38,33 @@ def __str__(self):
         typestring = 'ap_{signed}fixed<{args}>'.format(signed='u' if not self.signed else '', args=args)
         return typestring
 
+def convert_precision_string(precision):
```
thesps, line 41: So, @vloncar, where should we move this to?

PR #229, opened 2020-10-20
Handle no-bias fusing of Dense and BN

thesps:
@vloncar fixed a bug related to Keras layers with `use_bias=False` and the Dense-BatchNorm layer fusion. The 'bias' values from the BatchNorm end up in the Dense layer, but the Dense layer's bias precision is overruled since the Keras layer had `use_bias=False`. Wrong results follow...

Now, in that scenario, the Dense layer (which is now the fused Dense-BatchNorm) picks up the bias precision of the BatchNorm.


thaarres (approved):
Great, thanks for the fix!

PR #231, opened 2020-10-22
Enhancements for QKeras

thesps:
Not all QKeras quantizers are currently supported by hls4ml. Notably, `po2` weights, and `auto_po2` alpha. This PR will eventually add support for those quantizers. For the moment, I have overhauled the mechanism we were using to overload the `product` function for different specializations (binary/ternary, standard). If what I've done here looks okay, I intend to then add additional `product` function implementations. 

One significant change is the `convert_precision_string` method (and its use). I added this for PR #223 (which this would now supersede). I've now made that a method of Vivado Backend. Other Backends would need to implement the same (or we separate configuration from model further, as discussed). In order to simplify things for the `product_type` logic (which is now in the Python world rather than C++ templates), I added `precision = self.backend.convert_precision_string(precision)` as the final step of the `HLSConfig.get_precision` method. So now, no precision strings are carried by the `HLSModel`, only our `IntegerPrecisionType` and `FixedPrecisionType`.


thesps:
I also see the tests failing on Vivado 2017.4, but locally they do work on Vivado 2019.2. Is there a way to run only the 2018.2 branch in Jenkins just to check that?

vloncar:
Unfortunately 2018.2 is not installed in Jenkins anymore. I will update the whole setup next week to use 2020.1. The tests in #220 also fail in 2017.2 but not in 2018 and later versions.

thesps:
Could we also have 2019.2? I think ideally we should use at least the final release from the previous two years, and the latest version. So right now that would be 2018.3, 2019.2, 2020.1. (One fewer release each consecutive year also makes it easier 😁)

vloncar:
Sure, if there is enough space for multiple installations why not.

zhuangh:
Hi @thesps @vloncar really glad to see this PR :)

thesps:
This is ready for review now. I would say it's not the final development for QKeras, there are still some cases which don't work perfectly. But this PR does add support for `po2` quantizers using bitshift operations, which we didn't have before.

On the way to doing this I moved all internal types from using strings to using our `IntegerPrecisionType`, `FixedPrecisionType` etc. We would need to make sure this works nicely with #245 also.

thesps:
Even though previous tests passed, there was actually an issue with conversion of conv models. That should be fixed now. There is also an outstanding problem with merging the new product function handling with new `io_stream` specifically for `BatchNormalization` layer, but I don't think current tests probe the issue.

vloncar (approved):

vloncar:
Given that the final issue with this PR has been fixed and we have extensively tested this at CERN, I am merging it now.

PR #247, opened 2020-11-23
Add profiling for PyTorch models

AdrianAlan:
This PR extends profiling to support Pytorch `nn.Sequential` models. The two new methods mirror the behavior of the methods applied to Keras models.

Further collapsing  of three similar methods `weights_hlsmodel`, `weights_keras`, `weights_torch` (same for the `activations_<type>`) into one would be desirable.

### DIFF
--- a/hls4ml/model/profiling.py
+++ b/hls4ml/model/profiling.py
@@ -357,18 +357,16 @@ def numerical(model=None, hls_model=None, X=None, plot='boxplot'):
     plt.title("Distribution of (non-zero) weights")
     plt.tight_layout()
 
-    if X is not None and isinstance(model, keras.Model):
-        print("Profiling activations")
-        data = activations_keras(model, X, fmt='summary', plot=plot)
-        ap = plots[plot](data, fmt='summary') # activation plot
-        plt.title("Distribution of (non-zero) activations")
-        plt.tight_layout()
-    elif X is not None and isinstance(model, torch.nn.Sequential):
-        print("Profiling activations")
-        data = activations_torch(model, X, fmt='summary', plot=plot)
-        ap = plots[plot](data, fmt='summary')
-        plt.title("Distribution of (non-zero) activations")
-        plt.tight_layout()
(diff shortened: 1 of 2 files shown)


thesps:
Thanks a lot for this! I've tested it on both some `torch` and `tensorflow` models, and the functionality is all there.

Before merging this, though, we should take the opportunity to clean up the imports a bit. It wasn't great before, but now we have both `torch` and `tensorflow` options, we need to add some protections such that it should "just work" if the user has only one or other of those modules. We want those to be "soft dependencies". It would look something like this:
```
try:
    import tensorflow
    import qkeras
    __tf_profiling_enabled__ = True
except ImportError:
    __tf_profiling_enabled__ = False
# and the same for torch
```
And then, before running code like `isinstance(model, keras.Model)` and `isinstance(model, torch.nn.Sequential)`, we should check whether the appropriate imports were successful or not. It's similar to what we do for the [converters](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/converters/__init__.py#L24-L40).
@vloncar thinks we should handle it by separating the TF and torch specific portions into different modules, but I think it may be overkill. I'm not too bothered either way.

Do you think you can handle this? Otherwise (I think) I have a fairly clear idea of what sections needs to be 'protected'.

thesps:
The latest change works for me! I've tested it with environments either with/without TF and torch separately and it behaves nicely. I think a final nice thing would be to remove `tensorflow` from the extras require https://github.com/fastmachinelearning/hls4ml/blob/master/setup.py#L28

thesps (approved):

thesps:
As far as I'm concerned, this is ready. @vloncar the tests are failing, but is this because the test framework has been changed recently?

vloncar:
Tests fail ever since onnx released version 1.8.0. This happens even on master with old testing infrastructure. The new containers used in #249 don't have this issue, so I would merge that one first. This one passes when used on a new infrastructure.

Issue #255, opened 2020-12-05
Pytorch/Brevitas improvement for hls4ml


Duchstf:
Creating an issue here for discussion on Pytorch/Brevitas improvement. I'll create an initial PR based off Pedro's codes soon (probably next week). 

I think the main things I want to discuss here are:

1. What kinds of functionality/improvement we need to target in order to make Pytorch/Brevitas support on par with the current Keras/QKeras scheme. 

2. I think we should start gather more example models for Pytorch/Brevitas.

Duchstf:
Closing issue, further discussion should be moved to the PR.

PR #257, opened 2020-12-05
Pytorch Improvement

Duchstf:
As discussed in #255, this PR aims to improve Pytorch/Brevitas support for `hls4ml`. Many thanks to Pedro for initializing some works on this. 

The Pytorch converter is now modeled after the Keras converter (i.e function to process a layer is accessed using a decorator).

However, since Pytorch's default data format is `channel_first,` synthesizing Pytorch CNN models with streaming is not possible yet. Hopefully this will be addressed in a later PR.

Current status:

1. Conv1D, Conv2D layers.
2.  Batchnorm 1D & 2D layer.


Duchstf:
@vloncar I added a requirement to pass in input shape for Pytorch in the config file, and it seems like it's breaking the tests. Can you take a look at the error and let me know what happened (the error message is not very informative, but I guess that it means a hls project is not created model so testing failed)?

vloncar:
@Duchstf if pytorch is not installed, it fails. in `__init__.py` of converters, you import pytorch converters without checks, they in turn import torch and once that fails the whole `import hls4ml` fails.

Duchstf:
@vloncar Wait, but I have it [here](https://github.com/Duchstf/hls4ml/blob/bd652956122821ed0385abd9931a16f43f74de74/hls4ml/converters/__init__.py#L43-L47) ...

vloncar:
@Duchstf ...but not [here](https://github.com/Duchstf/hls4ml/blob/bd652956122821ed0385abd9931a16f43f74de74/hls4ml/converters/__init__.py#L8) ;-)

Duchstf:
Moving codes to #285.

PR #267, opened 2020-12-19
Added io_stream activation function implementations

veyron8800:
Hello,

This PR adds support for additional activation functions with the new io_stream option from PR #220 . These are activations that were previously supported with io_series and io_parallel. They are as follows:

1. TanH
2. Hard Sigmoid
3. Leaky ReLU
4. Thresholded ReLU
5. Softplus
6. Softsign
7. ELU
8. SeLU
9. PReLU

Their implementations are validated through a test script which can be found [here.](https://github.com/veyron8800/cnn_stream_dev_unit_tests)

Also of note, I switched around the result writing order in the C++ test bench to correct a bug. If data file paths are provided in the yaml configuration, then writing results to disk is done before printing to console. However, keeping data in the output stream is only done when printing to console. This caused no results to be printed at the checkpoint, because the stream was emptied out by writing to the log on that iteration. Reversing the order of these tasks allows data to be printed to console, while also being written to file. This also ensures that there is no leftover data in the stream that has not been read by the test bench.

Thanks


hls4ml/templates/vivado/nnet_utils/nnet_activation_stream.h
```diff
@@ -341,6 +341,309 @@ void softmax(hls::stream<data_T> &data, hls::stream<res_T> &res){
...
+        res.write(out_data);
+    }
+}
+
+
+// *************************************************
+//       ELU Activation
+// *************************************************
+template<class data_T, class res_T, typename CONFIG_T>
+void elu(hls::stream<data_T> &data, typename data_T::value_type alpha, hls::stream<res_T> &res) {
```
vloncar, line 543: It is possible to parse the ELU activation without `alpha` parameter. In that case it should default to `1.0`. We handle this [case](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/templates/vivado/nnet_utils/nnet_activation.h#L718) in the other implementation, can you add it here as well?

vloncar (changes requested):
Hi Andrew, thanks a lot for this and sorry for the delay. Can you add the missing function I mention below?

hls4ml/templates/vivado/nnet_utils/nnet_activation_stream.h
veyron8800: @vloncar  Nice catch, I totally missed that function definition while reading through nnet_activation.h. Just added it now with the latest commit.

veyron8800:
@vloncar  Also, no worries about the delay. Hope you enjoyed the holiday season!

vloncar (approved):

PR #268, opened 2020-12-23
Sphinx Documentation - Website Rebuild

ngpaladi:
This PR rebuilds the current website using Sphinx in preparation for implementing automatically generated API documentation. It includes a workflow to autobuild the site and push to the `gh-pages` branch.


.github/workflows/build-sphinx.yml
```diff
@@ -0,0 +1,33 @@
...
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+        git add .
+        git commit -m "Update Sphinx Documentation" -a || true
+    - name: Push Documentation Changes
+      uses: ad-m/github-push-action@master
+      with:
+        branch: gh-pages
+        directory: gh-pages
+        github_token: ${{ secrets.GITHUB_TOKEN }}
```
Duchstf, line 33: This token only needs access to write to the repo right? In that case can you change it to:

```
${{ secrets.PERSONAL_TOKEN }}
```

This is Javier's token to publish the website.

sphinx-docs/reference.rst
```diff
@@ -0,0 +1,53 @@
...
+* Javier Duarte [University of California San Diego]
+* Sergo Jindariani, Benjamin Kreis, Ryan Rivera, Nhan Tran [Fermilab]
+* Edward Kreinar [Hawkeye360]
+* Song Han, Philip Harris, Dylan Rankin [MIT]
+* Zhenbin Wu [University of Illinois at Chicago]
+* Mark Neubauer [University of Illinois Urbana-Champaign]
+* Shih-Chieh Hsu [University of Washington]
+* Giuseppe Di Guglielmo [Columbia University]
+* Duc Hoang [Rhodes College]
+
```
Duchstf, line 53: Should add yourself as one of the contributors.

Duchstf (commented):
Thank you so much for working on this, it looks very good!!! Here are a few small comments:

- In the demo website, the left selection menu for different submodules all have "package" at the end. I think you can remove it and just leave it with the names of the submodules (so `hls4ml.converters package` to `hls4ml.converters`). 
- Similar with the sub-submodules, I don't think they need "Module" at the end. 

If these changes are Sphinx's default and are hard to make then I guess we can just leave it like that 😄. 

Also, I think it's fine to just leave the links to the auto-generated doc there. Since it's "auto" anyways, we can just add the doctrings as we move along. This means we don't need everything to be completely standardized and clean before adding the links (this would take several months before something like that is established and people have time to go back and add docstrings).

.github/workflows/build-sphinx.yml
ngpaladi: Ok, will do.

sphinx-docs/reference.rst
ngpaladi: Ok, will do.

ngpaladi:
Thanks, I'll add the links to the autogenerated documentation back. With respect to the "package"/"module" in the names, I don't believe I can disable that and keep regenerating the documentation every time in case someone adds a new file. It seems to be built into `sphinx-apidoc` which generates the files, but I can look for a workaround.

ngpaladi:
This should resolve all issues with the exception of the naming. I think there exists a workaround but I'd say it's probably low priority.

Duchstf:
Yea it's completely fine!

Duchstf:
@jmduarte @vloncar @thesps @nhanvtran Not sure if people have more comments, I think the demo website looks good, I also reviewed this PR and I think it's good to be merged.

Duchstf (approved):

nhanvtran:
Thanks @ngpaladi ! This looks great! I have a few comments on the content but that's somewhat different than the port over to sphinx so let's handle that in a different PR and merge this one.  

we should wait to hear from @vloncar @thesps or @violatingcp too

vloncar:
@ngpaladi Looks great, thanks a lot!
Is there any hard requirement for this to be placed under `sphinx-docs`? The current contents of `docs` are useless, we should remove and replace them with these docs.

ngpaladi:
That absolutely can be done, I just wasn't sure if the current items in `docs/` were being used as a sort of "bundled" set of instructions.

vloncar (approved):

PR #272, opened 2021-01-11
fix n_in in onnx converter for Dense layers

jmitrevs:
Previously n_in was set to None for all dense layers except the first. This fixes it.


jmitrevs:
I think this is superseded by #285, so I will close this.

PR #277, opened 2021-01-21
Fix to profiling HLSModel types display

thesps:
After the previous update to profiling, which changed the axis labels `'0'` and `'1'` to the easier to understand `'w'` and `'b'`, the grey boxes showing HLSType are no longer added to the weights profiling plot. This PR fixes that.


thesps:
@GiuseppeDiGuglielmo this should fix the issue you saw.

vloncar (approved):

PR #246, opened 2020-11-20
Support for 1D dot product 

DelonShen:
I'm still not too familiar with everything in the code base but I think this should be alright for 1D dot products. Let me know what changes need to be made 

#243

PR #278, opened 2021-01-24
Make dot1d compatabile with nnet_mult.h

DelonShen:
nnet_mult.h broke the dot1d implementation. small change to fix.

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_merge.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_merge.h
@@ -20,8 +20,7 @@
 #ifndef NNET_MERGE_H_
 #define NNET_MERGE_H_
 
-#include "nnet_common.h"
-#include "nnet_dense.h"
+#include "nnet_mult.h"
 #include "hls_stream.h"
 #include <math.h>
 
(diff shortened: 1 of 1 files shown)


vloncar:
Thanks @DelonShen 

I tried this, but now it misses timing:
```
INFO: [BIND 205-100] After resource sharing, estimated clock period (5.0263ns) exceeds the target (target clock period: 5ns, clock uncertainty: 0.625ns, effective delay budget: 4.375ns).
INFO: [BIND 205-100] The critical path consists of the following:
	wire read on port 'data2_49_V_read' (firmware/nnet_utils/nnet_merge.h:124) (0 ns)
	'call' operation ('mult[49].V', firmware/nnet_utils/nnet_merge.h:136) to 'product' (5.03 ns)
```
I used the example from [here](https://github.com/fastmachinelearning/hls4ml/pull/246#issuecomment-733194895). I tried adding `product_type` to the config, in line with the `Dense` and `Conv` layers, but the issue persists. Reverting to plain multiplication with `*` makes the problem go away. @thesps any ideas?

vloncar:
This was resolved in #280

PR #280, opened 2021-01-31
Fixes for the use of resource strategy and new `product` functions

vloncar:
Two related fixes. The first one ensures that the proper multiplication function is called wrt the strategy specified in config. The second one adds new `product` functions to `SeparableConv` layers, which was missed in previous PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,7 @@ pipeline {
         dir(path: 'test') {
           sh '''#!/bin/bash --login
               conda activate hls4ml-py36
+              pip install git+git://github.com/google/qkeras.git@v0.8.0#egg=qkeras
               pip install -U ../ --user
               ./convert-keras-models.sh -x -f keras-models.txt
               pip uninstall hls4ml -y'''
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,7 @@ pipeline {
         dir(path: 'test') {
           sh '''#!/bin/bash --login
               conda activate hls4ml-py36
+              pip install tensorflow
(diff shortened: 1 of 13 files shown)


nhanvtran:
thanks @vloncar so this will make it possible to use "dense_large" in conv2d? (just double checking to make sure i'm interpreting the code right)

vloncar:
Resource strategy already works for new (`io_stream`) algorithm. This fixes the old one that only works with very small models.

thesps (approved):

vloncar:
This PR now includes a new fix for issues encountered in #278.

thesps:
Could you move the Pooling1D fix to a new PR?

vloncar:
Remind me, what is blocking this PR?

thesps:
Latency and II with `Resource: Strategy` are still not correct. e.g. for the `RF=112` test:
```
Synthesis report for qkeras_mnist_dense-xcvu9p_flgb2104_2_e-c5-io_parallel-rf112-ap_fixed_16_6_-Resource-2020.1
...
    |  Latency (cycles) |  Latency (absolute) |  Interval | Pipeline |
    |   min   |   max   |    min   |    max   | min | max |   Type   |
    +---------+---------+----------+----------+-----+-----+----------+
    |      281|      283| 1.405 us | 1.415 us |  117|  118| dataflow |
    +---------+---------+----------+----------+-----+-----+----------+
```
PR #284 goes part of the way, but I think we need to get to the bottom of it before merging, since it might be related to the new `product` functions and/or this PR, it's not clear.
I think we should keep this clean, so could you separate `2c541f1 - Fix 1D pooling` into a new PR, and pull #284 ?

PR #283, opened 2021-02-10
autopep8 + clang-format

jmduarte:
Ran autopep8 recursively on all python files with the `setup.cfg` in the repo:
```
autopep8 --in-place --aggressive --aggressive *.py -v
```

```
[pep8]
ignore = E121,E123,E126,E226,E24,E704,W503,W504
max-line-length = 199
```

We can combine this with a GitHub action to automatically check/correct PRs:
https://github.com/marketplace/actions/autopep8

Let me know what you think!

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+---
+# BasedOnStyle:  Google
+ColumnLimit:     120
+IndentWidth:     4
+BreakBeforeBraces: Stroustrup
+...
+
--- a/hls4ml/converters/__init__.py
+++ b/hls4ml/converters/__init__.py
@@ -40,6 +40,7 @@ try:
 except ImportError:
     __tensorflow_enabled__ = False
 
+
 def parse_yaml_config(config_file):
     """Parse conversion configuration from the provided YAML file.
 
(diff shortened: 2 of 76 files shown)


vloncar:
Can we postpone this until after we merge the Intel backend and flesh out a strtegy for integrating Phil's CNN implementation? It's just one command, we can run it at any point, so I would prefer to merge the big stuff before that.

jmduarte:
@vloncar oh yes definitely! 

I should have said this is just a placeholder / for discussion for now. I figured we'd rebase/merge it during a quiet time without other major open PRs.

jmduarte:
I tried to use `clang-format` to format the HLS code. Maybe you have strong opinions on the formatting to use. 

One issue is that pragmas are pushed all the way to the left by default, but apparently this will be fixed in upcoming release: 
https://reviews.llvm.org/D92753

For now, I'm using a hack I found here: 
https://stackoverflow.com/questions/31353748/how-could-i-indent-c-pragma-using-clang-format
but it's not always working (if there are multiple pragmas in a row).

jmitrevs:
This PR has not been very active recently, but I think we should make some decisions soon and incorporate something along this line.

jmduarte:
Closed in favor of #678

PR #284, opened 2021-02-15
Resolve II and Latency issues

thesps:
Reviewing #280 I saw that the latency and II of some of the test models is not as expected. 
e.g. the model [1] with RF=2 was synthesizing with II=19, and the model [2] with RF=112 synthesized with II= 117-min, 118-max. You can see this in [the reports](https://correlator2.fnal.gov:8080/blue/organizations/jenkins/hls4ml/detail/PR-280/4/pipeline) of #280. 
Basically, with ` Strategy: Resource`, some functions that should have been pipelined were not. These additional pragmas resolve that. Regarding the `inline` pragmas in the dense- and softmax-wrapper functions: those were automatically inlined in the `Strategy: Latency` tests, and only the wrapper function is inlined, not the actual dense or softmax implementation underneath.

I've marked the PR WIP since I still see a different min/max latency for the `Strategy: Resource` which we should also resolve. Although we could take these changes sooner if that one takes too long to find.

[1] KERAS_3layer-xcvu9p_flgb2104_2_e-c5-io_parallel-rf2-ap_fixed_16_6_-Resource-2020.1
[2] qkeras_mnist_dense-xcvu9p_flgb2104_2_e-c5-io_parallel-rf112-ap_fixed_16_6_-Resource-2020.1

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_activation.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_activation.h
@@ -221,6 +221,7 @@ void init_invert_table(typename CONFIG_T::inv_table_t table_out[CONFIG_T::table_
 
 template <class data_T, class res_T, typename CONFIG_T>
 void softmax_latency(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in]){
+    #pragma HLS pipeline
     // Initialize the lookup tables
 #ifdef __HLS_SYN__
     bool initialized = false;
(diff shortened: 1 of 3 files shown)


thesps:
To clarify - I noticed the issue while reviewing PR280, but it seems it has been around for much longer.

thesps:
The different min & max latency for `Strategy: Resource` comes from the top level `dataflow` pragma, seemingly due to a check that is inserted for each function in the dataflow region. The addition of [`stable` pragma](https://www.xilinx.com/html_docs/xilinx2019_1/sdaccel_doc/hls-pragmas-okr1504034364623.html#unique_735555730) for each variable used for layer input/output seems to resolve the issue. Before removing the WIP label and looking to merge, I plan to test on hardware/cosim to make sure this is safe, ideally including the combination `io_type: io_stream` and `Strategy: Resource` which makes more use of `dataflow`.

vloncar:
Fantastic! Thanks a lot Sioni

thesps:
I tested and didn't see any discrepancies in cosim for `Strategy: Resource` with both `io_parallel` and `io_stream`. Ready for review/merge.

vloncar (approved):

PR #285, opened 2021-02-17
Pytorch & ONNX Clean up & Improvement

Duchstf:
In this PR:

* Cleaned up and debug ONNX & Pytorch codes.
* Organize ONNX & Pytorch according to Keras scheme.
* Initial support for ONNX & Pytorch from Python API. 

The codes were tested and it worked with some of my test models (MLP 2 layers/3layers). Due to problems with CNN backend (no support for channel_first), I didn't test it on any CNN models. 

@thesps @vloncar Let me know how I should proceed and if I need to add more tests. 

Thanks!


hls4ml/converters/onnx/pooling.py
```diff
@@ -0,0 +1,58 @@
...
+        layer['stride'] = strides[0]
+        layer['pool_size'] = layer['y_filt'] = kernel_shape[0]
+        
+        #Padding
+        pads = compute_pads_1d(node, layer)
+        layer['pad_left'] = pads[0]
+        layer['pad_right'] = pads[1]
+
+        if all(x == 0 for x in pads): # No padding, i.e., 'VALID' padding
+            layer['n_out'] = int(math.ceil(float(layer['y_in'] - layer['y_filt'] + 1) / float(layer['stride'])))
```
vloncar, line 23: Same as with convolution, computing output shape should be shared by all parsers

hls4ml/converters/onnx_to_hls.py
```diff
@@ -8,6 +8,7 @@
 
 MAXMULT = 4096
 
+####----------------------Data reader---------------------######
```
vloncar, line 11: Docstrings instead of ASCII art comments

hls4ml/converters/onnx_to_hls.py
```diff
@@ -115,7 +116,64 @@ def compute_pads_2d(operation, layer):
     
     return pads
 
-def onnx_to_hls(yamlConfig):
+####----------------------Optimizer---------------------######
+def _hls4ml_onnx_optimizer(graph):
```
vloncar, line 120: Why is this not implemented as a regular hls4ml optimizer? `Transpose` is a valid layer and should appear in the internal model, then if it is placed before and after a real layer (which only happens when using one very poorly designed ONNX conversion tool) we can optimize it away.

hls4ml/converters/onnx_to_hls.py
```diff
@@ -125,252 +183,95 @@ def onnx_to_hls(yamlConfig):
     layer_list = []
 
     #Extract model architecture
-    model = ModelProto()
-    with open(yamlConfig['OnnxModel'], 'rb') as fid:
-        model.ParseFromString(fid.read())
+    print('Interpreting Model ...')
+    if 'OnnxAPIModel' in config:
```
vloncar, line 187: I would prefer to just have `OnnxModel` and then figure out how to proceed based on the object that is in the config (a real model or a string that is its path on the filesystem).

hls4ml/converters/onnx_to_hls.py
```diff
@@ -125,252 +183,95 @@ def onnx_to_hls(yamlConfig):
...
     inputs_map = {}
-
-    passes = ['fuse_transpose_into_gemm', 'fuse_matmul_add_bias_into_gemm', 'eliminate_nop_transpose', 'fuse_consecutive_transposes']
-    model = shape_inference.infer_shapes(model) # have to infer shapes before optimizing the model
-    model = optimizer.optimize(model, passes)
-    model = shape_inference.infer_shapes(model) # have to infer shapes before optimizing the model
-    
     reader = ONNXDataReader(model)
+    
+    #Input layer info (assuming that there is only one input layer)
```
vloncar, line 202: Why assuming just one? You can get the list of input layers from the onnx graph and create one input layer for each as you parse.

hls4ml/converters/onnx_to_hls.py
```diff
@@ -125,252 +183,95 @@ def onnx_to_hls(yamlConfig):
...
-        layer_list.append(input_layer)
+    sanitize_layer_name(input_layer)
+    layer_list.append(input_layer)
 
-    # Check for unsupported layer type
-    for operation in model.graph.node:
-        if operation.op_type not in supported_operations:
-            raise Exception('ERROR: Unsupported operation type: {}'.format(operation.op_type))
+    # Defined supported layers and check for unsupported layer type
+    skip_layers = ['Squeeze', 'Unsqueeze', 'Dropout', 'Identity', 'Flatten', 'Reshape']
```
vloncar, line 216: Reshape should be handled, we have models that need to handle it, more will come in the future. Squeeze/Unsqueeze can probably be folded into Reshape as well.

hls4ml/converters/onnx_to_hls.py
```diff
@@ -125,252 +183,95 @@ def onnx_to_hls(yamlConfig):
...
+        layer_list.append(layer)
 
 
     #################
     ## Generate HLS
     #################
 
     print('Creating HLS model')
-    hls_model = HLSModel(yamlConfig, reader, layer_list, input_layers, output_layers)
+    hls_model = HLSModel(config, reader, layer_list)
```
vloncar, line 276: If you don't pass input/output layers only sequential models are supported. This should be solved now or documented as a limitation in the docs for future improvement

hls4ml/converters/pytorch/convolution.py
```diff
@@ -0,0 +1,80 @@
...
+    layer['filt_width'] = pytorch_layer.kernel_size[0] 
+    layer['stride_width'] = pytorch_layer.stride[0]
+    layer['pad_left'] = layer['pad_right'] = pytorch_layer.padding[0]
+    layer['dilation'] = pytorch_layer.dilation[0]
+    
+    #Ouput info
+    layer['out_width'] = int(
+        (layer['in_width'] + 2*layer['pad_left'] - layer['dilation']*(layer['filt_width']-1)- 1)/layer['stride_width'] + 1)
+    
+    output_shape=[input_shapes[0][0], layer['n_filt'], layer['out_width']] #Channel first as default
```
vloncar, line 32: Same as mentioned in the ONNX converter, this can be made into an utility function.

hls4ml/converters/pytorch/convolution.py
```diff
@@ -0,0 +1,80 @@
...
+    layer['filt_height'] = pytorch_layer.kernel_size[0]
+    layer['filt_width'] = pytorch_layer.kernel_size[1]
+    layer['stride_height'] = pytorch_layer.stride[0]
+    layer['stride_width'] = pytorch_layer.stride[1]
+    layer['dilation'] = pytorch_layer.dilation[0]
+    layer['pad_top'] = layer['pad_bottom'] = pytorch_layer.padding[0]
+    layer['pad_left'] = layer['pad_right'] = pytorch_layer.padding[1]
+    
+    #Ouput info
+    layer['out_height']  = int((layer['in_height'] + 2*layer['pad_top'] - layer['dilation']*(layer['filt_height']-1)- 1)/layer['stride_height'] + 1)
```
vloncar, line 71: Ditto

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -9,98 +9,179 @@
 
 from hls4ml.model import HLSModel
 
-class PyTorchDataReader:
+####---------------Initial Data Reader------------------######
```
vloncar, line 12: Docstrings instead of comments

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -9,98 +9,179 @@
...
 
-    #Loop through layers
-    print('Topology:')
-    modelstr = repr(reader.torch_model).split('\n')
-    for pytorch_layer in modelstr:
-        layer_match = re.match(r'\((\d)\): (\w+)\((.*)\)', pytorch_layer.strip())
-        if layer_match is None:
-            continue
+    print('Interpreting Model ...')
+    if 'PytorchAPIModel' in config:
```
vloncar, line 100: Use just `PyTorchModel`

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -9,98 +9,179 @@
...
 
     #################
     ## Generate HLS
     #################
-
-    reader = PyTorchDataReader(yamlConfig)
+    
     print('Creating HLS model')
-    hls_model = HLSModel(yamlConfig, reader, layer_list)
+    hls_model = HLSModel(config, reader, layer_list)
```
vloncar, line 186: Same as with ONNX, this should have input and output layers

hls4ml/utils/config.py
```diff
@@ -216,3 +216,35 @@ def make_layer_config(layer):
         config['LayerName'] = name_config
 
     return config
+
+######---------------Pytorch config generation--------------######
```
vloncar, line 220: Docstring here

setup.py
```diff
@@ -35,8 +35,7 @@ def get_version(rel_path):
         'profiling': [
             'pandas',
             'seaborn',
-            'matplotlib'
-        ]
+            'matplotlib']
```
vloncar, line 38: How is this cleaner code? This is the last nitpick :-)

vloncar (changes requested):
Thanks Duc, awesome work. Minor stuff only. There aren't many things to change, the comments are duplicated between PyTorch and ONNX converters

setup.py
Duchstf: Sorry I added a dependency & then remove it. Didn't realize it changed the code.

hls4ml/converters/onnx_to_hls.py
Duchstf: Sorry, I'm not sure what you mean by this? so do suggest that I implement it [here](https://github.com/fastmachinelearning/hls4ml/tree/master/hls4ml/model/optimizer) instead of an onnx optimizer?

vloncar: What ONNX models have this issue? PyTorch or Keras? If the issue is with PyTorch, does [this](https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html) help?

Performing graph transformations in the converter will make it harder to maintain it in the future. We've been there before. For example, your current transformation is dangerous, you need to make extra checks to ensure that the axes of transpose match (meaning you are transposing back into original format). Also how would you add future optimizations on top of this? You're modifying the graph as you iterate it, that's asking for trouble.

hls4ml/utils/config.py
```diff
@@ -216,3 +216,35 @@ def make_layer_config(layer):
...
+def config_from_onnx_model(model, granularity='model', default_precision='ap_fixed<16,6>', default_reuse_factor=1):
+    
+    config = {}
+
+    model_config = {}
+    model_config['Precision'] = default_precision
+    model_config['ReuseFactor'] = default_reuse_factor
+    model_config['Strategy'] = 'Latency'
+    #model_config['Compression'] = False
+    #model_config['Trace'] = False
```
jmduarte, line 246: Remove commented code

hls4ml/utils/config.py
```diff
@@ -216,3 +216,35 @@ def make_layer_config(layer):
...
+def config_from_pytorch_model(model, granularity='model', default_precision='ap_fixed<16,6>', default_reuse_factor=1):
+    
+    config = {}
+
+    model_config = {}
+    model_config['Precision'] = default_precision
+    model_config['ReuseFactor'] = default_reuse_factor
+    model_config['Strategy'] = 'Latency'
+    #model_config['Compression'] = False
+    #model_config['Trace'] = False
```
jmduarte, line 230: Remove commented code

hls4ml/converters/keras/pooling.py
```diff
@@ -1,10 +1,7 @@
 import math
 from hls4ml.converters.keras_to_hls import parse_default_keras_layer
 from hls4ml.converters.keras_to_hls import keras_handler
-from hls4ml.converters.keras_to_hls import parse_data_format
-from hls4ml.converters.keras_to_hls import compute_padding_1d
-from hls4ml.converters.keras_to_hls import compute_padding_2d
-
+from hls4ml.converters.utils import *
```
jmduarte, line 4: No star import

jmduarte (changes requested):

hls4ml/utils/config.py
```diff
@@ -216,3 +216,35 @@ def make_layer_config(layer):
         config['LayerName'] = name_config
 
     return config
+
+######---------------Pytorch config generation--------------######
+def config_from_pytorch_model(model, granularity='model', default_precision='ap_fixed<16,6>', default_reuse_factor=1):
```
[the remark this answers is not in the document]
Duchstf: This makes perfect sense! However I would plan to include this with the quantization PR (which will require us to generate the config with higher granularity).

The purpose of this PR is mainly to set up a basic software framework that we can develop on later.

vloncar:
@Duchstf Is this ready for a new review? Can you resolve the conflicts?

Duchstf:
@vloncar Thanks for checking in! I was actually about to message you! 

I addressed most of the comments, but I'm still kind of stuck at squeeze/unsqeeze and transpose layer. Is there any implementation of these layers in the Keras converter?

After this I think I'll need to move the "double transpose" optimization to hls model graph, and then add in input/output layers for onnx.

vloncar:
Hi Duc,

There is [`Transpose`](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/model/hls_layers.py#L1486) layer in hls4ml, your converters should use that, and then you should write an optimizer to get rid of the double transpose. Squeeze/Unsqueeze can be mapped to `Reshape`. I don't understand the input/output layers for ONNX, can you elaborate?

Duchstf:
@vloncar Thank you! I'll try to work on this.

Regarding the input/outputs, there is some inconsistency in ONNX that I need to resolve. For example, this is how the output layer is obtained in onnx in our converter:

```
output_layers = [x.name for x in model.graph.output]
```

If I have this model for example:

Screen Shot 2021-03-30 at 1 41 11 PM

Then `output_layers` would be `["36"]`, however, the correct output layer is named `"Transpose_2"` in this case. So if you try to access this layer in the hls model graph it would not be available. 

Another problem in this case is that the transpose layer is optimized away, so I need to remap the output layer to the Softmax layer.

Duchstf:
@jmduarte @vloncar Hello, sorry to be late on this!! 

I addressed all the major comments now, and I think the current code base provide a nice foundation to be extended in the future. The code has been tested through a [rigorous test bench that Javier wrote](https://github.com/jmduarte/pytorch_dev_hls4ml/blob/main/MLP_performance.py) for the 3-layer MLP jet-tagging model. 

With the current codes ONNX can also support more complicated non-sequential architecture, whereas Pytorch can only process linear models for now (noted in the doc). More plans on other additions to come. 

I'm requesting that this PR is merged now to prepare for an up-coming GNN PR of Javier's student. I will also work on later extensions after I can create more example ONNX models through the QONNX frontend.

Thanks!

jmduarte (approved):

jmduarte:
The changes look good to me.

I've tested this branch with the 3-layer jet classification model and find perfect agreement between the performance for the ONNX-converted and PyTorch-converted hls4ml models.

From here: https://github.com/jmduarte/pytorch_dev_hls4ml/blob/main/MLP_performance.py

```
PyTorch Accuracy: 0.7705041666666667
PyTorch j_g eff_bkg @ eff_sig=0.5: 0.01709535431921063
PyTorch j_g auc:                   0.9428119136939714
PyTorch j_q eff_bkg @ eff_sig=0.5: 0.04826059063347199
PyTorch j_q auc:                   0.9077707675852923
PyTorch j_w eff_bkg @ eff_sig=0.5: 0.0043961201501877345
PyTorch j_w auc:                   0.955058513686781
PyTorch j_z eff_bkg @ eff_sig=0.5: 0.0013408182643446685
PyTorch j_z auc:                   0.9506058228986847
PyTorch j_t eff_bkg @ eff_sig=0.5: 0.0036755876241293976
PyTorch j_t auc:                   0.9636456150795238
PyTorch average eff_bkg @ eff_sig=0.5: 0.004639871231215533
PyTorch average auc                  : 0.9439785265888506
PyTorch->hls4ml ap_fixed<16,6> Accuracy: 0.7500875
PyTorch->hls4ml j_g eff_bkg @ eff_sig=0.5: 0.01775768195756931
PyTorch->hls4ml j_g auc:                   0.9406825048141373
PyTorch->hls4ml j_q eff_bkg @ eff_sig=0.5: 0.048664141884480865
PyTorch->hls4ml j_q auc:                   0.9064459952657805
PyTorch->hls4ml j_w eff_bkg @ eff_sig=0.5: 0.004151022110972048
PyTorch->hls4ml j_w auc:                   0.9530614087511615
PyTorch->hls4ml j_z eff_bkg @ eff_sig=0.5: 0.0021025282510930016
PyTorch->hls4ml j_z auc:                   0.9489382343086378
PyTorch->hls4ml j_t eff_bkg @ eff_sig=0.5: 0.0031534871093382897
PyTorch->hls4ml j_t auc:                   0.9621728642823041
PyTorch->hls4ml average eff_bkg @ eff_sig=0.5: 0.004554429724725144
PyTorch->hls4ml average auc                  : 0.9422602014844041
```
```
ONNX Accuracy: 0.7705916666666667
ONNX j_g eff_bkg @ eff_sig=0.5: 0.01714750610175856
ONNX j_g auc:                   0.9428539660674484
ONNX j_q eff_bkg @ eff_sig=0.5: 0.048286459303408455
ONNX j_q auc:                   0.907730672199003
ONNX j_w eff_bkg @ eff_sig=0.5: 0.0043961201501877345
ONNX j_w auc:                   0.9550627050807532
ONNX j_z eff_bkg @ eff_sig=0.5: 0.0013095151142043261
ONNX j_z auc:                   0.9506387713869444
ONNX j_t eff_bkg @ eff_sig=0.5: 0.003790449737383441
ONNX j_t auc:                   0.963492754765287
ONNX average eff_bkg @ eff_sig=0.5: 0.004619214586735845
ONNX average auc                  : 0.9439557738998872
ONNX->hls4ml Accuracy: 0.7500875
ONNX->hls4ml j_g eff_bkg @ eff_sig=0.5: 0.01775768195756931
ONNX->hls4ml j_g auc:                   0.9406825048141373
ONNX->hls4ml j_q eff_bkg @ eff_sig=0.5: 0.048664141884480865
ONNX->hls4ml j_q auc:                   0.9064459952657805
ONNX->hls4ml j_w eff_bkg @ eff_sig=0.5: 0.004151022110972048
ONNX->hls4ml j_w auc:                   0.9530614087511615
ONNX->hls4ml j_z eff_bkg @ eff_sig=0.5: 0.0021025282510930016
ONNX->hls4ml j_z auc:                   0.9489382343086378
ONNX->hls4ml j_t eff_bkg @ eff_sig=0.5: 0.0031534871093382897
ONNX->hls4ml j_t auc:                   0.9621728642823041
ONNX->hls4ml average eff_bkg @ eff_sig=0.5: 0.004554429724725144
ONNX->hls4ml average auc                  : 0.9422602014844041
```

jmduarte:
@Duchstf can you fix the minor conflict we introduced?

@vloncar can we merge this after that's fixed? a follow-up PR should be to add testing to the test suite.

Duchstf:
@jmduarte there are some conflicts from rebasing so I just resolve the easier conflict from merging ... hopefully this doesn't introduce much pain for Abdul ...

vloncar:
I tried dogfooding the onnx converter with a model produced from qonnx project (but without the `Quant` nodes)
[test_285.zip](https://github.com/fastmachinelearning/hls4ml/files/6902912/test_285.zip)
The model is very simple (try it out in [Netron](https://netron.app/)), it should be easily parsable.

Duchstf:
@vloncar Thanks for the model, yes it shouldn't be too hard to fix!

Duchstf:
@vloncar I made some changes and it was able to parse all the layers, this is the current output:

```
Interpreting Model ...
Output layers:  ['Softmax']
Input shape: [None, 28, 28, 1]
Topology:
Layer name: Transpose4, layer type: Transpose, current shape: [[None, 28, 28, 1]]
Layer name: conv2d_0_m, layer type: Conv2D, current shape: [[None, 1, 28, 28]]
Layer name: Relu2, layer type: Activation, current shape: [[None, 32, 14, 14]]
Layer name: conv2d_1_m, layer type: Conv2D, current shape: [[None, 32, 14, 14]]
Layer name: Relu1, layer type: Activation, current shape: [[None, 64, 6, 6]]
Layer name: conv2d_2_m, layer type: Conv2D, current shape: [[None, 64, 6, 6]]
Layer name: Relu, layer type: Activation, current shape: [[None, 64, 3, 3]]
Layer name: PushTranspose_0, layer type: Transpose, current shape: [[None, 64, 3, 3]]
Layer name: dense, layer type: Dense, current shape: [[None, 576]]
Layer name: Add, layer type: Merge, current shape: [[None, 10]]
Layer name: Softmax, layer type: Softmax, current shape: [[None, 10]]
Creating HLS model
[0, 3, 1, 2]
  File "qonnx_test.py", line 5, in <module>
    hls_model = hls4ml.converters.convert_from_onnx_model(onnx_model, output_dir='output_hls')
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/converters/__init__.py", line 325, in convert_from_onnx_model
    return onnx_to_hls(config)
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/converters/onnx_to_hls.py", line 305, in onnx_to_hls
    hls_model = HLSModel(config, reader, layer_list, input_layers, output_layers)
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/model/hls_model.py", line 317, in __init__
    self._make_graph(layer_list)
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/model/hls_model.py", line 332, in _make_graph
    self.graph[name] = self.make_node(kind, name, layer, inputs, outputs)
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/model/hls_model.py", line 363, in make_node
    node = layer_map[kind](self, name, attributes, inputs, outputs)
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/model/hls_layers.py", line 375, in __init__
    self.initialize()
  File "/data/dhoang/onnx_merge/hls4ml/hls4ml/model/hls_layers.py", line 1519, in initialize
    raise Exception('ERROR: Transpose of tensors with rank > 3 is not yet supported.')
Exception: ERROR: Transpose of tensors with rank > 3 is not yet supported.
```

`[0, 3, 1, 2]` here is the print out of the transpose layer's `perm`, I know that the transpose layers only operate on the conv's input/output, so I'm not sure whether to fix this or throwing an error in this case.

jmduarte:
I think the problem is that the current HLS Transpose layer is written with the `Keras` `perm` option in mind: 

https://www.tensorflow.org/api_docs/python/tf/keras/layers/Permute#args

which for the same use-case would be written like this: `[3, 1, 2]`, i.e. they exclude the batch/samples dimension (that is not ever transposed with anything).

In ONNX, they explicitly include the batch/samples dimension, so the `perm` array is written like this `[0, 3, 1, 2]`.

So maybe we should just remove the first entry of the `perm` array in the ONNX converter before passing it to the `Transpose` HLS layer?

Duchstf:
@vloncar @jmduarte I pushed some small fixes to support parsing this model, everything parses fine now and the results of the 3-layer test was not affected!

Here are some detailed changes:

- Remove some bugs in Conv layer parsing.
- Extend the merge layer parsing to recognize when an Add layer is `AddBias` and use the current infrastructure in tf to merge it with `MatMul` ( which was parsed to `Dense`).
- Extend the reader to find the correct bias weights in `AddBias.`
- Ignore the batch dimension in `Transpose`.
- Some extensions to make the codes more efficient.

vloncar:
This needs a far more thorough testing to be considered ready. For example, just adding a simple `MaxPool` layer to the model above again breaks the conversion.

I've created [test case](https://gist.github.com/vloncar/dd89536ed36cf3b55eea93506d20be55) with some nonsense models created with Keras. I think this should (reasonably) pass before we merge this PR. The similar thing should be done for ONNX models exported from PyTorch.

Duchstf:
@vloncar @jmduarte So here is the summary of the testing that I did:

1. For the keras2onnx test, the converter can parse all the models (need to remove the Clip in the 2d layer but I think that's a keras2onnx bug), I also tried to vary different architecture and they were parsed just fine. For compilation, the branch model compiled, but not the 1d/2d model. I think this is mainly a backend problem (channel first operations not supported) and not the converter. Additionally, I also manually checked all the output shapes to make sure they are consistent. 

2. I made a similar testing script but converting from pytorch models, in my script, I can compile my test2d and branch model, but compilation for 1d model also failed because of a similar backend problem.

Overall, I think the converter is in good shape and is ready to be merged now, I can take these testing scripts, refine them and do a follow-up PR on testing (maybe after the Keras testing infrastructure is integrated). 

For reproducibility, I put the testing scripts and well as the conda environment setup[ in this repo](https://github.com/Duchstf/onnx_test).

vloncar:
`Clip` is not a bug. ReLU is really a `Clip(min=0, max=None)` and you can test for this condition and emit an `Activation` node in hls4ml (and bail otherwise until we support Clip in HLS).

The way converter handles `Gemm` and inserts a `Conv1D` if applied to 3D tensor is a wrong approach. What would happen if the test script did `Dense` after `Conv2D`? Your converter would fail. You should ensure that `Gemm` handler doesn't blindly assume that input is 2D and compute the correct output shape regardless of the input dimensionality. hls4ml now contains converters that will do the change from `Dense`->`Conv1D/2D` via an optimizer.

Those are my last two nitpicks :smile:

Duchstf:
@vloncar fixed 😉

vloncar (approved):
I'll merge this now, but it needs to be followed up by cleanups

PR #288, opened 2021-03-01
Command line tool and API consistency

vloncar:
Currently, when using API to write projects the projects don't keep the yml config in the output directory, so they can't be used by `hls4ml build` command. The changes ensure that we always write the yml in the output dir. I also made the minor change to get rid of the warning newer Python produces about comparing to a literal value.


thesps (approved):
Warning fixed 👍
And the config writer is a nice touch 👍

PR #290, opened 2021-03-02
Further cmdline and API changes

vloncar:
This PR handles the case when the API is used to convert the Keras model with `convert_from_keras_model`. The configuration dict will have a python object that is not serializable to YAML, so it cannot be used by `hls4ml convert` and `hls4ml build`, so we add a hook to PyYAML to serialize and deserialize such objects. I moved the `parse_config` from the `hls4ml` command to the converter, since the function is usable for API as well.

### DIFF
--- a/hls4ml/converters/__init__.py
+++ b/hls4ml/converters/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import os
+import yaml
 import importlib
 
 from hls4ml.utils.config import create_vivado_config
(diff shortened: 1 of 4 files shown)


jmduarte (approved):
Tested it out on the Tiny ResNet model and it works. Changes look good to me.

I guess this is more relevant for #285, but should we have a similar mechanism for PyTorch models?

thesps:
I also tested, also making sure it doesn't break command line only flow without TF in the environment. Nice 👌

PR #294, opened 2021-03-03
Docstrings for most common functions

vloncar:
In addition to the other documentation changes, here are some docstrings for most commonly used API functions.


thesps (approved):
Nice, all these docstrings are good descriptions, much appreciated.

You sneaked in a non-docstring-related change, but I'll allow it.

PR #299, opened 2021-03-08
Fix 'resource' strategy for CNN layers

vloncar:
For `resource` strategy in `Conv1D/2D` we need to transpose the weights. The previous transpose operations were not correct. This will require a new release, 0.5.1.

### DIFF
--- a/hls4ml/__init__.py
+++ b/hls4ml/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = '0.5.0'
+__version__ = '0.5.1'
 
 from hls4ml import converters
 from hls4ml import report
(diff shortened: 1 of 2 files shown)


thesps:
Could you provide a code snippet and/or plot showing the comparison of layer output from Keras (or other) and hls4ml before & after this change?

vloncar:
[Here](https://gist.github.com/vloncar/abbe2a789e9f71f263e493c804269f64). You'll have to do the plotting yourself I'm afraid.

vloncar:
Yeah, sorry, I posted the wrong one. I'll update it now.

vloncar:
I updated the gist. BTW, you don't need `io_stream` to trigger the bug.

thesps (approved):
Thanks!

Issue #300, opened 2021-03-09
Zero size array is nnet_common.h


KOVI89alipes:
In C++ it is illegal to declare an array of zero length. 
https://stackoverflow.com/a/6180200/14582299

But it's allowed in the GNU C extension
https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

Zero size array appears in the `reduce` template function

https://github.com/fastmachinelearning/hls4ml/blob/c3509cfeb5f80054e01dead612392a9411ebfbe2/hls4ml/templates/vivado/nnet_utils/nnet_common.h#L47
```c++
template<class T, int N, class Op>
 T reduce(T* x, Op op){
	static constexpr int leftN = pow2(floorlog2(N - 1)) > 0 ? pow2(floorlog2(N - 1)) : 0;
	static constexpr int rightN = N - leftN > 0 ? N - leftN : 0;
	if(N == 1){
		return x[0];
	}else if(N == 2){
		return op(x[0],x[1]);
	}else{
		T left[leftN];
		T right[rightN]; <<<<<<<<<<<<
		#pragma HLS array_partition variable=left complete
		#pragma HLS array_partition variable=right complete
		ReduceLeft: for(int i = 0; i < leftN; i++){
			left[i] = x[i];
		}
		ReduceRight: for(int i = 0; i < rightN; i++){
			right[i] = x[i+leftN];
		}
		return op(reduce<T,leftN,Op>(left, op), reduce<T,rightN,Op>(right, op));
	}
 }
```

And it gives an error in MSVC compilation

Possible fix - replace `if` with `if constexpr`

```c++
 template<class T, int N, class Op>
 T reduce(T* x, Op op){
	static constexpr int leftN = pow2(floorlog2(N - 1)) > 0 ? pow2(floorlog2(N - 1)) : 0;
	static constexpr int rightN = N - leftN > 0 ? N - leftN : 0;
	if constexpr(N == 1){
		return x[0];
	}else if constexpr (N == 2){
		return op(x[0],x[1]);
	}else{
		T left[leftN];
        T right[rightN];
	
    	#pragma HLS array_partition variable=left complete
		#pragma HLS array_partition variable=right complete
		ReduceLeft: for(int i = 0; i < leftN; i++){
			left[i] = x[i];
		}
		ReduceRight: for(int i = 0; i < rightN; i++){
			right[i] = x[i+leftN];
		}
[...]

PR #302, opened 2021-03-10
Updated the reduce function to no longer making unnecessary copies - …

anders-wind:
…saving a memory usage of NlogN. Also making it C++ standard compliant by removing a zero-sized array

This solves: #300

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_common.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_common.h
@@ -41,27 +41,18 @@ enum strategy { latency, resource };
   * before applying and accumulate the result over the rolled dimension.
   * --- */
  template<class T, int N, class Op>
- T reduce(T* x, Op op){
-	static constexpr int leftN = pow2(floorlog2(N - 1)) > 0 ? pow2(floorlog2(N - 1)) : 0;
-	static constexpr int rightN = N - leftN > 0 ? N - leftN : 0;
-	if(N == 1){
-		return x[0];
-	}else if(N == 2){
-		return op(x[0],x[1]);
-	}else{
-		T left[leftN];
-		T right[rightN];
-		#pragma HLS array_partition variable=left complete
-		#pragma HLS array_partition variable=right complete
-		ReduceLeft: for(int i = 0; i < leftN; i++){
-			left[i] = x[i];
-		}
-		ReduceRight: for(int i = 0; i < rightN; i++){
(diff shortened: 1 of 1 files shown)


Awia00:
I dont have access to a linux machine currently, but I will try to get @KOVI89alipes to run the code.

vloncar:
I am running some CI tests on this (the server was down before). We'll see if anything regressed.

anders-wind:
Looks like CI checks out :) Is there anything you would like us to try manually?

thesps:
Indeed, the KERAS_3layer performance is unchanged. 

I made [this standalone test](https://gist.github.com/thesps/d87efc2558235e6b0244ee9b0cd41e06) as well, and everything looks good: identical resources, latency, II, and numerical performance with the old and new version.

> Is there anything you would like us to try manually?

Given this started as a solution to #300, we should make sure this actually works on Windows. It would be good if you or @KOVI89alipes can check that. 

And as @vloncar says we need to check that a model using Pooling and io_stream is unaffected. The [dummy model](https://gist.github.com/vloncar/abbe2a789e9f71f263e493c804269f64) in #299 should do the trick. I'll use it to test your PR as well as that one.

> saving a memory usage of NlogN.

FYI this isn't really a factor in evaluating this: the axes of performance are FPGA resource consumption, latency, throughput (II) and numerical accuracy. On those counts the new version is identical to the old - which is good, and expected.

thesps:
> And as @vloncar says we need to check that a model using Pooling and io_stream is unaffected. The dummy model in #299 should do the trick. I'll use it to test your PR as well as that one.

I've done this. Again, looks good: no change to any of the metrics with the new vs old `reduce` function.

So now I think we just need the confirmation that it does compile on Windows.

anders-wind:
> FYI this isn't really a factor in evaluating this: the axes of performance are FPGA resource consumption, latency, throughput (II) and numerical accuracy. On those counts the new version is identical to the old - which is good, and expected.

Okay - I have very limited knowledge about FPGA design and code generation. But good to hear that the generated code is atleast identical.

KOVI89alipes:
I checked this fix, it compiles both in GCC10.1 and MSVC(VS2017) with no extra warning

thesps:
And how about in Vivado HLS on Windows? It needs to work on whichever compilers they supply

KOVI89alipes:
Vivado 2020.1
Windows 10

Both CSim and HLS synthesis passed

thesps (approved):

PR #310, opened 2021-03-30
Fix profiling for models with layers with multiple inputs

jmduarte:
This PR attempts to fix profiling for models with layers with multiple inputs. This comes up with ResNet-type models that have `Add()` layers for example.

This is part of a bigger update for TinyML, which lives here: https://github.com/jmduarte/hls4ml/tree/pynq but I cherry-picked this commit.

### DIFF
--- a/hls4ml/model/profiling.py
+++ b/hls4ml/model/profiling.py
@@ -248,13 +248,10 @@ def activations_keras(model, X, fmt='longform', plot='boxplot'):
         # or histogram bin edges and heights
         data = []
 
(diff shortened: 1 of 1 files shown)


jmduarte:
Example of profiling for CIFAR10 small ResNet-type model. Unfortunately, accuracy loss is still large
```
Keras Accuracy:  0.59
hls4ml Accuracy: 0.03
```

profiling_numerical
profiling_compare

thesps (approved):
Looks good*, thanks for this! 

*The code, not the results 😆

PR #312, opened 2021-04-01
Profiling: Set relative_diff=0 if diff=0 and abs(ymodel)=0

jmduarte:
See title.

- profiling compare() function should set the relative_diff=0 if diff=0 and abs(ymodel)=0. Currently, it's set to 1.


Duchstf (approved):
Thank you! It looks good!

PR #320, opened 2021-04-13
Fix Average 

jmduarte:
- fix formula for average (unless I'm missing something?)


vloncar (approved):
Ouch, that was an ugly copy/paste error. Thanks Javier!

PR #321, opened 2021-04-14
Add set_data_types_from_keras_model() for automatic precision inference

maksgraczyk:
Profiling information gathered e.g. by `hls4ml.model.profiling.numerical()` can be used for setting data types in an HLSModel config heuristically so that the optimal precision configuration is achieved faster.

This PR adds `set_data_types_from_keras_model()` doing that with a Keras model.

### DIFF
--- a/hls4ml/utils/config.py
+++ b/hls4ml/utils/config.py
@@ -3,6 +3,7 @@ import numpy as np
 import h5py
 import json
 import math
+from hls4ml.model.profiling import activations_keras, weights_keras
 from collections import OrderedDict
 
 def create_vivado_config(output_dir='my-hls-test', project_name='myproject',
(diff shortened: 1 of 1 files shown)


thesps:
The test failure is due to the new import from `profiling` that is introduced in utils. Specifically [this line](https://github.com/maksgraczyk/hls4ml/blob/auto-type-setup/hls4ml/model/profiling.py#L3) that imports matplotlib (but the pandas and seaborn import would fail too). You can see in [setup.py](https://github.com/maksgraczyk/hls4ml/blob/auto-type-setup/setup.py#L34-L38) that these packages are 'extras' for profiling and not automatically installed. We want to keep it this way so that users not using profiling can skip those packages. 

For your PR, I think that just means we need to add some more `try: import / except: ImportError` guards around the packages only needed for making the plots (matplotlib, seaborn, pandas). The methods from profiling that you've called should work without those, and wrapping the imports that way should resolve the import error.

maksgraczyk:
If you want to familiarise yourself with the new method, you can have a look at this Jupyter notebook: https://gist.github.com/maksgraczyk/031ad1ec8191e237c2f5008adade2257

maksgraczyk:
I've made 2 extra commits after getting feedback from one of the users: the first one makes sure that only layers actually present in an HLSModel config are processed (it turns out that some Keras layers may not be actually present in an HLSModel config) and the second one allows `set_data_types_from_keras_model()` to be called also from `hls4ml.utils` rather than only `hls4ml.utils.config`.

hls4ml/utils/config.py
```diff
@@ -50,10 +57,265 @@ def _get_precision_from_quantizer(quantizer):
...
+
+        # Assuming that all of type_w, type_b and type_i are FixedPrecisionType objects.
+        n = layer.output_shape[1]
+        max_w = 2 ** (type_w.integer - 1) - 2 ** (type_w.integer - type_w.width)
+        max_i = 2 ** (type_i.integer - 1) - 2 ** (type_i.integer - type_i.width)
+        max_b = 2 ** (type_b.integer - 1) - 2 ** (type_b.integer - type_b.width)
+        min_b = 2 ** (type_b.integer - type_b.width)
+
+        max_val = n * max_w * max_i + max_b
+        min_val = min_b
```
thesps, line 151: I think in the general case the minimum value could come either from the product of weight * input or the bias, depending on their precisions, not always from the bias.

vloncar: bias may not exist (e.g., `Dense(..., use_bias=False)`), in that case we insert zeros and set precision to uint(1), which may throw this function off.

maksgraczyk: I've just pushed the commits addressing both issues.

PR #323, opened 2021-04-16
Make profiling return pre- and post-optimization plots (up to 4 plots in total)

maksgraczyk:
Currently, `hls4ml.model.profiling.numerical()` returns 2 profiling plots, most often showing HLSModel and Keras/PyTorch/... layers with HLSModel data types overlaid. These may be confusing because hls4ml applies optimizations resulting in differences between final HLSModel and Keras/PyTorch/... layers.

This PR attempts to resolve this problem by making `hls4ml.model.profiling.numerical()` return up to 4 plots instead of 2: predominantly weights of HLSModel layers before and after optimizations + activations/outputs of HLSModel/Keras/PyTorch/... layers before and after optimizations.

### DIFF
--- a/hls4ml/model/profiling.py
+++ b/hls4ml/model/profiling.py
@@ -208,6 +208,31 @@ def weights_hlsmodel(model, fmt='longform', plot='boxplot'):
         data = pandas.DataFrame(data)
     return data
 
+
+def activations_hlsmodel(model, X, fmt='summary', plot='boxplot'):
+    if fmt == 'longform':
+        raise NotImplemented
+    elif fmt == 'summary':
+        data = []
+
+    _, trace = model.trace(np.ascontiguousarray(X))
+
+    if len(trace) == 0:
+        raise RuntimeError("HLSModel must have tracing on for at least 1 layer (this can be set in its config)")
+
+    for layer in trace.keys():
+        print("   {}".format(layer))
+
+        if fmt == 'summary':
+            y = trace[layer].flatten()
+            y = abs(y[y != 0])
+
+            data.append(array_to_summary(y, fmt=plot))
+            data[-1]['weight'] = layer
+
+    return data
+
+
 def weights_keras(model, fmt='longform', plot='boxplot'):
     suffix = ['w', 'b']
     if fmt == 'longform':
--- a/hls4ml/model/profiling.py
+++ b/hls4ml/model/profiling.py
@@ -243,7 +243,12 @@ def weights_keras(model, fmt='longform', plot='boxplot'):
         name = layer.name
         weights = layer.get_weights()
         for i, w in enumerate(weights):
-            l = '{}/{}'.format(name, suffix[i])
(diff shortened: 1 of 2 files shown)


hls4ml/model/profiling.py
```diff
@@ -4,8 +4,13 @@
 import numpy as np
 import pandas
 import seaborn as sb
+import uuid
+import os
+import shutil
+from collections import defaultdict
 
 from hls4ml.model.hls_model import HLSModel
+from hls4ml.converters import convert_from_config
```
thesps, line 13: An unintended consequence of this seems to be that we now have a circular import, so when a user does `import hls4ml`, [this](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/model/__init__.py#L6) will fail. It's possible to `from hls4ml.model import profiling` afterwards, but we should see if there is some workaround to avoid an extra step.

hls4ml/model/profiling.py
```diff
@@ -208,15 +227,66 @@ def weights_hlsmodel(model, fmt='longform', plot='boxplot'):
...
+
+def _keras_layer(layer):
+    return layer.get_weights(), ['w', 'b']
+
+
+keras_process_layer_map = defaultdict(lambda: _keras_layer,
+                                      {
+                                          'BatchNormalization': _keras_batchnorm,
+                                          'QBatchNormalization': _keras_batchnorm
+                                      })
```
thesps, line 254: Nice, looks good

hls4ml/model/profiling.py
maksgraczyk: Do you have an example of the user code causing that circular import? I can import hls4ml just fine and tests here seem to pass as well.

thesps: Right, I also see that `import hls4ml` works fine because of the `try: import ...` in `hls4ml/model/__init__.py`, but then I'm finding that using `hls4ml.model.profiling` will throw an error like `AttributeError: module 'hls4ml.model' has no attribute 'profiling'` because that `try` encountered an exception and `profiling` is not imported. If you temporarily move the `import` in that `__init__.py` outside of the `try`, you can see the cause.
How have you been using the `numerical` method after importing?
I find it works to do both:
```
import hls4ml
from hls4ml.model import profiling
```

maksgraczyk: I see what you mean now. I remember encountering the same AttributeError and I had to write `import hls4ml.model.profiling` explicitly.

When I move the import out of the try-except block, I get a circular import error as well. I'll try to fix this.

maksgraczyk: I've moved the problematic convert_from_config import to inside `get_unoptimized_hlsmodel()` function (which is the only place it's used) and this seems to solve the issue.

thesps: Perfect, that fixes it

thesps (approved):

PR #329, opened 2021-04-29
Added support for keras RNN 

chaitanyaPaikara:
Working keras-converter and RNN template code for LSTM layer, will update with GRU code later. Review @vloncar


hls4ml/templates/vivado_template.py
```diff
@@ -71,6 +71,33 @@
...
+    static const unsigned io_type = nnet::{iotype};
+    static const unsigned strategy = nnet::{strategy};
+    static const unsigned reuse_factor = {reuse};
+    typedef {accum_t} accum_t;
+    typedef {bias_t} bias_t;
+    typedef {weight_t} weight_t;
+    template<class x_T, class y_T, class res_T>
+    using product = nnet::product::{product_type}<x_T, y_T, res_T>;
+}};\n"""
+
```
thesps, line 100: Since the contents of these are the same as the Dense config, do we need to define new templates? Can you just pick up the existing Dense template?

hls4ml/templates/vivado/nnet_utils/nnet_common.h
```diff
@@ -32,6 +32,9 @@ namespace nnet {
 enum io_type {io_parallel = 0, io_serial, io_stream};
 enum strategy { latency, resource };
 
+// Activation enum
+enum activ_type {activ_relu = 0, activ_sigmoid, activ_tanh, activ_softmax};
+
```
thesps, line 37: I think we should find a nicer mechanism to enable multiple activation functions from the RNN. One method could be to make classes for the activation functions, and then include in the RNN layer config the object that corresponds to the appropriate activation. 

The aim would be to eliminate the `if (CONFIG_T::activation_type == activation_relu) { relu<...>(...) }` constructions in the HLS and replace it with a single line like:
`CONFIG_T::template activation<...>::activation(...)`

It's similar to what we do now with the `nnet::mult`  - classes [here](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/templates/vivado/nnet_utils/nnet_mult.h), config [here](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/templates/vivado/nnet_utils/nnet_dense.h#L33-L34), and call to product [here](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/templates/vivado/nnet_utils/nnet_dense_latency.h#L96).

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,799 @@
...
+// Resources:
+//  - http://www.wildml.com/2015/09/recurrent-neural-networks-tutorial-part-1-introduction-to-rnns/
+//  - https://github.com/pangolulu/rnn-from-scratch
+// Notes:
+//  - RNN naming conventions adopted from the above links
+//      - newstate = activation(U*input + W*state)
+//      - output   = V*newstate
+//  - If softmax is needed on output, perform *outside* this operations
+template<class data_T, class res_T, typename CONFIG_T, typename ACT_CONFIG_T>
+void simple_rnn(
```
thesps, line 63: It looks like these `simple_rnn` and `simple_rnn_static` methods are not accessible through the hls4ml conversion, only the LSTM and GRU. Is that correct? If so, these implementations should be removed. Or otherwise, they should be made accessible through the conversion flow.

hls4ml/converters/keras/recurrent.py
```diff
@@ -0,0 +1,38 @@
...
+    weights_shape = data_reader.get_weights_shape(layer['name'], 'kernel')
+    recurrent_weights_shape = data_reader.get_weights_shape(layer['name'], 'recurrent_kernel')
+    return_sequences_config = keras_layer['config']['return_sequences']
+    layer['n_sequence'] = input_shapes[0][1]
+    layer['n_sequence_out'] = layer['n_sequence'] if return_sequences_config else 1
+    layer['n_in'] = weights_shape[0]
+    layer['n_out'] = int(weights_shape[1]/div_factor)
+    layer['recurr_n_in']=recurrent_weights_shape[0]
+    layer['recurr_n_out']=recurrent_weights_shape[1]
+    layer['recurr_n_subout']=[recurrent_weights_shape[1]]
```
vloncar, line 34: There are still remnants of sublayer behavior, they should be removed since we don't abandoned that idea.

hls4ml/converters/keras/recurrent.py
```diff
@@ -0,0 +1,38 @@
...
+    if keras_layer['class_name'] == 'LSTM':
+        div_factor = 4
+    elif keras_layer['class_name'] == 'GRU':
+        div_factor = 3
+    else:
+        div_factor = 1
+
+    layer = parse_default_keras_layer(keras_layer, input_names)
+
+    weights_shape = data_reader.get_weights_shape(layer['name'], 'kernel')
```
vloncar, line 25: Instead of inferring attributes from weights, we should extract the number of hidden units and use that to compute the shapes. Keras uses this (`units`), as well as pytorch and onnx (as `hidden_size` attribute).

Also, other LSTM layer properties should be parsed as well. See the [docs](https://keras.io/api/layers/recurrent_layers/lstm/) for all things that can be set. Even if we don't support them all in the current HLS implementation we should be aware of them so we can provide a message to the user if something is not supported as opposed to silently ignoring them and doing something unintended.

hls4ml/converters/keras/recurrent.py
```diff
@@ -0,0 +1,38 @@
...
+    layer['n_sequence'] = input_shapes[0][1]
+    layer['n_sequence_out'] = layer['n_sequence'] if return_sequences_config else 1
+    layer['n_in'] = weights_shape[0]
+    layer['n_out'] = int(weights_shape[1]/div_factor)
+    layer['recurr_n_in']=recurrent_weights_shape[0]
+    layer['recurr_n_out']=recurrent_weights_shape[1]
+    layer['recurr_n_subout']=[recurrent_weights_shape[1]]
+    layer['recurr_n_part'] = 1
+
+    output_shape = [input_shapes[0][0], layer['n_sequence_out'], layer['n_out']]
```
vloncar, line 37: If `return_sequences` is set to `False`, which is the default, the output shape is 2D, not 3D. This will cause the issues when parsing the architecture.

hls4ml/converters/keras_to_hls.py
```diff
@@ -118,6 +118,8 @@ def parse_default_keras_layer(keras_layer, input_names):
         layer['activation'] = keras_layer['config']['activation']
     if 'epsilon' in keras_layer['config']:
         layer['epsilon'] = keras_layer['config']['epsilon']
+    if 'recurrent_activation' in keras_layer['config']:
```
vloncar, line 121: Activations are separated out here because in hls4ml we do them separately from the layer in which they are defined. Since that is not the case for RNNs, `recurrent_activation` should not be parsed here, but rather in the appropriate handler.

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
...
+    def initialize(self):
+        shape = [self.attributes['n_sequence_out'],int(self.attributes['recurr_n_out']/4)]
+        dims = ['N_SEQUENCE_OUT_{}'.format(self.index), 'N_LAYER_{}'.format(self.index)]
+        compression = self.model.config.get_compression(self)
+        reuse_factor = self.model.config.get_reuse_factor(self)
+        self.reuse_factor_recr = reuse_factor
+        if self.model.config.is_resource_strategy(self):
+            if self.model.config.backend.name == 'Vivado':
+                self.model.config.backend.set_closest_reuse_factor(self)
+                self.model.config.backend.set_closest_reuse_factor_recr(self)
```
vloncar, line 1135: Since this function changes the object passed to it and doesn't return values that are assigned to an object, it would be better to extend `set_closest_reuse_factor` and not introduce the `_recr` variant and add `recr` parameter to it.

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
         else:
             self.set_attr('strategy', 'latency')
 
+class LSTM(Layer):
+    def initialize(self):
+        shape = [self.attributes['n_sequence_out'],int(self.attributes['recurr_n_out']/4)]
```
vloncar, line 1127: Again, this depends on the `return_sequences` parameter

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
...
+        params['strategy'] = self.get_attr('strategy')
+
+        lstm_config = self._config_template[0].format(**params)
+
+
+        lstm_params = self._default_config_params()
+        lstm_params['n_in'] = self.get_output_variable().dim_names[1] + ' * 3'
+        lstm_params['table_size'] = 1024
+        lstm_params['type'] = self.get_attr('recurrent_activation')
+        lstm_params['table_t'] = 'ap_fixed<18,8>'
```
vloncar, line 1184: `table_size` and `table_t` should be configurable, not hardcoded, especially not to `ap_fixed` string

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
...
+        lstm_act_config = self._config_template[2].format(**lstm_params)
+
+        act_params = self._default_config_params()
+        act_params['n_in'] = self.get_output_variable().dim_names[1]
+        act_params['table_size'] = 1024
+        act_params['type'] = self.get_attr('activation')
+        act_params['table_t'] = 'ap_fixed<18,8>'
```
vloncar, line 1191: Also here

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
...
+            self.get_attr('recurrent_activation'), self.index)
+        params['act_t'] = '{}_config{}'.format(
+            self.get_attr('activation'), self.index)
+        gru_config = self._config_template[0].format(**params)
+
+        gru_params = self._default_config_params()
+        gru_params['n_in'] = self.get_output_variable().dim_names[1] + ' * 2'
+        gru_params['table_size'] = 1024
+        gru_params['type'] = self.get_attr('recurrent_activation')
+        gru_params['table_t'] = 'ap_fixed<18,8>'
```
vloncar, line 1268: Same

hls4ml/model/hls_layers.py
```diff
@@ -1122,6 +1122,181 @@ def initialize(self):
...
+        gru_act_config = self._config_template[2].format(**gru_params)
+
+        act_params = self._default_config_params()
+        act_params['n_in'] = self.get_output_variable().dim_names[1]
+        act_params['table_size'] = 1024
+        act_params['type'] = self.get_attr('activation')
+        act_params['table_t'] = 'ap_fixed<18,8>'
```
vloncar, line 1275: Same

hls4ml/model/hls_layers.py
```diff
@@ -1812,6 +1987,8 @@ def _get_transforms_config(self, params):
     'Conv2D'                 : Conv2D,
     'BinaryConv2D'           : Conv2D,
     'QConv2D'                : Conv2D,
+    'LSTM'                   : LSTM,
```
vloncar, line 1990: Please put these at the end instead of the middle of the list of CNN layers

hls4ml/templates/vivado_template.py
```diff
@@ -340,6 +413,9 @@
 batchnorm_function_template = 'nnet::normalize<{input_t}, {output_t}, {config}>({input}, {output}, {scale}, {bias});'
 conv1d_function_template = 'nnet::conv_1d_{data_format}<{input_t}, {output_t}, {config}>({input}, {output}, {w}, {b});'
 conv2d_function_template = 'nnet::conv_2d_{data_format}<{input_t}, {output_t}, {config}>({input}, {output}, {w}, {b});'
+lstm_function_template = 'nnet::lstm_loop<{input_t}, {output_t}, {config}>({input}, {output}, {w}, {wr}, {b}, {br});'
```
vloncar, line 416: Please put these at the end instead of the middle of the list of CNN layers

hls4ml/templates/vivado_template.py
```diff
@@ -361,6 +437,7 @@
 batchnorm_include_list = ['nnet_utils/nnet_batchnorm.h', 'nnet_utils/nnet_batchnorm_stream.h']
 conv1d_include_list = ['nnet_utils/nnet_conv1d.h', 'nnet_utils/nnet_conv1d_stream.h']
 conv2d_include_list = ['nnet_utils/nnet_conv2d.h', 'nnet_utils/nnet_conv2d_stream.h']
+lstm_include_list = ['nnet_utils/nnet_recurrent.h']
```
vloncar, line 440: Please put these at the end instead of the middle of the list of CNN layers

hls4ml/templates/vivado_template.py
```diff
@@ -379,6 +456,9 @@ def __init__(self):
         self.register_templates('BatchNormalization'     , batchnorm_function_template,   batchnorm_config_template, batchnorm_include_list)
         self.register_templates('Conv1D'                 , conv1d_function_template,      [conv1d_config_template, conv_mult_config_template], conv1d_include_list)
         self.register_templates('Conv2D'                 , conv2d_function_template,      [conv2d_config_template, conv_mult_config_template], conv2d_include_list)
+        self.register_templates('LSTM'                   , lstm_function_template,        [lstm_config_template, rnn_mult1_config_template, activ_config_recr_template, activ_config_template, rnn_mult2_config_template], lstm_include_list)
```
vloncar, line 459: Please put these at the end instead of the middle of the list of CNN layers

hls4ml/templates/vivado_template.py
```diff
@@ -399,9 +479,9 @@ def __init__(self):
         self.register_templates('Resize'                 , resize_function_template,      resize_config_template, resize_include_list)
         self.register_templates('Transpose'              , transpose_function_template,   transpose_config_template, transpose_include_list)
         self.register_templates('GarNet'                 , garnet_function_template,      garnet_config_template, garnet_include_list)
-        self.register_templates('GarNetStack'            , garnet_stack_function_template,garnet_stack_config_template, garnet_include_list)        
-    
-    def get_valid_reuse_factors(self, layer):
+        self.register_templates('GarNetStack'            , garnet_stack_function_template,garnet_stack_config_template, garnet_include_list)    
+
+    def get_valid_reuse_factors(self, layer, recr = False):
```
vloncar, line 484: I mentioned this before, I think you can get away without `recr` parameter. Or refactor this function so that it accepts `n_in` and `n_out` parameters and move their computation to the caller

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,851 @@
...
+        relu<typename CONFIG_T::state_t, typename CONFIG_T::state_t, ACT_CONFIG_T>(rawstate, newstate);
+    }
+    else if (CONFIG_T::activation_type == activ_sigmoid){
+      sigmoid<typename CONFIG_T::state_t, typename CONFIG_T::state_t, ACT_CONFIG_T>(rawstate, newstate);
+    }
+    else if (CONFIG_T::activation_type == activ_tanh){
+        tanh<typename CONFIG_T::state_t, typename CONFIG_T::state_t, ACT_CONFIG_T>(rawstate, newstate);
+    }
+
+    std::cout << "Activated State: [ "; for (int ii = 0; ii < CONFIG_T::n_state; ii++) std::cout << newstate[ii] << " "; std::cout << "]" << std::endl;
```
vloncar, line 133: This is not useful unless we are debugging the implementation itself so it should be removed.

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,851 @@
...
+	    res_T     h_newstate[CONFIG_T::n_state],
+	    res_T     s_newstate[CONFIG_T::n_state],
+	    typename CONFIG_T::weight_t     param  [CONFIG_T::n_state*4*CONFIG_T::n_in],
+	    typename CONFIG_T::weight_t     param_r[CONFIG_T::n_state*4*CONFIG_T::n_state],
+	    typename CONFIG_T::bias_t     param_b[CONFIG_T::n_state*4],
+            typename CONFIG_T::bias_t     param_br[CONFIG_T::n_state*4]
+	    ) {
+  // Initialize the state variable -- will maintain state between function calls
+  # ifndef __SYNTHESIS__
+  std::cout << "S Pre-State: [ "; for (int ii = 0; ii < CONFIG_T::n_state; ii++) std::cout << s_newstate[ii] << " "; std::cout << "]" << std::endl;
```
vloncar, line 290: Same as above

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,851 @@
...
+      reset_state = false;
+    }
+    if (CONFIG_T::n_sequence_out == 1)
+      for(int i=0; i<(CONFIG_T::n_state); i++){
+        res[i] = h_newstate[i];
+    }
+}
+
+template<class data_T, class res_T, typename CONFIG_T>
+  void lstm_loop(
```
vloncar, line 503: I would prefer if this was called just `lstm` or `lstm_stack` and then the loop iteration should be `lstm_loop`

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,851 @@
...
+      h_state[iacc] =  (res_T)(tmpres_h[iacc]*(1-tmpres_zr[iacc]) + h_state[iacc]*tmpres_zr[iacc]);
+      h_newstate[iacc] = h_state[iacc];
+    }
+  // # ifndef __SYNTHESIS__
+  //   std::cout << "Post-State: h [ "; for (int ii = 0; ii < CONFIG_T::n_state; ii++) std::cout << h_newstate[ii] << " "; std::cout << "]" << std::endl;
+  // # endif
+}
+
+template<class data_T, class res_T, typename CONFIG_T>
+  void gru_loop(
```
vloncar, line 765: Same for this

vloncar (changes requested):
Thanks for updating the previous PR to the latest code! This is a great start

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
chaitanyaPaikara: Yeah, we can remove them as people hardly use basic RNN layers.

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
chaitanyaPaikara: This was part of the basic RNN layer. Will be removing those functions as they are not used anyway.

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
chaitanyaPaikara: changed to gru_stack

hls4ml/templates/vivado/nnet_utils/nnet_recurrent.h
chaitanyaPaikara: changed to lstm_stack

thesps:
Hi @chaitanyaPaikara, let's only mark review items as "resolved" when the code changes that address the issue are pushed please.

hls4ml/templates/vivado/nnet_utils/nnet_common.h
chaitanyaPaikara: Followed the same method used for `nnet::mult` and passing objects for appropriate activation now.

hls4ml/converters/keras/recurrent.py
chaitanyaPaikara: I followed the same format used for other layers which were using weight_shape to figure out the shape. I've added all the architecture related properties of the RNN layer to the layer dictionary. Only properties associated with training aren't added.

chaitanyaPaikara:
Hi @thesps and @vloncar, I've pushed the latest changes addressing the above issues. Please let me know once you've reviewed these changes.

chaitanyaPaikara:
Hi @vloncar and @thesps, could you please review the added changes and let me know if there are other issues.

hls4ml/converters/keras/recurrent.py
thesps: What about adding the parsing of all the other layer properties, could you add that @chaitanyaPaikara ?

vloncar: `return_state`, `go_backwards`, `stateful`, `time_major`, `reset_after`... Preferably the conversion should check the unsupported parameters in `initialize` and raise exceptions in case of configuration we can't handle yet.

vloncar:
Done in #560. Closing

PR #332, opened 2021-05-07
Line Buffer Convolution Implementation

Keb-L:
This PR introduces an alternative stream-based implementation for Convolutional (and Pooling) layers that reduces resource utilization compared to the current implementation. Additionally, this PR introduces some new YAML configuration parameters (ConvImplementation, BramFactor, TargetCycles). 

The new convolution implementation, called "Line Buffer", uses shift registers to keep track of the last <kernel height - 1> rows of input pixels and maintains a shifting snapshot of the convolution kernel. This implementation supports non-square kernels and strides greater than the width of the kernel. See the following animation for a demonstration of the Line Buffer implementation. 
Phil_Animation_cropped
(Input image is at the top-left, output image on the bottom left. The top right shows the internal state of the shift registers and convolutional kernel. The red square indicates the current pixels contained within the convolutional kernel)

A YAML configuration has been added (see YAML Configuration additions) to mix-and-match between the two convolution implementations. The line buffer implementation has been set to be the default. Top-level functions have been changed to be a switch between the two convolution implementations.

Updated layers:
- `Conv1D`, `Conv2D`
- `DepthwiseConv1D`, `DepthwiseConv2D`
- `Pooling1D`, `Pooling2D`

YAML Configuration additions:
- `ConvImplementation`
  - Flag to switch between Line Buffer (added) and Encoded (current) implementations. Functions similarly to the softmax implementation parameter. Values: `linebuffer` (default), `encoded`
- `BramFactor`
  - Threshold value to import weights as BRAM ports. Importing weights as BRAM ports reduces synthesis time. Defaults to infinity.
BramFactor
- `TargetCycles`
  - Balances throughput of `Conv1D`, `Conv2D`, `Dense` layers by tuning the Reuse Factor to achieve specified latency in clock cycles. Reduces resource utilization by slowing down fast layer (experimental). Defaults to None (unused). 
TargetCycles

Current limitations:
- Target Cycles still requires further tuning to be more broadly applicable. It works in most cases, but is not guaranteed to achieve the specified latency in all cases. Further studies are needed to tune the RF computation.
- Pooling using kernels of height 1 do not work with line buffer implementation. This will be fixed in a future PR.

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_stream.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_stream.h
@@ -23,7 +23,7 @@ void compute_scaled_indices_1d(
 }
 
 template<class data_T, class res_T, typename CONFIG_T>
-void conv_1d_cl2(
+void conv_1d_encoded_cl(
     hls::stream<data_T> &data,
     hls::stream<res_T>  &res,
     typename CONFIG_T::weight_t weights[CONFIG_T::filt_width * CONFIG_T::n_chan * CONFIG_T::n_filt],
@@ -53,12 +53,12 @@ void conv_1d_cl2(
             #pragma HLS PIPELINE II=CONFIG_T::reuse_factor
         }
         compute_scaled_indices_1d<data_T, CONFIG_T>(i_iw, pixel_idx);
-        compute_output<data_T, res_T, CONFIG_T>(data.read(), data_window, res, res_pack, outputs_ready, weights, biases, pixel_idx);
+        compute_encoded_output<data_T, res_T, CONFIG_T>(data.read(), data_window, res, res_pack, outputs_ready, weights, biases, pixel_idx);
     }
 }
 
 template<class data_T, class res_T, typename CONFIG_T>
-void conv_1d_cl(
+void conv_1d_buffer_cl(
     hls::stream<data_T> &data,
     hls::stream<res_T>  &res,
     typename CONFIG_T::weight_t weights[CONFIG_T::filt_width * CONFIG_T::n_chan * CONFIG_T::n_filt],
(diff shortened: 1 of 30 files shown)


hls4ml/templates/vivado/nnet_utils/nnet_pooling_stream.h
```diff
@@ -2,8 +2,10 @@
 #define NNET_POOLING_STREAM_H_
 
 #include "utils/x_hls_utils.h"
+#include "ap_shift_reg.h"
```
Duchstf, line 5: Might not work for Python API if `ap_shift_reg.h` is not included in vivado template.

Keb-L: Added the ap_shift_reg.h header file from Vivado HLS library to ap_types. The Python API should now work with the compile() function without throwing an error.

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_stream.h
```diff
@@ -53,9 +53,45 @@ void conv_1d_cl(
...
+{
+    #pragma HLS inline region
+    switch(CONFIG_T::implementation){
+        case conv_implementation::linebuffer:
+            conv_1d_buffer_cl<data_T, res_T, CONFIG_T>(data, res, weights, biases);
+            break;
+        case conv_implementation::encoded:
+            conv_1d_encoded_cl<data_T, res_T, CONFIG_T>(data, res, weights, biases);
+            break;
+    }
```
Duchstf, line 93: Hello, I'm wondering if specifying `io_parallel` (non-streaming algorithm) would also be one of the options here?

hls4ml/templates/vivado_template.py
```diff
@@ -465,6 +476,30 @@ def set_closest_reuse_factor(self, layer):
...
+        if targ_cycles is not None:
+            if layer.get_attr('class_name') == 'Dense': 
+                kernel_multiplies = layer.get_attr('n_out')
+            elif layer.get_attr('class_name') == 'Conv1D':  
+                kernel_multiplies = layer.get_attr('out_width')
+            elif layer.get_attr('class_name') == 'Conv2D': 
+                kernel_multiplies = layer.get_attr('out_height') * layer.get_attr('out_width')
+            else: 
+                print('Target clock unsupported layer')
+                return
```
thesps, line 493: QKeras layers are not supported by this. They have `'class_name'` attributes like `QDense`, `QConv2D`. Is it deliberate that they are not supported? It seems like it should apply to those too. 
If they should be supported, the condition would become either
`if layer.__class__.__name__ == 'Dense'` or `if 'Dense' in layer.get_attr('class_name')`.
It should support the `QConv2DBatchNorm` layer too, if it does apply to QKeras layers.

Keb-L: If QKeras layers end up converting to the Dense, Conv1D, or Conv2D layer implementations, then the target cycles parameter should work for them. I'll make some edits to support the QKeras layers.

vloncar:
@Keb-L can you also merge https://github.com/Keb-L/hls4ml/pull/4?

hls4ml/templates/vivado_template.py
vloncar: `class_name` attribute comes from the converter and corresponds to the Keras layer class name, it may not match the hls4ml class name. A safer check would be e.g., `'Dense' in layer.__class__.__name__'`

vloncar (approved):

PR #338, opened 2021-05-14
Special case for a Conv with a single output pixel

jmduarte:
- Special case for a Conv with a single output pixel
- Without this, RTL cosimulation or running on FPGA freezes / hangs forever (probably because of stream size of 1)
- Test script: https://gist.github.com/jmduarte/66cb7b32c9a7a71b10b5dc310947475e

### DIFF
--- a/hls4ml/model/optimizer/__init__.py
+++ b/hls4ml/model/optimizer/__init__.py
@@ -10,6 +10,7 @@ from hls4ml.model.optimizer.passes.fuse_biasadd import FuseBiasAdd
 from hls4ml.model.optimizer.passes.conv_same_pad import InsertZeroPaddingBeforeConv1D
 from hls4ml.model.optimizer.passes.conv_same_pad import InsertZeroPaddingBeforeConv2D
 from hls4ml.model.optimizer.passes.pointwise import OptimizePointwiseConv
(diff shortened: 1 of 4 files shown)


jmduarte:
@vloncar I also struggled to reproduce the issue, but now I have an example of it here: https://gist.github.com/jmduarte/66cb7b32c9a7a71b10b5dc310947475e

It turns out this only affects the Encoded ConvImplementation. I was able to get it to "break cosim" with a 3x3 kernel size and 3x3 input image, stride = 3, "same" padding. It also seems to affect different reuse factors and Latench and Resource strategies.

There are 3 cases you can test with this code on this branch:
1. LineBuffer works even skipping optimizer:
```
skip_opt = True
conv_impl = 'LineBuffer'
```
2. Encoded breaks in cosim when skipping optimizer:
```
skip_opt = True
conv_impl = 'Encoded'
```
3. Encoded works in cosim with the optimizer:
```
skip_opt = False
conv_impl = 'Encoded'
```

Lastly, I also rebased this branch to master.

Back to your original question, yes, it would be nice to support that case if we deem it necessary (or even if it's just a simple optimization?). Would we also want to make this optimizer only match ConvImplementation == Encoded? Or do we think it could be an improvement even over LineBuffer?

PR #339, opened 2021-05-14
support other quantizers in QConv2DBatchnorm

jmduarte:
- simple fix


nicologhielmetti (commented):
@jmduarte I think it's fine, since the Conv2DBatchnorm layer is basically a Conv2D layer in hls4ml

PR #341, opened 2021-05-14
Support special case of broadcasting during Merge layers

jmduarte:
- Support special case of broadcasting during Merge layers
- The case is `Tensor(shape=[H, W, C]) + Tensor(shape=[1, 1, C]) = Tensor(shape=[H, W, C])`
- Was used for skip connections in small ResNet models
- Only for `io_stream`

We also have a use case for `io_parallel` and `Tensor(shape=[N, C]) + Tensor(shape=[N, 1]) = Tensor(shape=[N, C])` so it'd be great to generalize this PR to handle that as well

### DIFF
--- a/hls4ml/model/optimizer/passes/repack_stream.py
+++ b/hls4ml/model/optimizer/passes/repack_stream.py
@@ -24,14 +24,47 @@ class Repack(Layer):
     def config_cpp(self):
         return None
 
+class Broadcast(Layer):
+    ''' Inserted between layers for broadcasting.'''
+
+    def initialize(self):
+        shape = self.attributes['target_shape']
+        if shape[0] is None:
+            shape = shape[1:]
+        dims = ['N_SIZE_{}_{}'.format(i, self.index) for i in range(1, len(shape) + 1)]
+        self.add_output_variable(shape, dims)
+
+    def function_cpp(self):
+        params = self._default_function_params()
+        return [self._function_template.format(**params)]
+
+    def config_cpp(self):
+        params = self._default_config_params()
+        params['in_height'] = self.get_input_variable().shape[0]
+        params['in_width'] = self.get_input_variable().shape[1]
+        params['n_chan'] = self.get_input_variable().shape[2]
+        params['n_dupl'] = int(np.prod(self.get_output_variable().shape)/np.prod(self.get_input_variable().shape))
+        return self._config_template.format(**params)
+
 repack_function_template = 'nnet::repack_stream<{input_t}, {output_t}, {size}>({input}, {output});'
 repack_include_list = ['nnet_utils/nnet_stream.h']
 
+broadcast_function_template = 'nnet::broadcast_stream<{input_t}, {output_t}, {config}>({input}, {output});'
+broadcast_config_template = """struct config{index} : nnet::broadcast_config {{
+static const unsigned in_width = {in_width};
+static const unsigned in_height = {in_height};
+static const unsigned n_chan = {n_chan};
+static const unsigned n_dupl = {n_dupl};
+}};\n"""
+broadcast_include_list = ['nnet_utils/nnet_stream.h']
+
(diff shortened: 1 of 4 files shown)


hls4ml/model/optimizer/passes/repack_stream.py
```diff
@@ -53,3 +86,38 @@ def transform(self, model, node):
...
         return True
+
+class BroadcastStream(OptimizerPass):
+    def match(self, node):
+        if node.__class__.__name__ == 'Merge':
+            try: 
+                inp1 = node.get_input_variable(node.inputs[0])
+                inp2 = node.get_input_variable(node.inputs[1])
+                return inp1.shape != inp2.shape
+            except:
```
vloncar, line 97: Can you remind me why does this have to be in a try/except block?

hls4ml/model/optimizer/passes/repack_stream.py
```diff
@@ -53,3 +86,38 @@ def transform(self, model, node):
...
+                return False
+        else:
+            return False
+        
+    def transform(self, model, node):
+        if model.config.backend.name not in ['Vivado'] or \
+            model.config.get_config_value('IOType') != 'io_stream':
```
vloncar, line 104: Doesn't this also affect io_parallel?

hls4ml/templates/vivado/nnet_utils/nnet_stream.h
```diff
@@ -90,6 +98,23 @@ void repack_stream(hls::stream<data_T> &data, hls::stream<res_T> &res) {
     }
 }
 
+template<class data_T, class res_T, typename CONFIG_T>
+void broadcast_stream(hls::stream<data_T> &data, hls::stream<res_T> &res) {
+    BroadcastLoop: for (int i = 0; i < CONFIG_T::in_height * CONFIG_T::in_width * CONFIG_T::n_chan / data_T::size; i++) {
+        #pragma HLS PIPELINE
+        data_T in_data = data.read();
+	for (int j = 0; j < CONFIG_T::n_dupl; j++) {
```
vloncar, line 106: Can you replace tabs with spaces?

vloncar (commented):
Thanks Javier. Looks good.

hls4ml/model/optimizer/passes/repack_stream.py
jmduarte: It may just be an ugly way to ensure that `len(node.inputs) == 2`, but since `Merge` layer already [asserts this](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/model/hls_layers.py#L1412), maybe this try/except block is not needed... I'll test and remove

hls4ml/templates/vivado/nnet_utils/nnet_stream.h
jmduarte: Yes

hls4ml/model/optimizer/passes/repack_stream.py
jmduarte: So I believe this transformation is skipped if IOType is io_parallel given the above condition (i.e. it returns False immediately if IOType is io_parallel). Basically, the above condition is equivalent to:
```python
if not (model.config.backend.name == 'Vivado' and model.config.get_config_value('IOType') == 'io_stream'): return False
```

Are you saying you want it to also be applied in the case of io_parallel? That would require some additional modifications, and is currently planned to be done in a follow-up PR to generalize this broadcasting to more cases + io_parallel and io_stream.

vloncar: Sorry for my poor choice of words, I was indeed asking if the broadcasting issue affects `io_parallel` as well, which I thought it does. I agree that support for it should not be a part of this PR and a follow-up PR is a better way forward.

vloncar (approved):

PR #349, opened 2021-06-23
VivadoAccelerator backend [WIP]

nicologhielmetti:


scripts/hls4ml
```diff
@@ -167,7 +167,7 @@ def build(args):
         elif backend == 'Mentor':
             raise NotImplementedError
         else:
-            raise Exception('Backend values can be [Vivado, Intel, Mentor]')
+            raise Exception('Backend values can be [Vivado, Intel, Mentor, Pynq]')
```
jmduarte, line 170: `Pynq` -> `VivadoAccelerator`

scripts/hls4ml
```diff
@@ -157,7 +157,7 @@ def build(args):
     # Check if vivado_hls is available
     if 'linux' in sys.platform or 'darwin' in sys.platform:
         backend = yamlConfig.get('Backend', 'Vivado')
-        if backend == 'Vivado':
+        if backend in ['Vivado', 'Pynq']:
```
jmduarte, line 160: `Pynq` -> `VivadoAccelerator`

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -0,0 +1,357 @@
...
+        '''
+        filedir = os.path.dirname(os.path.abspath(__file__))
+        if self.vivado_accelerator_config.get_interface() == 'axi_stream':
+            copyfile(os.path.join(filedir, '../templates/vivado_accelerator/' +
+                                  self.vivado_accelerator_config.get_device() + '/tcl_scripts/axi_stream_design.tcl'),
+                     '{}/design.tcl'.format(model.config.get_output_dir()))
+        else: 
+            copyfile(os.path.join(filedir, '../templates/vivado_accelerator/' +
+                                  self.vivado_accelerator_config.get_device() + '/tcl_scripts/axi_lite_design.tcl'),
+                     '{}/design.tcl'.format(model.config.get_output_dir()))
```
thesps, line 328: Shouldn't the filename of the `.tcl` script be derived somehow from the contents of the `supported_devices.json`?
And also with a check whether that interface is supported for that device against the contents of the file.

hls4ml/writer/vivado_writer.py
```diff
@@ -134,7 +134,7 @@ def _make_array_pragma(variable):
 
     @staticmethod
     def _make_stable_pragma(variable):
-        template = '#pragma HLS STABLE variable={name}'
+        template = '//#pragma HLS STABLE variable={name}'
```
thesps, line 137: I think it would be better to remove the line that calls `_make_stable_pragma` for this, rather than write a commented out line to the HLS.

hls4ml/templates/vivado_accelerator_config.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def get_board(self):
+        return self.board
+
+    def get_driver_path(self):
+        return '../templates/vivado_accelerator/' + self.board + '/' + self.driver + '_drivers/' + \
+               self.get_driver_file()
+
+    def get_driver_file(self):
+        driver_ext = '.py' if self.driver is 'python' else '.h'
```
jmduarte, line 126: for some reason, I needed to change this line to 
```python
driver_ext = '.py' if self.driver == 'python' else '.h'
```
to get it to work for me. Otherwise, it was setting `driver_ext = '.h'` even though `self.driver = 'python'`

thesps (changes requested):

test/keras-to-hls.sh
```diff
@@ -41,10 +47,14 @@ function print_usage {
    echo "      Prints this help message."
 }
 
-while getopts ":x:c:sr:g:t:d:y:h" opt; do
+while getopts ":x:b:B:c:s:r:g:t:d:y:h" opt; do
```
thesps, line 50: Instead of `:x:b:B:c:s:r:g:t:d:y:h` this needs to be `:x:b:B:c:sr:g:t:d:y:h` (no colon after `s`). See the output of the Jenkins log Keras to HLS stage, or when you execute these locally:
```
./keras-to-hls.sh  -s KERAS_3layer
No models specified.
./keras-to-hls.sh  -s KERAS_conv1d_small
No models specified.
./keras-to-hls.sh  -s KERAS_conv2d_model
No models specified.
./keras-to-hls.sh  -s jetTagger_Conv2D_Small
No models specified.
./keras-to-hls.sh  -s jetTagger_Conv2D_Small_NoBatchNorm
No models specified.
```

The `-s` option for io_stream takes no arguments, but adding the `:` after `s` expects an argument, so the model name is being treated as the argument to `-s`, hence the "No models specified" message.

thesps (changes requested):

thesps:
The recent update of handling the backend through the hls4ml script, and then propagating that through testing is super but needs a little more work. See one inline review comment about the `keras-to-hls.sh` script. 

Then the major thing: it doesn't seem like the backend is being propagated through the chain completely as intended. I didn't find out where exactly. The test that you added using the new backend converts to an HLS project, the `hls4ml_config.yml` file that's generated does contain `Backend: VivadoAccelerator` and the `AcceleratorConfig` section, but the firmware and project files correspond to `Vivado` backend rather than `VivadoAccelerator` (no `myproject_axi*` files, BD .tcl scripts, drivers etc).  

And a third thing: I think the new information in the testing arguments also needs to propagate to the project directory name. e.g. right now the new test:
`KERAS_3layer b:pynq-z2 B:VivadoAccelerator s:Resource` creates a project in the directory
`KERAS_3layer-xcvu9p_flgb2104_2_e-c5-io_parallel-rf1-ap_fixed_16_6_-Resource-2019.2`.
One thing is that the name is not super helpful to understand that actually it's a VivadoAccelerator test on a pynq-2z, but the other is that it would clash if we had a test like:
`KERAS_3layer s:Resource`, which would output to the same directory.

So we should add the backend to the project directory name. And I think even though the VivadoAccelerator backend can handle being given `Board: pynq-z2`, `Part: xcvu9p-flgb2104-2-e`, for this testing I think we should specify both properly to make the names easier to understand. (So `KERAS_3layer b:pynq-z2 B:VivadoAccelerator s:Resource` -> `KERAS_3layer b:pynq-z2 B:VivadoAccelerator x:xc7z020clg400-1 s:Resource` for clarity + including the backend name in the directory)

Issue #325, opened 2021-04-27
Report extensions


thesps:
It would be nice to extend [vivado_report](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/report/vivado_report.py) utility to include in the extracted results:

- per-layer resource & latency breakdown
- Post logic synthesis resources

Duchstf:
This is something of interest for me now, so I'll take it.

thesps:
Great! I made the WIP/hacky PR #342 with the idea to make the HLS generate easier to parse reports that may be relevant.

Duchstf:
Hello, I'm looking into this now, could you elaborate your ideas a bit:

> Basically, I think we'll just need to add another type of template to the backend templates. Or, extend the function templates we have now to be able to define both the function definition as well as the string to call it.


Wouldn't adding another type of template too time-consuming? and may cause more confusions? and I don't really understand "extend the function templates we have now to be able to define both the function definition as well as the string to call it."

thesps:
What I mean is, right now our templates are the strings needed to call the function, e.g. for Dense:
`dense_function_template = 'nnet::dense<{input_t}, {output_t}, {config}>({input}, {output}, {w}, {b});'`, which might become this in a project:
`nnet::dense<input_t, layer2_t, config2>(fc1_input, layer2_out, w2, b2);`

The thing that worked in the hacky PR to get sensible names in the report is to define a new function with a unique name, and no template parameters, that just calls the specific `nnet::<layer>` function for that layer, e.g.
```
void fc1(input_t* fc1_input, layer2_t* layer2_out, weight2_t* w2, bias2_t* b2){
    #pragma HLS inline off
    nnet::dense<input_t, layer2_t, config2>(fc1_input, layer2_out, w2, b2); // fc1
}
```
So the writer would now need to generate the new function definition, and the call to that function, and the call to the `nnet::<layer> function`. You _could_ have templates for each of those parts (which is what the hacky PR does), but it kind of doesn't make sense to hardcode them all, since the function signature defines how the function should be called afterwards, too.
i.e. if the definition template is: 
`'void {name}({input_t}* {input}, {output_t}* {output}, {weight_t}* {w}, {bias_t}* {b})'`
[...]

PR #355, opened 2021-07-08
[WIP] Added naive ability to parse vivado synthesis reports

mackncheesiest:
I noticed that the current `hls4ml.report.parse_vivado_report` doesn't support parsing the resulting `vivado_synth.rpt` if your model was built with `vsynth=True`. I'm currently in the process of trying to sweep a number of different designs for the purpose of determining whether they're feasible for my use case, and for that it is much more useful to have at least the post synthesis LUT and FF estimates as they are much more accurate than the Vivado HLS estimates.

However, after looking at the report's format, it made sense why `hls4ml` hasn't parsed it yet -- it's not necessarily output in the most machine readable format, and there don't seem to be other arguments to `report_utilization` that improve that.

Looking around online, [edalize](https://github.com/olofk/edalize/) has [an approach](https://github.com/olofk/edalize/blob/master/edalize/vivado_reporting.py) that uses [pyparsing](https://pypi.org/project/pyparsing/), but that seemed like quite a bit to either (i) pull in the whole `edalize` package as a dependency and use it here or (ii) go through their inheritance hierarchy, figure out how they're pulling off the parsing and repeat that internally here

So, for my own needs, I just threw together this small set of changes, and I figured I'd open up this PR to
- have at least something available if anyone else would benefit from similar functionality
- get some feedback on if anyone can think of a different/smarter/less hard-coded approach that I might be missing (or if the `edalize` approach seems like the best option after all)

### DIFF
--- a/hls4ml/report/vivado_report.py
+++ b/hls4ml/report/vivado_report.py
@@ -158,6 +158,23 @@ def parse_vivado_report(hls_dir):
     else:
         print('Synthesis report not found.')
 
+    vivado_syn_file = hls_dir + '/vivado_synth.rpt'
+    if os.path.isfile(vivado_syn_file):
+        vivado_synth_rpt = {}
+        with open(vivado_syn_file) as f:
+            for line in f.readlines():
+                if 'CLB LUTs' in line:
+                    vivado_synth_rpt['LUT'] = line.split('|')[2].strip()
+                elif 'CLB Registers' in line:
+                    vivado_synth_rpt['FF'] = line.split('|')[2].strip()
+                elif 'RAMB18 ' in line:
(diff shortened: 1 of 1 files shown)


vloncar:
Hi @mackncheesiest, thanks for this, we were also thinking about what the best solution to parsing the report would be. One alternative to edalize that could be explored is to export as a spreadsheet and parse the excel file with [xlrd](https://pypi.org/project/xlrd/). Should be simpler, given that we will work with a tabular report. Quartus backend (in development) also relies on external packages to parse reports, so maybe external packages for reports are inevitable. For me personally this naive approach is sufficient, for now, until we add `-hierarchical` flag. What do others think?

mackncheesiest:
Related: #325

PR #358, opened 2021-07-15
Support Keras Permute layer

jmduarte:
- add support for Keras Permute layer
- fix some minor bugs in transpose implementation
- code to test here: https://gist.github.com/jmduarte/d0b0036fffef32820ec2112e095bb7b7

### DIFF
--- a/hls4ml/converters/keras/core.py
+++ b/hls4ml/converters/keras/core.py
@@ -120,3 +120,18 @@ def parse_batchnorm_layer(keras_layer, input_names, input_shapes, data_reader, c
         layer['n_filt']=input_shapes[0][3]
 
     return layer, [shape for shape in input_shapes[0]]
+
+@keras_handler('Permute')
+def parse_permute_layer(keras_layer, input_names, input_shapes, data_reader, config):
+    assert(keras_layer['class_name'] == 'Permute')
+
+    layer = parse_default_keras_layer(keras_layer, input_names)
+
+    layer['class_name'] = 'Transpose'
+    perm = keras_layer['config']['dims']
+    layer['perm'] = perm
+
+    output_shape = input_shapes[0]
+    output_shape[perm[0]], output_shape[perm[1]] = output_shape[perm[1]], output_shape[perm[0]]
+
+    return layer, output_shape
--- a/hls4ml/converters/keras/core.py
+++ b/hls4ml/converters/keras/core.py
@@ -120,19 +120,3 @@ def parse_batchnorm_layer(keras_layer, input_names, input_shapes, data_reader, c
         layer['n_filt']=input_shapes[0][3]
 
     return layer, [shape for shape in input_shapes[0]]
-
-@keras_handler('Permute')
-def parse_permute_layer(keras_layer, input_names, input_shapes, data_reader, config):
-    assert(keras_layer['class_name'] == 'Permute')
-
-    layer = parse_default_keras_layer(keras_layer, input_names)
-
-    layer['class_name'] = 'Transpose'
-    swap = keras_layer['config']['dims']
-    perm = list(range(len(input_shapes[0])))
-    perm[swap[0]], perm[swap[1]] = perm[swap[1]], perm[swap[0]]
-    layer['perm'] = perm
-    
-    output_shape = [input_shapes[0][p] for p in perm]
-
-    return layer, output_shape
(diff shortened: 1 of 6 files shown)


hls4ml/converters/keras/core.py
```diff
@@ -120,3 +120,19 @@ def parse_batchnorm_layer(keras_layer, input_names, input_shapes, data_reader, c
         layer['n_filt']=input_shapes[0][3]
 
     return layer, [shape for shape in input_shapes[0]]
+
+@keras_handler('Permute')
```
vloncar, line 124: The organization of handlers should follow Keras's convention, and by that `Permute` layer is not a "core" layer, rather a "reshaping" layer, so it should be placed in `reshape.py` module along with other similar handlers.

hls4ml/converters/keras/core.py
```diff
@@ -120,3 +120,19 @@ def parse_batchnorm_layer(keras_layer, input_names, input_shapes, data_reader, c
...
+def parse_permute_layer(keras_layer, input_names, input_shapes, data_reader, config):
+    assert(keras_layer['class_name'] == 'Permute')
+
+    layer = parse_default_keras_layer(keras_layer, input_names)
+
+    layer['class_name'] = 'Transpose'
+    swap = keras_layer['config']['dims']
+    perm = list(range(len(input_shapes[0])))
+    perm[swap[0]], perm[swap[1]] = perm[swap[1]], perm[swap[0]]
```
vloncar, line 133: Hmm... in 3d this breaks. `perm` should just contain values of `dims` with a 0-based index. Aka `perm = [dim - 1 for dim in keras_layer['config']['dims']]`

hls4ml/model/hls_layers.py
```diff
@@ -1513,14 +1513,18 @@ def initialize(self):
         self.set_attr('dim', '{}d'.format(len(inp.shape)))
         if len(perm) == 4 and perm[0] == 0:
             perm = [i - 1 for i in perm[1:]]
+        elif len(perm) == 3 and len(inp.shape) == 2 and perm[0] == 0:
```
vloncar, line 1516: This check shouldn't be needed. Even the previous one for 3d looks like a bad idea. We expect `perm` to be zero-based, not to do extra work after conversion.

hls4ml/model/hls_layers.py
```diff
@@ -1529,7 +1533,7 @@ def initialize(self):
 
     def function_cpp(self):
         params = self._default_function_params()
-        params['dim'] = self.get_attr('dim')
+        params['dim'] = '3d' # always use 3d instead of self.get_attr('dim')
```
vloncar, line 1536: I think we should reevaluate this decision and consider making a 2d transpose as well. It would be simpler and compile-time constant, therefore better for Vivado than the general 3d one that is difficult for the compiler to optimize.

hls4ml/model/hls_layers.py
```diff
@@ -1842,6 +1846,7 @@ def _get_transforms_config(self, params):
     'Resize'                 : Resize,
     'UpSampling2D'           : Resize,
     'Transpose'              : Transpose,
+    'Permute'                : Transpose,
```
vloncar, line 1849: You don't need this since you put `Transpose` as the `class_name` attribute in the handler function.

vloncar (changes requested):
Thanks Javier. Can you look into the comments I made?

hls4ml/converters/keras/core.py
jmduarte: I totally agree.. I copied the TF.Keras convention from here: https://github.com/tensorflow/tensorflow/blob/v2.5.0/tensorflow/python/keras/layers/core.py#L567-L621
which defines it in core, but I'm happy to move it

hls4ml/model/hls_layers.py
jmduarte: Agreed this can be removed.

hls4ml/model/hls_layers.py
jmduarte: Yes, I think it makes sense to make a new 2d specific transpose

hls4ml/converters/keras/core.py
jmduarte: yes, agreed.. i think it can be 

```python
layer['class_name'] = 'Transpose'
perm = [dim - 1 for dim in keras_layer['config']['dims']]
layer['perm'] = perm
output_shape = [input_shapes[0][0]] + [input_shapes[0][p+1] for p in perm]
```

assuming `input_shapes[0][0]` is `None`

hls4ml/model/hls_layers.py
jmduarte: Yes, I think this logic can be removed

jmduarte:
Thanks for the updates @vloncar, the simple test now works for both 2D and 3D (https://gist.github.com/jmduarte/d0b0036fffef32820ec2112e095bb7b7)

vloncar (approved):

PR #361, opened 2021-07-19
Fix Concatenate on different axes

jmduarte:
- Keras Concatenate layer assumes `axis=0` is the batch dimension, so if we are parsing that directly, we need to consider that, i.e.
  - for 2D concatenation, `axis = 2` or `axis = -1`, means we use `concatenate2d_1`; otherwise we use `concatenate2d_0`
  - for 3D concatenation, `axis = 3` or `axis = -1`, means we use `concatenate3d_2`; `axis = 2` or `axis = -2` means we use `concatenate3d_1`; otherwise we use `concatenate3d_0`
- Also fix shape inference

### DIFF
--- a/hls4ml/converters/keras/merge.py
+++ b/hls4ml/converters/keras/merge.py
@@ -10,13 +10,14 @@ def parse_merge_layer(keras_layer, input_names, input_shapes, data_reader, confi
 
     layer['op'] = layer['class_name'].lower()
 
+    output_shape = input_shapes[0][:]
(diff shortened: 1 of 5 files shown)


vloncar (approved):

jmduarte:
@vloncar do you know why these tests are failing?

vloncar:
Last time it failed sure to the machine running out of memory (too many people used it at the same time). It appears that after that I restarted only the failing step in the pipeline instead of the whole thing :face_palm:

PR #328, opened 2021-04-29
Fixes for Binary & Ternary NNs with QKeras

thesps:
Our conversion flow wasn't really working for QKeras Binary & Ternary NNs before (it did at some point, not sure when it broke...). This PR restores that flow. It also brings support for the `threshold` parameter to `ternary` quantizer for `QActivation`, which we didn't have before. 

[Here's some code](https://gist.github.com/thesps/afab2760be617c9d66ad23bca3d34d3a) testing several example models, combining binary/ternary weight quantizers, binary/ternary activations (specified with functions, classes, and strings), alpha (weight scale factor), and BatchNormalization.

The models are created with random weights, then converted, and the output is compared vs QKeras evaluation. 

On master branch the output is like:
```
0
...
Discrepancies 0 / 10000
1
...
Couldn't convert, compile, and predict model
2
...
Couldn't convert, compile, and predict model
3
...
Couldn't convert, compile, and predict model
4
...
Couldn't convert, compile, and predict model
5
...
Couldn't convert, compile, and predict model
6
...
Couldn't convert, compile, and predict model
```

With the PR the output is like:
```
0
...
Discrepancies 0 / 10000
1
...
Discrepancies 4 / 10000
2
...
Discrepancies 1 / 10000
3
...
Discrepancies 40 / 10000
4
...
Discrepancies 7 / 10000
5
...
Discrepancies 2 / 10000
6
...
Discrepancies 1 / 10000

PR #362, opened 2021-07-19
New testing

thesps:
This PR adds `pytest` tests and Gitlab CI file to run those tests on the CERN Gitlab infrastructure. This is intended to run in addition to the current Jenkins set up for now. 

The tests currently included in this new set extend our current testing to execute inference of models and require "good enough" performance at the model output vs some Python reference. There are some completely synthetic models running on random data with a requirement on the relative and absolute difference at the model output, and some real models (using updated `example-models`) with a requirement on the model accuracy. One set of tests also checks attributes of the original Keras model vs the HLSModel to check that the conversion is correct directly.

Several of these tests currently fail! The intention is to fix the issues causing these tests to fail separately from this PR, but as a matter of priority. Some of them are actually already fixed in existing PRs.

Also worth noting: the tests currently implemented are not the final set, we should continue to add more.


thesps:
Reopened the PR from the branch on the organisation repo rather than my fork. Usually the Gitlab pulls PRs to run the test, but I think this messed up here because I had the same branch on the org as well as my fork.

jmduarte (changes requested):
General comments: 
- Can we open up the gitlab so users who make PRs can look at the passing/failing tests, i.e. I can't see: https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/pipelines/2839318

.gitlab-ci.yml
```diff
@@ -0,0 +1,19 @@
...
+    - pip install .[profiling]
+  script:
+    - cd test/pytest
+    - pytest -rA --cov-report xml --cov=hls4ml --junitxml=report.xml
+  artifacts:
+    when: always
+    reports:
+      junit: 
+        - test/pytest/report.xml
+        - test/pytest/coverage.xml
```
jmduarte, line 19: do we know how much of the code is covered by the tests?

jmduarte:
Interestingly, when I run the tests locally on this branch, I only get 12 failing tests, and merging #328 fixes 6 of them:
```
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer0-bias_quantizer0-binary_tanh-False-False]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer1-bias_quantizer1-binary_tanh-False-True]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer2-bias_quantizer2-activation_quantizer2-True-True]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer3-bias_quantizer3-ternary-True-False]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer4-bias_quantizer4-activation_quantizer4-True-False]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer5-bias_quantizer5-activation_quantizer5-True-False]
PASSED test_qkeras.py::test_btnn[10-kernel_quantizer6-bias_quantizer6-activation_quantizer6-False-True]
```

so the remaining 6 failures are:
```
FAILED test_cnn_mnist.py::test_accuracy[resource-io_parallel] - assert (0.9885 > 0.95 and 0.8867981790591806 < 0.01)
FAILED test_keras_api.py::test_activation_thresholdedrelu[ThresholdedReLU] - AssertionError:
FAILED test_qkeras.py::test_single_dense_activation_exact[8] - AssertionError:
FAILED test_qkeras.py::test_quantizer[quantizer5] - AssertionError:
FAILED test_qkeras.py::test_quantizer[quantizer6] - AssertionError:
FAILED test_softmax.py::test_softmax[latency-flat_distribution] - assert 0.887 >= 0.98
```

thesps:
For the binary/ternary tests, those are basically the test cases I included in that PR #328. That PR is a bit needlessly tangled up with the graph manipulation PR #367 since these test models are all simple shallow NNs requiring the final node to be replaced. So I've commented out that test with a note to include it with that PR. If we merge this one, I can add those tests as part of that binary/ternary PR.

Of these other 6 failing I think it's okay to remove them for now with the aim to adding them back later.

`FAILED test_cnn_mnist.py::test_accuracy[resource-io_parallel] - assert (0.9885 > 0.95 and 0.8867981790591806 < 0.01)`
This must be some bug in hls4ml, probably we're not doing a transpose properly for the weights in this case. We can open an issue for this.

`FAILED test_keras_api.py::test_activation_thresholdedrelu[ThresholdedReLU] - AssertionError:`
This one I'm not actually sure about, I'm commenting it out for now but again with a note to fix the issue and include the test. 

```
FAILED test_qkeras.py::test_quantizer[quantizer5] - AssertionError:
FAILED test_qkeras.py::test_quantizer[quantizer6] - AssertionError:
```
For these two I actually know the bug, I think it's something we discussed that the data type we set in the config util for `quantized_relu` is not quite right. I think the fix is just to add 1 bit to the width for `quantized_relu` in that utility. But again, I'll comment out those two tests with a note to include them with the fix that can follow up in a separate PR.

```
FAILED test_qkeras.py::test_single_dense_activation_exact[8] - AssertionError:
FAILED test_softmax.py::test_softmax[latency-flat_distribution] - assert 0.887 >= 0.98
```
These other two are things that might not be "bugs", but hopefully tests we can make pass in future. For now I comment them out as well

test/pytest/test_keras_api.py
```diff
@@ -0,0 +1,483 @@
...
+    assert list(hls_model.get_layers())[2].attributes['class_name'] == model.layers[1]._name
+    assert list(hls_model.get_layers())[0].attributes['input_shape'] == list(model.layers[0].input_shape[1:])
+    assert list(hls_model.get_layers())[1].attributes['n_in'] == model.layers[0].input_shape[1:][0]
+    assert list(hls_model.get_layers())[1].attributes['n_out'] == model.layers[0].output_shape[1:][0]
+    assert list(hls_model.get_layers())[2].attributes['activation'] == str(model.layers[1].activation).split()[1]
+    assert list(hls_model.get_layers())[1].attributes['activation'] == str(model.layers[0].activation).split()[1]
+
+
+# # DONE
+# def test_reshape():
```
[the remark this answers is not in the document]
thesps: I inherited this tests from a previous student, and didn't clean up. This reshape one doesn't seem to work, so I've remove that completely.

test/pytest/test_keras_api.py
```diff
@@ -0,0 +1,483 @@
...
+
+
+# # DONE
+# keras_conv2d = [Conv2D]
+# padds_options = ['same', 'valid']
+# chans_options = ['channels_first', 'channels_last']
+# @pytest.mark.parametrize("conv2d", keras_conv2d)
+# @pytest.mark.parametrize("chans", chans_options)
+# @pytest.mark.parametrize("padds", padds_options)
+# def test_conv2d(conv2d, chans, padds):
```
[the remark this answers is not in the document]
thesps: I've checked, and apart from the tests with `channels_first`, this works and seems useful. So I've uncommented all that and it's now included.

I did a bit more tidy up of this file as well, since the three different activation test functions could be combined.

vloncar (approved):

.gitlab-ci.yml
```diff
@@ -0,0 +1,19 @@
+pytest:
+  image: gitlab-registry.cern.ch/fastmachinelearning/hls4ml-testing:0.1.base
```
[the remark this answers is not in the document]
jmduarte: OK I understood that the Dockerfile is to be kept private for now due to licensing

.gitlab-ci.yml
jmduarte: Not critical, but it would be nice to add coverage badge to the README

test/pytest/test_keras_api.py
```diff
@@ -0,0 +1,355 @@
...
+    assert list(hls_model.get_layers())[1].attributes['n_out'] == model.layers[0].output_shape[1:][0]
+    assert list(hls_model.get_layers())[2].attributes['activation'] == str(model.layers[1].activation).split()[1]
+    assert list(hls_model.get_layers())[1].attributes['activation'] == str(model.layers[0].activation).split()[1]
+
+@pytest.mark.parametrize("activation_function", [Activation(activation='relu', name='Activation'),
+                                                 LeakyReLU(alpha=1.0),
+                                                 ELU(alpha=1.0),
+                                                 PReLU(alpha_initializer="zeros",),])
+                                                 # TODO: add ThresholdedReLU test when it can be made to pass
+                                                 #ThresholdedReLU(theta=1.0)])
```
jmduarte, line 60: add reference to #376

test/pytest/test_qkeras.py
```diff
@@ -0,0 +1,220 @@
...
+  return randX(1000, 1)
+
+@pytest.mark.parametrize('quantizer', [(quantized_bits(8,0)),
+                                       (quantized_bits(8,4)),
+                                       (quantized_bits(4,2)),
+                                       (quantized_bits(4,0)),
+                                       (quantized_bits(10,0)),])
+                                       # TODO include these after fixes
+                                       #(quantized_relu(4)),
+                                       #(quantized_relu(10))])
```
jmduarte, line 191: Add reference to #377

jmduarte (changes requested):
I ran it locally and everything looks good. I have some minor comments, but I think it's *basically* ready to be merged

thesps:
Latest changes: I added all the references to the relevant Github issues, thanks for opening those!

Regarding the comment:
> I wonder if we should change the output directory name to include io_type and strategy so we can save all the artifacts for further inspection in case of failing tests

I decided to do this for all the tests basically, so every test that creates a new HLSModel now sets the output directory to a name like `hls4mlprj_<module name>_<test name>_<parameters...>`. All of the generated `.tar.gz` project archives are uploaded as job artifacts (see for example the pipeline of the most recent commit [here](https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/15712656/artifacts/browse/test/pytest/)). At the moment none of these tests run synthesis - we don't have the runner set up yet - but when we do that we'll probably want to upload more than just the `.tar.gz`, like the logs and reports.

Side note: by making every test write its projects to a different directory, I noticed the `test_keras_api.test_pooling` tests didn't produce any projects. It needed an overhaul, but now it performs some useful tests.

thesps:
In terms of code coverage, I found out how to capture that properly on Gitlab now. After the pipeline runs on Gitlab the percentage is printed in the summary on the right. The detailed breakdown is visible in the log now, as well as the xml that is uploaded. I think the xml can be used to annotate the code with coverage info, e.g. for a PR you can see whether the changes are covered, but I didn't yet find out how to do that. There should now be badges too, although they don't seem to want to display in the markdown. Left one is for master branch (no report yet), right for the branch of this PR.

[Coverage](https://gitlab.cern.ch/fastmachinelearning/hls4ml/badges/master/coverage.svg) [Coverage](https://gitlab.cern.ch/fastmachinelearning/hls4ml/badges/gitlab-pytest/coverage.svg)

jmduarte (approved):
Looks good!

PR #363, opened 2021-07-19
Explicitly set Transpose output precision

jmduarte:
- Explicitly set Transpose output precision as the same as input precision


vloncar:
@jmduarte Can you elaborate why is this needed? Setting it like this will prevent us from explicitly overriding it in config. Also, if we do this, the same should be done to padding layers.

thesps:
This seems to make sense to me - a transpose shouldn't be changing the data type so we can effectively prevent it from being configured like this.

jmduarte:
Sorry, I should've elaborated before!

Basically, I find when the overall model precision is set and it's different from the layer precision before a `Permute` layer, the current behavior is to take the model precision as the output precision of `Permute` layer.

But this fails because the input precision is then different from the output precision of the `Permute` layer. It's also nonintuitive.

Here's a way to reproduce the issue on `fastmachinelearning:master`, but this works on `jmduarte:permute`: https://gist.github.com/jmduarte/d0b0036fffef32820ec2112e095bb7b7

So the solutions I imagine are
- Don't let users mix Model-level precision specification and Layer-lavel precision specification
- Allow `Permute` layer to output different precision than input by adding a cast in HLS (seems counterintuitive)
- Force `Permute` layer to output the same precision (this PR)

If we agree with the approach of this PR, I'm happy to apply it to the other layers (like ZeroPadding, etc.)

vloncar:
I got burned by this many times, so I agree with the proposal. I would also add a check in `config.py` to make sure these layers are skipped when creating an initial config to not confuse the user that something can be changed. How do we conceptually define which layers have to have matching input and output precision? For example, only layers that change the shape of tensor without applying any "operation" on them? Does `UpSampling2D` fits this?

thesps:
I agree with the principle of this PR, and Javier's option:

- Force Permute layer to output the same precision (this PR)

Regarding what this should be applied to:

> only layers that change the shape of tensor without applying any "operation" on them

I think that covers most of the cases. Another example I can think of is the linear activation layer inserted after a Dense/Conv. That can be really nasty in patterns where you have `Dense -> (Linear) Activation -> (Real) Activation` since there can be an  accidental cast in the middle. Another one that I think came up previously could be Pooling. There it _might_ make sense to allow user configuration, but the default should be for the output type to match the input.

vloncar:
Regarding linear activation I would propose we always remove it, and the "cast" use case should insert a different layer (that can call `nnet::linear` underneath)

jmduarte:
Updated the PR at adfcf58  to
- explicitly set output precision of `Resize`, `ZeroPadding1D`, `ZeroPadding2D`, and `Permute` HLSLayers to be the same as the input precision
- add `Reshape` and `Permute` Keras layers to skipped layers in `config.py` (should `ZeroPadding` or `UpSampling` layers be here?)

@thesps @vloncar let me know if it looks good now or if we want additional things in this PR (like skipping linear activations?)

thesps:
Code wise it looks good to me, I will give a test later. 

> or if we want additional things in this PR (like skipping linear activations?)

I think we can save that for another PR

vloncar:
@thesps Did you have time to test this?

PR #364, opened 2021-07-20
Support pointwise conv1d for io_parallel

vloncar:
This relies on #354 which is required to compile the generated C++

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
@@ -118,5 +118,78 @@ void conv_1d_latency_cl(
     }
 }
 
+template<class data_T, class res_T, typename CONFIG_T>
+void pointwise_conv_1d_cl(
+    data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
+    res_T  res[CONFIG_T::out_width * CONFIG_T::n_filt],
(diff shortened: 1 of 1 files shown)


jmduarte:
Looks good to me!

Test code works: https://gist.github.com/jmduarte/aee1d689118727cdfcb1cdabfdab6b7f

Should we also make a similar function for 1x1 Conv2Ds? Or is the idea is that would never work for any reasonably sized Conv2D anyway?

Also, this is probably out of scope for this PR, but I'm wondering if we can improve this implementation for reuse_factor > 1 by changing the way we partition the arrays, etc.

jmduarte (approved):
Looks good to me!

Test code works: https://gist.github.com/jmduarte/aee1d689118727cdfcb1cdabfdab6b7f

Should we also make a similar function for 1x1 Conv2Ds? Or is the idea is that would never work for any reasonably sized Conv2D anyway?

Also, this is probably out of scope for this PR, but I'm wondering if we can improve this implementation for reuse_factor > 1 by changing the way we partition the arrays, etc. We may have people that can look into this since it's related to DeepMET at Level-1 (https://www.dropbox.com/s/3qvdr2kqq9h1n4y/DeepMET_hls4ml_21July2021.pdf?dl=0)

vloncar:
I made a Conv2D as well. It will come as part of the next step, where we use this instead of Dense for multidimensional tensors (as we discussed). With the support in place, we can explore better HLS implementations.

PR #366, opened 2021-07-21
Support doing Dense over n-dimensional (n>1) tensor with 1x1 Conv

vloncar:
As was discussed before, we currently don't support `Dense` layer being applied to `n`-dimensional tensor (with `n>1`, not counting the batch dimension). This operation is equivalent to doing 1x1 convolution over the same tensor, which we support as a special case. This PR does that by adding an optimizer that replaces `Dense` with `PointwiseConv1D/2D`. To support this, converter also had to be extended to properly handle the output shape as well as the `Dense` layer itself in the hls4ml IR.

Test case can be found [here](https://gist.github.com/vloncar/2b7bb326bb4b8878e6bb251612c8e670)

### DIFF
--- a/hls4ml/model/optimizer/passes/multi_dense.py
+++ b/hls4ml/model/optimizer/passes/multi_dense.py
@@ -1,0 +1,52 @@
+from hls4ml.model.optimizer import OptimizerPass
+from hls4ml.model.hls_model import Dense
+
+
+class ReplaceMultidimensionalDenseWithConv(OptimizerPass):
+    def match(self, node):
+        return node.__class__.__name__ == 'Dense' and \
+            len(node.get_input_variable().shape) > 1
+
+    def transform(self, model, node):
+        dim = len(node.get_input_variable().shape) - 1        
+        input_shape = node.get_input_variable().shape
+
+        pointwise_attrs = {
+            'data_format': 'channels_last',
+            'padding': 'valid',
+            'n_chan': input_shape[-1],
+            'n_filt': node.get_attr('n_out'),
+        }
+
+        if dim == 1:
+            pointwise_attrs.update({
+                'in_width': input_shape[0],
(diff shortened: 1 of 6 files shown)


vloncar:
I should mention that the HLS implementation for 1x1 convolution for `io_parallel` is based on the `latency` implementations of Conv1D/2D. By no means are these the best possible implementations and we should explore better ways of doing this (like in the `dense_resource` function). But after this PR the changes to HLS will be localized to a single function in the`.h` files without touching the rest of the conversion flow.

jmduarte (approved):
Changes look good to me, and I tested it with the DeepMET and IN models I've been playing with

jmduarte:
Actually, I have a small fix related to #368: https://github.com/vloncar/hls4ml/pull/42

but perhaps we can discuss if this is the best way to do it.

hls4ml/model/optimizer/passes/multi_dense.py
```diff
@@ -0,0 +1,52 @@
...
+                'pad_top': 0,
+                'pad_bottom': 0,
+                'pad_left': 0,
+                'pad_right': 0,
+            })
+        else:
+            raise Exception('Cannot replace Dense over {dim}D tensor with Conv{dim}D.'.format(dim=dim))
+
+        class_name = 'PointwiseConv' + str(dim) + 'D'
+        pw_node = model.make_node(class_name, node.name, pointwise_attrs, node.inputs.copy())
```
jmduarte, line 49: We somehow have to make sure the `pw_node` picks up the weights/biases from `node` (I'm not sure why it's not happening by default)

jmduarte (changes requested):

vloncar:
Do we need anything else before we merge this?

jmduarte (approved):

PR #367, opened 2021-07-22
Update graph insert/remove/replace

thesps:
- In `remove_node`, when the node being removed is an output node, the model outputs need to be reconnected to the input of the removed node.
- In `remove_node`, we raise an exception that we cannot handle nodes with multiple outputs (nor inputs). However, even for a node with a single output, that output may be consumed my multiple other nodes. So I've added a loop over nodes connected to that output, reconnecting them to the removed node's input whereas previously only one node would have been reconnected
- Rather than `replace_node` duplicating that logic, I've modified it to use `insert_node` then `remove_node` directly.

### DIFF
--- a/hls4ml/model/hls_model.py
+++ b/hls4ml/model/hls_model.py
@@ -427,7 +427,12 @@ class HLSModel(object):
             if len(node.inputs) > 1 or len(node.outputs) > 1:
                 raise Exception('Cannot rewire a node with multiple inputs/outputs')
             prev_node = self.graph.get(node.inputs[0])
-            next_node = next((x for x in self.graph.values() if node.outputs[0] in x.inputs), None)
+            next_nodes = [x for x in self.graph.values() if node.outputs[0] in x.inputs]
+            if len(next_nodes) == 0:
+                if node.outputs[0] in self.outputs:
+                    self.outputs = [node.inputs[0] if x == node.outputs[0] else x for x in self.outputs] 
+                else:
+                    raise Exception('Cannot rewire a node which has no child nor is an output node')
             if prev_node is not None:
                 if next_node is not None:
                     for i,_ in enumerate(next_node.inputs):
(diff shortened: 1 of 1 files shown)


thesps:
I've been testing this PR with various models that will cause optimizer passes to insert, remove, & replace nodes. Including those from PR #366 and PR #328

I've now started creating some synthetic models to insert, remove & replace nodes in arbitrary places, [here](https://gist.github.com/thesps/aa5d7907aa8902727e3506e73a1b5a5e). It's currently failing, I think I know why but I've returned the PR to [WIP] while testing this way a bit more.

thesps:
I've expanded the notebook test cases linked previously in the conversation into a new `test_graph.py` for the pytest CI. Here I create a few models and insert, remove & replace nodes from them systematically. The models are simple but attempt to capture the various different ways in which these graph manipulations can trip up.

The models before any manipulations are these:

### Basic:
model

### Branched:
model

### 'Skip' / branched with differing length branches
This is tested in a 'left' and 'right' version (ie which input to the `concat` is the 'skip')
model

All the Dense layers are 1 input & 1 output, doing `y = 2x + 1` so the output can be easily verified (which is what the tests do). The tests also check that the list of layer names of the manipulated `HLSModel` match a naive insertion/deletion of the layer name into the list of layer names. Hopefully these two things together capture the various failure modes.

The test includes all the combinations I could think of that touch the layer connected to the model input & output as well as layers in the middle. For the branched models, we can now control which branch an inserted layer should be connected on (only the branch leading to the `before` node if that argument of `insert_node` specified, or both if left to `None`).

One thing that the tests don't cover is a branched model with multiple model outputs, since it's not possible yet with `HLSModel.predict`, and the alternative method with Vivado isn't possible yet in the test environment. I've checked the models with this feature from #328 and it's working. The models in the tests use a `Concatenate` to merge the branches.

I think this is now ready for review & merge, but I'm happy for more feedback, and any more suggested types of model topologies or insert/remove/replace patterns are welcome/

PR #371, opened 2021-08-06
Fix graph for multiple model inputs

jmduarte:
Purpose: Fix graph for multiple model inputs. 

- Issue: Currently, the second model input assumes the first model input is also an input to it. This PR fixes that behavior.
- This bug has a practical effect in that it will result in unnecessarily clone the first model input (because it sees it is "input" to two different layers) and cause an issue synthesizing the resulting HLS.
- Incorrect graph before the fix:
model_hls4ml
- Correct graph after the fix:
model_hls4ml
- Test showing this behavior (fails on `fastmachinelearning:master`, succeeds on `jmduarte:multiple_inputs_fixes`): https://gist.github.com/jmduarte/d9146904980b6574032e90621ef383fb
- There's also a related issue of plotting graphs with layers with multiple outputs that I will try to fix in this PR

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -385,7 +385,11 @@ class Layer(object):
 
     def get_input_node(self, input_name=None):
         if input_name is not None:
-            return self.model.graph.get(input_name)
+            nodes = [node for node in self.model.graph.values() if input_name in node.outputs]
+            if len(nodes) == 0:
+                return None
+            else:
+                return nodes[0]
         else:
             return self.model.graph.get(self.inputs[0])
 
--- a/hls4ml/model/hls_model.py
+++ b/hls4ml/model/hls_model.py
@@ -323,7 +323,9 @@ class HLSModel(object):
             name = layer['name']
             inputs = layer.get('inputs', [])
             outputs = layer.get('outputs', [])
-            if len(inputs) == 0:
+            if kind in ['InputLayer', 'Input']:
+                inputs = ['input']
+            elif len(inputs) == 0:
                 inputs = [next(reversed(self.graph), 'input')]
             if len(outputs) == 0:
                 outputs = [name]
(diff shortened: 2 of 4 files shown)


jmduarte:
I made a slightly more complex test case, and I think it works there now:
https://gist.github.com/jmduarte/d9146904980b6574032e90621ef383fb

model_hls4ml

vloncar:
What is the relation of this and #367?

jmduarte:
@thesps should comment as well, but I think #367 is trying to generalize/fix the logic of `remove_node` and `insert_node` (and use them in `replace_node`).

Whereas this PR is fixing few specific bugs:
1. Ensuring that true input nodes don't have any inputs except for "input". This was broken previously because `input_2` treated `input_1 `as an input just because it came earlier. See https://github.com/fastmachinelearning/hls4ml/pull/371/files#diff-fb08e31764e30a21598f9f6b2b8a08e53ba4d222153748b639ee4c7e49044156L327-R329
2. Fix to broadcast layer to prevent it from being inserted multiple times / with an incorrect name (See https://github.com/fastmachinelearning/hls4ml/pull/371/files#diff-db0cc483225975e21a2e729dccbf4b7cc69a613b753f38d316232e00f414d3ceL116-R120)
3.  Fix to `insert_node` to work when inserting a layer like `broadcast` (which has multiple outputs)  (See https://github.com/fastmachinelearning/hls4ml/pull/371/files#diff-596f42295045305371ca3a443fce3a9c78302f2462ded2f2b16e2c3791944e95L388-R392)


Note that 3. touches `insert_node`, but seems to be an independent change to #367 (i.e. doesn't introduce any conflicts).
I agree it would be good to test both together.

jmduarte:
@thesps I added a test case as requested (I think). Could you please check again?

thesps (approved):

PR #373, opened 2021-08-06
Remove old install scripts

vloncar:
The install scripts are not really useful and install ancient packages that don't work with modern hls4ml. This just confuses the users (I heard several complaints already). hls4ml should be installed with pip or conda, as the docs say.


jmduarte (approved):
looks good!

PR #378, opened 2021-08-13
Propagate correct shape for Flatten layer

vloncar:
The recent addition of supporting doing `Dense` over multi-dimensional tensors (#366) introduced an issue with the output shape of `Flatten` layer being ignored. Since `Flatten` is skipped, the optimizer would check for the preceding layer's output which can be the layer before `Flatten` and replace `Dense` with 1x1 `Conv`.

This is fixed by properly handling `Flatten` as a `Reshape`. This allows the optimzer to properly handle the shapes and `Reshape` is just a placeholder that has no impact on HLS.

[Test case](https://gist.github.com/vloncar/ff5864d9a1b7eee2665bc4f6b0d127ad).

Side note: this leaves `Dropout` as the only skipped layer. We might as well parse it and all other regularization layers properly into `HLSModel` and remove them in the `nop` optimizer where we have better ways of handling graph transformations.

### DIFF
--- a/hls4ml/converters/keras/reshape.py
+++ b/hls4ml/converters/keras/reshape.py
@@ -4,6 +4,18 @@ from hls4ml.converters.keras_to_hls import parse_default_keras_layer
 from hls4ml.converters.keras_to_hls import keras_handler
 from hls4ml.converters.utils import parse_data_format
 
+@keras_handler('Flatten')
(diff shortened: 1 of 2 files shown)


jmduarte (approved):
Looks good to me! I also needed this for TinyML Conv2D models and verify it works as intended!

Issue #381, opened 2021-08-19
Tests fail or pass depending on random seed


jmduarte:
Certain tests, e.g. `test_single_dense_activation_exact[4,6]` fail or pass depending on the random seed.

One general good practice is to use the [pytest-randomly plugin](https://github.com/pytest-dev/pytest-randomly) to set a random seed (and also repeat tests for different random seeds).

Also, we should try to understand the different failure modes for this particular test.

jmduarte:
Closed with merger of #382

PR #382, opened 2021-08-19
Set random seed for reproducible tests

jmduarte:
Address #381
- Set random seed with `pytest-randomly` for reproducible tests
- Remove 6-bit `test_single_dense_activation_exact` for now (fails much more often than 4-bit version)
- Requires this merge request (https://gitlab.cern.ch/fastmachinelearning/hls4ml-testing/-/merge_requests/1)
adding `pytest-randomly` to the Docker image for testing.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ pytest:
     - pip install .[profiling]
   script:
     - cd test/pytest
-    - pytest -rA --cov-report xml --cov-report term --cov=hls4ml --junitxml=report.xml
+    - pytest -rA --cov-report xml --cov-report term --cov=hls4ml --junitxml=report.xml --randomly-seed=42
   artifacts:
     when: always
     reports:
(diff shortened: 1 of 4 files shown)


thesps:
I'm finding the same behaviour as the CI testing when trying this offline, that I don't yet understand. With `pytest ... --randomly-dont-reorganize` the same set of 4 tests from `test_keras_api.py` fail each time (all the `test_activations` ones). These run in the same order but don't fail when run without `pytest-randomly` at all using `pytest -p no:randomly`. 

Then there is some kind of crash during the `test_qkeras.py` with an `Aborted (core dumped)` at the end.

Possibly some tests leave the environment in a bad state, though I don't have a good idea where that could be.

jmduarte:
@thesps so adding `--randomly-dont-reset-seed` (which prevents the plugin from resetting the random seed at the beginning of each test) makes all the tests pass again and avoids the segfault. I think this can be merged now and in the future, we can figure out how to re-enable these two (randomizing the order and randomizing the random seed at the beginning of each test).

thesps (approved):

PR #384, opened 2021-08-20
Support applying Softmax over multidimensional tensors

vloncar:
Current implementations of `softmax` are incorrect when applied to multidimensional tensors (used in e.g., ENet model). The implementations in this PR handle this properly for `axis=-1` (the default).

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -1384,6 +1384,9 @@ class Softmax(Activation):
                 self.set_attr('implementation', 'latency')
             else:
                 self.set_attr('implementation', self.model.config.get_strategy(self).lower())
+            
+            if self.model.config.get_config_value('IOType') == 'io_parallel':
+                assert len(self.get_input_variable().shape) == 1, 'Softmax with io_parallel strategy cannot be used on multidimensional tensors.'
 
 class TernaryTanh(Activation):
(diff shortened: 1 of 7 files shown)


vloncar:
I forgot the [test case](https://gist.github.com/vloncar/df2b2e2b8dfe483ca46967b6c1d786c6)

hls4ml/model/hls_layers.py
```diff
@@ -1384,6 +1384,9 @@ def initialize(self):
                 self.set_attr('implementation', 'latency')
             else:
                 self.set_attr('implementation', self.model.config.get_strategy(self).lower())
+            
+            if self.model.config.get_config_value('IOType') == 'io_parallel':
+                assert len(self.get_input_variable().shape) == 1, 'Softmax with io_parallel strategy cannot be used on multidimensional tensors.'
```
thesps, line 1389: Are you happy with this here, or want to move it to the backend for easier integration with the backend PR?

vloncar: The limits of the `io_stream` implementation (the `axis` restriction) is made as an assert in C++/HLS, I was thinking the same should be done here, but there is no way to know the shape info at that stage in the C++. For the integration with backend PR this line is simply placed in the backend initializer for the layer, it is not difficult to follow up.

PR #395, opened 2021-09-10
New backend development framework

vloncar:
This is the big PR that was months in the making. The general idea behind it was to decouple the internal model representation from the backend-specific requirements and to be able to create new backends that don't behave exactly the same nor even target the same type of processor (the FPGA).

The changes are extensive, and I'll try to briefly summarize them here:
- Layers now have only inputs, outputs and attributes. All information about the layer's state and configuration is stored in the attributes. All weights, variables and data types are attributes and there are mapping views to sort through them. Layers can define expected attributes and can be verified for correctness, or to produce a list of configurable attributes that user can tweak. Note that the list of expected attributes is not complete for every layer ATM (I was lazy).
- All model/layer transformations should happen in the optimizers. To support the many new uses of optimmizers, the framework was expanded to support many new types of optimizers (layer-specific, backend-specific, whole-model, templates, decorators etc).
- Since there is now a lot more optimizers now, a concept of "flows" is introduced. A flow is just an ordered set of optimizers that may depend on other flows. There are flows shared by each backend (related to proper conversion and common optimizations) and each backend provides its own flows, with one being the default. For example, Vivado backend defaults to an `ip` flow that applies all other flows and produces an IP (essentially the same functionality as we have now but broken into smaller chunks that can be controlled and hooked into)
- The internal model is initially backend-agnostic and doesn't define anything FPGA-specific. Backend-specific changes (including introductions of new layer attributes, such as the venerable `reuse_factor`) to the layers are applied via the flows/optimizers, which replace the layers, types and variables with their versions that can be converted to C++/HLS. For example, we start from simple objects representing tensors that are turned into stream- or array-based variables and the precision is converted from the starting data type to the specific implementation (`ap_` or `ac_`)
- The `function_cpp` and `config_cpp` have been removed from the layers and replaced with backend-specific optimizers (templates) that contain the same functionality. This is the single biggest change by affected line numbers and it came in the very last commit.
- In light of these changes, the code has seen some shuffling. `templates` submodule now only contains the big C++ templates, the accompanying python classes have been moved to `backends` submodule. The folders of this module contain the backend implementations and their optimizers. The three (and the 4th that will come in 2022) FPGA backends share the same ancestor `FPGABackend` that defines the common stuff.
- _(Not yet implemented)_ Since we're moving away from using the term `HLS` in the layers/types/variables I would propose we drop `hls_` from the file names as well, so `hls_model.py` becomes `graph.py ` and `HLSModel` becomes `ModelGraph`, `hls_layers.py` becomes just `layers.py`, `hls_attributes.py` becomes `attributes.py`, `hls_types.py` becomes `types.py` etc

(I'll expand this description over time and also will follow up with a dedicated document describing each feature and how to use it)

Known issues:
- Flows broke the `Optimizers` and `SkipOptimizers` feature of the config
- ~QKeras alpharelated stuff has not been wired correctly, so most of the QKeras models don't work.~ Fixed as of 
35bf771
- ~GarNet is broken since it wasn't updated to make use of new features~ Fixed as of ea7e1d0
- The new VivadoAccelerator backend has seen very little testing in the new branch since it was merged only this week so I would expect it to break, but shouldn't be too hard to fix
- Probably many, many more issues I forgot to mention. There are TODOs scattered around the code pointing to the fixes/enhancements that will be made soon


jmitrevs:
I will repeat my previous quartus development with this branch to test how it works.

thesps:
Nice! Since this is supposed to facilitate backend development, I thought I'd learn how it works by knocking up a quick-and-dirty (and nowhere near feature complete) "tensorflow backend". It's [here](https://gist.github.com/thesps/1ea9ce2f9d358eb80a17c0c66741c462), with a small demo, in case anybody wants to take inspiration to try something similar. It was indeed pretty simple to add a new working backend.

I've started collecting some comments that I'll expand as I get more experience with the new features (didn't really touch attributes yet):
  - import error on `calmjs` & `tabulate` from `quartus_backend.py`: add to install requires.
  - `output_rounding_saturation_mode` optimizer pass (& other QKeras passes) are not registered when QKeras isn't in the environment, but they are part of the [convert flow](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/optimizer/__init__.py#L14) so they are attempted, which then crashes out since the pass isn't found. Conversion fails for any model if QKeras isn't in the environment.

I think the `FPGABackend` name should be changed: the in-development `CatapultBackend` that should inherit from it will target more than FPGAs; and other possible backends that might be developed to target FPGAs, but with a different type of architecture to our current implementations, probably shouldn't inherit from that class. Possible alternatives: 'DataflowBackend', 'ReuseFactorBackend'?

> The internal model is initially backend-agnostic and doesn't define anything FPGA-specific

On this, there are still a few remnants of backend-specific stuff that we should aim to push to the backends.
The `HLSModel` has [attributes](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/hls_model.py#L319-L320) and [methods](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/hls_model.py#L521-L525) related to 'BRAM'. Can that stuff be handled with (Vivado) backend optimizer passes instead?

The HLSConfig is still very backend-specific. Given that the HLSModel class has very few references to the `config`, and only accesses the "project level" info (backend, output directory), could the HLSConfig class be moved somewhere closer to the FPGABackend? Or is the plan to overhaul config more thoroughly in future?

You wrote:

> For example, Vivado backend defaults to an ip flow that applies all other flows and produces an IP (essentially the same functionality as we have now but broken into smaller chunks that can be controlled and hooked into)

The way I understand it so far, the `vivado:ip` flow prepares the HLSModel to be written, compiled & synthesized. It doesn't actually 'produce an IP'. That's fine, that's the same "flow" (for the user) as what we have now. Is the intention though that what is now `write`, `compile`, `build` should be moved to passes as part of this IP flow? I think we should clarify at this point already what type of things should be done with optimizer passes vs other backend functions.

Kind of related:

> The new VivadoAccelerator backend has seen very little testing in the new branch since it was merged only this week so I would expect it to break, but shouldn't be too hard to fix

If we do move the various `write_hls` and related synthesis-tool interfacing methods to optimizer passes, the `VivadoAccelerator` backend might be better handled as a flow as part of Vivado backend.

jmitrevs:
There is a swap between round mode and saturation mode that is easy to fix. (I have it fixed in a PR I am working on.) Profiling also has some issues that I will look at.

jmitrevs:
Even though Quartus dense is "Resource" strategy only, I think we should allow a reuse factor of 1 for it. Currently Resource strategy forbids a reuse factor of 1 since then the Latency strategy is preferable in that case, but we don't have a Latency strategy version for Quartus.

jmitrevs:
By the way, the full `calmjs` is not needed, just the parser part (`pip install calmjs.parse`).

jmitrevs:
The parsing of the javascript reprt file after running quartus seems fragile. I ran into a key error exception when trying to get the II, while the html summary shows it. It may be worth reviewing how report is parsed. I can devote some time to that later this week or next week.

vloncar:
@jmitrevs I think the best approach here is to divide it up into smaller functions and then wrap each of them in try/except blocks so that the key errors at least give us more information on which part of report parsing failed. The key errors I encountered came from either broken Quartus installation (when the tool fails, but that is not caught by hls4ml), or from running synthesis/fpgasynth on different models in the same output directory and they were all a pain to figure out where they came from.

vloncar:
> * import error on `calmjs` & `tabulate` from `quartus_backend.py`: add to install requires.

Thanks, I'll add the two dependencies. I should note that IMO, we should handle this in a better way. Ideally, it should be possible to install hls4ml with only the desired backends (like `pip install hls4ml[quartus]`), so that the modularity of the codebase is reflected in the package, though I won't push for that in this PR.

> * `output_rounding_saturation_mode` optimizer pass (& other QKeras passes) are not registered when QKeras isn't in the environment, but they are part of the [convert flow](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/optimizer/__init__.py#L14) so they are attempted, which then crashes out since the pass isn't found. Conversion fails for any model if QKeras isn't in the environment.

There are many more issues with QKeras support in the PR :smile: For this one in particular, I think we should consider making QKeras a dependancy since many of the cool bits of hls4ml really require using qkeras as well. Also we may consider if it makes sense to be modular here as well and support installation of only the desired frontend ((Q)Keras/PyTorch/QONNX).

> I think the `FPGABackend` name should be changed: the in-development `CatapultBackend` that should inherit from it will target more than FPGAs; and other possible backends that might be developed to target FPGAs, but with a different type of architecture to our current implementations, probably shouldn't inherit from that class. Possible alternatives: 'DataflowBackend', 'ReuseFactorBackend'?

Agreed. Both of those alternative names sound fine, I'll leave that decision to the community.

> On this, there are still a few remnants of backend-specific stuff that we should aim to push to the backends. The `HLSModel` has [attributes](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/hls_model.py#L319-L320) and [methods](https://github.com/vloncar/hls4ml/blob/quartus_new/hls4ml/model/hls_model.py#L521-L525) related to 'BRAM'. Can that stuff be handled with (Vivado) backend optimizer passes instead?

I knew I forgot something after the big merge with the latest master. You're right, those were intended to be done with optimizer passes.

> The HLSConfig is still very backend-specific. Given that the HLSModel class has very few references to the `config`, and only accesses the "project level" info (backend, output directory), could the HLSConfig class be moved somewhere closer to the FPGABackend? Or is the plan to overhaul config more thoroughly in future?

The big overhaul.

> The way I understand it so far, the `vivado:ip` flow prepares the HLSModel to be written, compiled & synthesized. It doesn't actually 'produce an IP'. That's fine, that's the same "flow" (for the user) as what we have now. Is the intention though that what is now `write`, `compile`, `build` should be moved to passes as part of this IP flow? I think we should clarify at this point already what type of things should be done with optimizer passes vs other backend functions.

I named it `ip` since I didn't have a better name for it. The idea was that `write`, `compile` and `build` stay and be backend-specific, but we can move them all to passes (then it is like FINN). However I find that running the `apply_flow()` for all this is less intuitive, so I would still map the aforementioned functions to calls to `apply_flow`, just for convenience.

> If we do move the various `write_hls` and related synthesis-tool interfacing methods to optimizer passes, the `VivadoAccelerator` backend might be better handled as a flow as part of Vivado backend.

That's an interesting topic and that is the way I initially wanted for the two targets to exist, as part of a single backend, but during the accelerator development there were string arguments for having it as part of a separate backend. We can revisit it once we get the backend flexibility quirks sorted out (i.e., post merge).

jmitrevs:
`test_cnn_mnist.py` and `test_conv1d.py` both fail in `ip_parallel` mode due to a redefinition of a variable in the firmware, e.g.:
```c++
    layer6_t layer6_out[N_OUTPUTS_6*N_FILT_6];
    #pragma HLS ARRAY_PARTITION variable=layer6_out complete dim=0
    nnet::conv_1d_cl<layer5_t, layer6_t, config6>(layer5_out, layer6_out, w6, b6); // conv3_relu

    layer7_t layer7_out[N_OUTPUTS_6*N_FILT_6];
    #pragma HLS ARRAY_PARTITION variable=layer7_out complete dim=0
    nnet::relu<layer6_t, layer7_t, relu_config7>(layer6_out, layer7_out); // conv3_relu_relu

    layer7_t layer7_out[N_SIZE_1_8];
    #pragma HLS ARRAY_PARTITION variable=layer7_out complete dim=0
    layer9_t layer9_out[N_LAYER_9];
    #pragma HLS ARRAY_PARTITION variable=layer9_out complete dim=0
    nnet::dense<layer7_t, layer9_t, config9>(layer7_out, layer9_out, w9, b9); // fc1_relu
```

jmitrevs:
At least with onnx inputs with the quartus backend, unsupported layers (conv2d in this case) do not give graceful error messages but instead produce strange HLS with the layers omitted. This is very much a note to me for something for me to try to fix when I get a chance, though others can of course go ahead and fix it, too.

jmitrevs:
I was wondering if one change would be useful for things like QONNX parsing. Many Layers set precisions is the initialize phase, which makes it more complicated to change that in the optimizer stages. For example, the Dense initialize calls self.add_weights, but at that moment we do not know the quantization we should apply. The Quant node hasn't been optimized into the Dense node yet.

I think such calls should be removed from initialize and be part of another function call that can either be called after optimizations for all the nodes or be part of the optimization. Or are there alternate recommendations?

vloncar:
@jmitrevs Every tensor in hls4ml (output or weight) has a typa and that in turn has a precision specified. There is no option of not specifying the type or precision in this PR or in master branch. If we allowed that, more problems would arise, trust me, you don't want this :smile: This information is not immutable, it can be changed with optimizers (some optimizers do this). If no quantizer is specified, the weights will not be touched. If we have a quantizer by the time we do initialize, then the weights will be transformed, but even if a wrong quantizer is used, the original tensor is still available in `data_unquantized`.

jmitrevs:
My point wasn't to allow no type to be specified, but for the step creating the weights and such to be a separate step, not in the initialize, since for example in QONNX all of these will be rewrittten.

vloncar:
If this is a separate step, then we would have to take into account that you can create a model graph that is not fully initialized, thus the other optimizers have to consider that maybe some layer doesn't have all the expected information. Like I said, you don't want this. Once the `HLSModel` is created, it has to be complete and valid, the conversion-based optimizers may make it more general so the other optimizers can process it, but it should never be in the incomplete state because this will cause problems further down the line. Either way, I don't see how is this discussion related to this PR?

vloncar:
Most of the issues are now resolved, the pytests pass and hopefully Jenkins tests as well. From the remaining issues, the config-related one will be addressed before merging, and the overhaul of config itself will come after as a separate PR. I'd appreciate if people can devote some time to test this branch.

jmitrevs:
Actually looking more closely I think this PR fixes the things that were bothering me the most previously with `add_weights_variable` and `add_output_variable`. Those are kept more local. So I am satisfied.

thesps:
Can you push this also to branch `fastmachinelearning:pr/395` to trigger pytests in CI?

vloncar:
It took just ~120 commits, but this PR now passes all tests. There's still many tests to be added, especially for Quartus and all the new features of the IR.

hls4ml/templates/quartus/firmware/weights/b2.h
```diff
@@ -0,0 +1 @@
+hls_init_on_powerup static const bias_default_t b2[config2::n_out] = {0.016205917468717468,0.8594570113367865,0.8702152750686849,0.17781786564910607,0.892286699539321,-0.481921867493851,0.3944938477209121,0.8171092371058477,-0.8344882848947686,-0.38468736037425466,-0.5344749363317189,-0.059873800340414896,-0.6397887543009495,-0.14078283158879978,-0.7373485450427875,0.6293532137862843};
```
thesps, line 1: What's this? Some left over testing files? (Same for b2_test.h, w2.h, w2_test.h)

thesps (changes requested):
Can we get a pytest that targets the Quartus backend? Perhaps adding a parametrisation by backend to test_keras_api.py which tests conversion but also a basic test of the backend correctness.

hls4ml/templates/quartus/firmware/weights/b2.h
vloncar: Removed

vloncar:
I have added the tests for the layers supported by Quartus backend into the keras_api test.

The latest commit includes the agreed change to drop the use of "HLS" in the names of the classes. I left `HLSConfig` for now since the next PR will be a big overhaul of the config system that will address the remaining config-related issues.

This is now ready.

thesps (approved):

Issue #398, opened 2021-09-21
GlobalPooling1D broken


jmduarte:
See discussion on Google Group: https://groups.google.com/g/hls4mlhelp/c/DI5o7FLZQS0/m/ChidBM-WBAAJ?utm_medium=email&utm_source=footer&pli=1

GlobalPooling1D layers are currently broken due to 
(1) bad parsing of inputs in conversion/initialization
(2) missing HLS implementations for `io_stream` and `io_parallel`

```python
~/work/hls4ml/hls4ml/model/hls_layers.py in initialize(self)
   1205 class GlobalPooling1D(Layer):
   1206     def initialize(self):
-> 1207         shape = [self.attributes['n_out'], self.attributes['n_filt']]
   1208         dims = ['N_OUTPUTS_{}'.format(self.index), 'N_FILT_{}'.format(self.index)]
   1209         self.add_output_variable(shape, dims)

KeyError: 'n_out'
```

thesps:
Resolved now #399 is merged?

PR #399, opened 2021-09-21
Fix GlobalPooling1D Layers

jmduarte:
Address #398 

Test script (to be turned into a real test): https://gist.github.com/jmduarte/99aee1eee92524ad3b1dbf5c5a863296

Changes to do:
- [x] Fix parsing of configuration for GlobalPooling1D layers
- [x] Add `global_pooling1d_cl` `io_stream` HLS implementation
- [ ] Add `global_pooling1d_cf` `io_stream` HLS implementation
- [ ] Add `global_pooling1d_cl` `io_parallel` HLS implementation
- [ ] Add `global_pooling1d_cf` `io_parallel` HLS implementation
- [ ] Add testing module

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_pooling.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_pooling.h
@@ -87,7 +87,8 @@ struct pooling1d_config{
   // IO size
   static const unsigned n_in = 10;
   static const unsigned pool_width = 2;
-  static const unsigned n_out = n_in / pool_width;
+  static const unsigned stride_width = 2;
+  static const unsigned n_out = (n_in - pool_width) / stride_width + 1;
   static const unsigned pad_left = 0;
   static const unsigned pad_right = 0;
   // Pooling function
--- a/hls4ml/converters/keras/pooling.py
+++ b/hls4ml/converters/keras/pooling.py
@@ -72,8 +72,8 @@ def parse_pooling_layer(keras_layer, input_names, input_shapes, data_reader, con
     
     return layer, output_shape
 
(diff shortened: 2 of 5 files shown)


hls4ml/templates/vivado/nnet_utils/nnet_pooling.h
```diff
@@ -87,7 +87,8 @@ struct pooling1d_config{
   // IO size
   static const unsigned n_in = 10;
   static const unsigned pool_width = 2;
-  static const unsigned n_out = n_in / pool_width;
+  static const unsigned stride_width = 2;
+  static const unsigned n_out = (n_in - pool_width) / stride_width + 1;
```
jmitrevs, line 91: These values get overwritten so I would prefer if they were explicitly indicated as dummy instead of having a formula, but I think that is a minor point, and not really part of this PR.

jmitrevs:
One issue discovered while creating the tests is that io_stream GlobalAveragePooling has a tendency to overflow when doing the sum. Unlike the io_parallel implementation, it does not use a wider accumulator size. I am trying to see what is the best way to fix this.

jmitrevs:
I added a test (and also an include of `nnet_dense.h` in `nn_conv_stream.h`). For the average overflow issue, one can manually increase the bit width of the layer (as done in `test_globalpooling1d.py`). I am not sure if it is better to automatically choose a wider width, but #327 made me wonder if it was best to leave it for the user. Let me know what you think we should do.

jmduarte:
@thesps if you could review this, that'd be great. I think it's ready, except for perhaps creating an optimizer that would set `accum_t` for `io_stream` (mimicking the way it's currently handed in the HLS for `io_parallel`): https://github.com/fastmachinelearning/hls4ml/blob/4564916cf837f7de89f232b37b3126003689f66e/hls4ml/templates/vivado/nnet_utils/nnet_pooling.h#L35

PR #402, opened 2021-09-25
Transpose2d, Concatenate2d, and up to 3 Clones for io_stream

jmduarte:
This PR has 3 independent, minor changes (I can split them into smaller PRs if needed):
- [x] Transpose2d for `io_stream`
- [x] Concatenate2d for `io_stream`
- [x] Clone streams up to 3 times (instead of just up to 2 times)
- [x] Add unit test for all 3

These features are needed, for example in this interaction network model: https://gist.github.com/jmduarte/9ea95df65feac29f9663c47a439e2d7e

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_array.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_array.h
@@ -12,10 +12,10 @@ struct transpose_config {
     static constexpr unsigned perm[3] = {2, 0, 1};
 };
 
-template<class data_T, typename CONFIG_T>
+template<class data_T, class res_T, typename CONFIG_T>
 void transpose_2d(
     data_T data[CONFIG_T::height * CONFIG_T::width],
-    data_T data_t[CONFIG_T::height * CONFIG_T::width]
+    res_T data_t[CONFIG_T::height * CONFIG_T::width]
 ) {
     #pragma HLS PIPELINE
 
@@ -26,10 +26,10 @@ void transpose_2d(
     }
 }
 
-template<class data_T, typename CONFIG_T>
+ template<class data_T, class res_T, typename CONFIG_T>
 void transpose_3d(
     data_T data[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width],
-    data_T data_t[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width]
+    res_T data_t[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width]
 ) {
     unsigned dims[3] = { CONFIG_T::depth, CONFIG_T::height, CONFIG_T::width };
     unsigned dims_t[3];
--- a/hls4ml/templates/vivado/nnet_utils/nnet_array.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_array.h
@@ -12,24 +12,24 @@ struct transpose_config {
     static constexpr unsigned perm[3] = {2, 0, 1};
 };
 
-template<class data_T, class res_T, typename CONFIG_T>
+template<class data_T, typename CONFIG_T>
 void transpose_2d(
     data_T data[CONFIG_T::height * CONFIG_T::width],
-    res_T data_t[CONFIG_T::height * CONFIG_T::width]
+    data_T data_t[CONFIG_T::height * CONFIG_T::width]
 ) {
     #pragma HLS PIPELINE
 
     for (int i = 0; i < CONFIG_T::height; i++) {
         for (int j = 0; j < CONFIG_T::width; j++) {
-	     data_t[j * CONFIG_T::height + i] = (res_T) data[i * CONFIG_T::width + j];
+            data_t[j * CONFIG_T::height + i] = data[i * CONFIG_T::width + j];
         }
     }
 }
 
-template<class data_T, class res_T, typename CONFIG_T>
+template<class data_T, typename CONFIG_T>
 void transpose_3d(
     data_T data[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width],
-    res_T data_t[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width]
+    data_T data_t[CONFIG_T::depth * CONFIG_T::height * CONFIG_T::width]
 ) {
     unsigned dims[3] = { CONFIG_T::depth, CONFIG_T::height, CONFIG_T::width };
     unsigned dims_t[3];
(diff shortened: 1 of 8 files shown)


hls4ml/templates/vivado/nnet_utils/nnet_array.h
```diff
@@ -12,24 +12,24 @@ struct transpose_config {
     static constexpr unsigned perm[3] = {2, 0, 1};
 };
 
-template<class data_T, typename CONFIG_T>
+template<class data_T, class res_T, typename CONFIG_T>
 void transpose_2d(
     data_T data[CONFIG_T::height * CONFIG_T::width],
-    data_T data_t[CONFIG_T::height * CONFIG_T::width]
+    res_T data_t[CONFIG_T::height * CONFIG_T::width]
```
thesps, line 18: What's the reason for introducing a data type for the output? `Transpose` is one of those layers that shouldn't be able to change the output precision. I think [we're still safe](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/model/hls_layers.py#L1569), but forcing it at the HLS level by using the same type doesn't seem like a bad idea.

test/pytest/test_transpose_concat.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+import hls4ml
+import numpy as np
+from tensorflow.keras.models import model_from_json, Model
+from tensorflow.keras.layers import Input, Permute, Concatenate
+import yaml
+
+@pytest.fixture(scope='module')
+def data():
+    X = np.random.rand(1, 2, 3)
```
thesps, line 10: I think we should increase the number of example to test on, bumping up the `1` -> `100` seems sensible, and the test still runs very fast.

hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h
```diff
@@ -299,6 +299,83 @@ void concatenate3d(
...
+    hls::stream<input1_T> &data1,
+    hls::stream<input2_T> &data2,
+    hls::stream<res_T> &res)
+{
+    ConcatLoopHeight: for (int i = 0; i < CONFIG_T::n_elem1_0; i++) {
+        #pragma HLS PIPELINE II=1
+
+        input1_T in_data1 = data1.read();
+	input2_T in_data2 = data2.read();
+	res_T out_data;
```
thesps, line 349: Minor: the indentation is a bit inconsistent

hls4ml/templates/vivado/nnet_utils/nnet_array.h
```diff
@@ -45,7 +45,7 @@ void transpose_3d(
                 idx_t[1] = idx[CONFIG_T::perm[1]];
                 idx_t[2] = idx[CONFIG_T::perm[2]];
 
-                data_t[idx_t[0] * dims_t[1] * dims_t[2] + idx_t[1] * dims_t[2] + idx_t[2]] = data[idx[0] * dims[1] * dims[2] + idx[1] * dims[2] + idx[2]];
+                data_t[idx_t[0] * dims_t[1] * dims_t[2] + idx_t[1] * dims_t[2] + idx_t[2]] = (res_T) data[idx[0] * dims[1] * dims[2] + idx[1] * dims[2] + idx[2]];
```
thesps, line 48: It's related to the other comment, but somehow doing a cast here makes me nervous

hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h
```diff
@@ -299,6 +299,83 @@ void concatenate3d(
     }
 }
 
+template<class input1_T, class input2_T, class res_T, typename CONFIG_T>
+void concatenate2d_0(
+    hls::stream<input1_T> &data1,
+    hls::stream<input2_T> &data2,
+    hls::stream<res_T> &res)
```
thesps, line 306: Unlike the transpose, it makes some sense that these types can be different because `data1` and `data2` could have gone through different layers and acquired different types. That said, there's still a risk that we could run into the kind of issue from #363. Maybe we can do something to help, like set a default value of the output type that would be lossless - like with integer bits `max(data1 integer bits, data2 integer bits)` and `fractional bits max(data1 fractional bits, data2 fractional bits)`

thesps (changes requested):

thesps:
Thanks for this, looks nice! To be clear, for the review, only the increase from 1 to 100 test examples is a requested change, the other items are for discussion. 

I tested with the 100 examples and it still works fine. I ran cosim as well since the io_stream version could have some dataflow behaviour: all good there too. 

Maybe something for the future: now we have layers to clone a stream 2 times and another to clone it 3 times. Can there be a `CloneN` solution?

hls4ml/templates/vivado/nnet_utils/nnet_array.h
jmduarte: The only reason I wanted to have this here is to have a consistent interface with the streaming version.

For the streaming version, we use `data_T::size` and `res_T::size` because the incoming streams and the outgoing streams should have different depths.

So it makes the python frontend code cleaner because this line works for both:
https://github.com/fastmachinelearning/hls4ml/blob/1da1c719900793eccdf04bf4a263ab5a3de47d8d/hls4ml/templates/vivado_template.py#L367

hls4ml/templates/vivado/nnet_utils/nnet_array.h
jmduarte: Right, I added the cast, but to clarify, I never expect data_T != res_T for the array version. Maybe, we should add an assert instead?

hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h
jmduarte: Right, I agree. Where do you think we could do this? In the python frontend, I suppose?

test/pytest/test_transpose_concat.py
jmduarte: Done

hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h
jmduarte: Fixed

jmduarte:
@thesps I responded to your comments and added some follow-up questions.

We should probably have a `CloneN` solution, but it's not clear to me how to handle the HLS templates for that.

hls4ml/templates/vivado/nnet_utils/nnet_array.h
thesps: Yeah that makes sense, and there is protection against them being different in the Python.

hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h
thesps: Yeah in the Python, I think it would make sense to write an optimizer pass to do it

jmduarte: @thesps I added an optimizer that attempts to do what you suggest. Namely, it takes the max of the incoming widths
```
newwidth = max(itype1.width, itype2.width)
newint = max(itype1.integer, itype2.integer)
```
but it also tries to account for signed vs unsigned (i.e. add 1 bit if we're mixing signed + unsigned) and if one of the two has a saturation/rounding mode, it takes that (with a preference for input1 if they both have different modes).

I also augmented the test so it triggers the optimizer.

Let me know if you have any suggestions!

jmduarte:
Added `concatenate1d` here to partially resolve #436

hls4ml/templates/vivado/nnet_utils/nnet_array.h
jmduarte: removed cast

jmduarte:
@thesps @vloncar rebased this PR and I think it's much cleaner now and good to go!

PR #403, opened 2021-09-28
fix conv1d io_parallel resource

jmitrevs:
With this fix `conv_1d_resource_cl` produces the same results as `conv_1d_latency_cl` and also io_stream versions of conv_1d. I will test it a bit more thoroughly tomorrow. The final fix we should also cherry-pick into the v0.6.0 candidate branch.

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_resource.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_resource.h
@@ -146,12 +146,12 @@ void im2col_1d_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan], data_T dat
     for (int kernel_col = 0; kernel_col < CONFIG_T::filt_width; kernel_col++) {
         #pragma HLS UNROLL
 
-	ChannelLoop:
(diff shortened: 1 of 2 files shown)


thesps:
Thanks for this! I pushed the PR branch to a local one (pr/403) to run the tests. We should include a test somewhat similar to the current `test_keras_mnist.py` to test a Conv1D model with the variations: `Strategy: {Resource, Latency}`, `IOType: {io_parallel, io_stream}` and check for functional correctness. But it could be a synthetic model rather than a trained one.

Regarding the implementation of the fix, I see you reversed the order of the loops in the HLS. Did you check that [the transpose that we do](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/model/hls_layers.py#L672) of the weights is actually the correct one?

jmitrevs:
I looked at the transpose of the weights but I didn't consider changing it because I thought io_stream resource might use it, and io_stream resource seemed to work fine. I didn't look in any detail, though.

thesps (approved):
I implemented a Conv1D test that checks the hls4ml model accuracy relative to the Keras original for [io_parallel, io_stream] x [resource, latency].
It fails on master branch for (io_parallel, resource) but they all pass with this PR.

Issue #377, opened 2021-08-13
Fix test_quantizer with quantized_relu(4) and quantized_relu(10)


jmduarte:
Fix test_quantizer with quantized_relu(4) and quantized_relu(10)

PR #405, opened 2021-09-29
Fixes: quantized_relu & unsigned profiling

thesps:
Note: bugfixes targeting release branch

## `quantized_relu`

Since [this change](https://github.com/fastmachinelearning/hls4ml/commit/02624c5fc491cc221326eac631b7219e76e0eef8#diff-8ec43234900e263f95104d90f7969c22a9d314817d20ce4d1f23f1a4354fc91a), we assign the wrong data type to `quantized_relu` activations. It sets the correct type for `quantized_bits`, just not `quantized_relu`. This is what causes the tests flagged in #377 to fail. 

Using the example of the hls4ml-tutorial Part3 QKeras model which uses `quantized_relu(6)` activations, on master branch we assign `ap_fixed<6,1>`. But appropriate choices would be instead `ap_fixed<7,1>` (signed) or `ap_ufixed<6,0>` (unsigned). I've chosen to go for the `ap_ufixed` case for the fix since it uses fewer bits. 

You can see that the types we currently assign to the `relu` layers are wrong in the profiling:

Screenshot 2021-09-29 at 16 35 10

The model accuracy is:

- 0.743656 QKeras on CPU (reference / target)
- 0.743542 hls4ml on master branch
- 0.743855 hls4ml on this PR (identical if `ap_fixed<7,1>` was used instead)

The synthesis results slightly prefer the `ap_ufixed` variation which has one fewer bit than the necessary `ap_fixed` option.

| `quantized_relu` precision | Accuracy | Latency | II | LUT | FF | DSP | BRAM18 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| `ap_ufixed<6,0>` | 0.743855 | 9 | 1 | 33706 | 3037 | 119 | 4 |
| `ap_fixed<7,1>` | 0.743855 | 9 | 1 | 37523 | 3155 | 119 | 4 |

The `quantized_relu` tests now pass, so they are included again, resolving #377. I added a few more variations as well.

## unsigned `profiling`
In fixing the above, I noticed that we don't display unsigned types correctly in the profiling, so I include the fix for that as well. (Separate commits in case they need to be taken separately).

Before (master):

Screenshot 2021-09-29 at 16 53 33

After (this PR):

Screenshot 2021-09-29 at 16 32 10

### DIFF
--- a/hls4ml/model/profiling.py
+++ b/hls4ml/model/profiling.py
@@ -159,32 +159,31 @@ def types_histogram(data, fmt='longform'):
 types_plots = {'boxplot' : types_boxplot,
                'histogram' : types_histogram}
 
-def ap_fixed_WIF(dtype):
+def ap_fixed_WIFS(dtype):
     from hls4ml.templates.vivado_template import VivadoBackend
     dtype = VivadoBackend.convert_precision_string(None, dtype) 
-    W, I, F = dtype.width, dtype.integer, dtype.fractional
-    return W, I, F
+    W, I, F, S = dtype.width, dtype.integer, dtype.fractional, dtype.signed
+    return W, I, F, S
 
 def types_hlsmodel(model):
     suffix = ['w', 'b']
     data = {'layer' : [], 'low' : [], 'high' : []}
     # Plot the default precision
     default_precision = model.config.model_precision['default']
-    # assumes ap_fixed
-    W, I, F = ap_fixed_WIF(default_precision)
+    W, I, F, S = ap_fixed_WIFS(default_precision)
     data['layer'].append('model')
     data['low'].append(-F)
-    data['high'].append(I-1)
+    data['high'].append(I-1 if S else I)
 
     for layer in model.get_layers():
         for iw, weight in enumerate(layer.get_weights()):
             wname = '{}/{}'.format(layer.name, suffix[iw])
             T = weight.type
             if T.name != 'model':
-                W, I, F = ap_fixed_WIF(T.precision)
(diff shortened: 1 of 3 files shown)


hls4ml/model/profiling.py
```diff
@@ -159,49 +159,48 @@ def types_histogram(data, fmt='longform'):
 types_plots = {'boxplot' : types_boxplot,
                'histogram' : types_histogram}
 
-def ap_fixed_WIF(dtype):
+def ap_fixed_WIFS(dtype):
```
jmduarte, line 162: does this support `ap_[u]int` datatypes? Seeing as they seem to be supported in [`config.py`](https://github.com/fastmachinelearning/hls4ml/pull/405/files#diff-8ec43234900e263f95104d90f7969c22a9d314817d20ce4d1f23f1a4354fc91aR63) maybe we should support them here too?

jmduarte (commented):
Thanks @thesps! This looks really good. I just have one minor comment

hls4ml/model/profiling.py
thesps: Yeah even though the function is called `ap_fixed_WIFS`, it does also work for the int types since the `IntegerPrecisionType` and `FixedPrecisionType` have the same members. [Here's a demo](https://gist.github.com/thesps/bd862e2bb1e91710fb2809676e2fff2e) evaluated on this PR branch. It also runs on master, but the box for the unsigned case is displayed wrongly as for `ap_fixed`.

jmduarte (approved):
Looks good!

PR #407, opened 2021-09-30
Speed up CI tests

thesps:
As we've been adding more tests, the CI testing takes longer to run. The timing varies a lot, but to begin with it would take ~15-20 minutes, while recently with more tests it's taking ~20-30 minutes. This PR speeds things up a bit by submitting each `test_*.py` file as a separate CI job, which can be assigned to different workers. Previously we just ran all pytests in one job, so each test would run sequentially. As we continue to develop more tests this should help avoid the CI really dragging.

Here the pipeline is [generated dynamically](https://www.objectif-libre.com/en/blog/2021/02/23/a-new-era-for-gitlab-ci-dynamic-child-pipelines/): the main `.gitlab-ci.yml` job runs the `generate_ci_yaml.py` script, which creates a `.yml` file defining all the jobs from the available test files. That's uploaded as an artifact and used to trigger the next stage. So to add a new test file you can still just add the `test_something.py` file to the pytests directory and it will be picked up by the CI.

Here's a view of it running, with some jobs finished and some still going:

Screenshot 2021-09-30 at 17 03 43

The timing seems to now be in the 10-15 minute range, pretty much limited to the time it takes to run the slowest test file - usually the MNIST CNN one.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,22 +1,27 @@
-pytest:
-  image: gitlab-registry.cern.ch/fastmachinelearning/hls4ml-testing:0.2.base
+stages:
+  - generate
+  - trigger
+  - test
+
+generator:
+  stage: generate
+  image: python:3.7-alpine
   tags: 
     - docker
   before_script:
-    - source ~/.bashrc
-    - git submodule init
-    - git submodule update
-    - conda activate hls4ml-testing
-    - pip install .[profiling]
-  script:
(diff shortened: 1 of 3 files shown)


vloncar (approved):

vloncar:
Note: This PR checks for example models in a bit of a hacky way. There is currently no cleaner way to do this, but this will be addressed in the future through an API for accessing example models.

PR #420, opened 2021-10-20
`axi_stream_driver` update

nicologhielmetti:


hls4ml/templates/vivado_accelerator/pynq-z2/python_drivers/axi_stream_driver.py
```diff
@@ -4,32 +4,36 @@
...
-        
-        super().__init__(bitfile_name, dtbo=dtbo, download=download, ignore_version=ignore_version, device=device)
-        
+    def __init__(self, bitfile_name, x_shape, y_shape, dtype=np.float32, dtbo=None, download=True, ignore_version=False,
+                 device=None):
+        super().__init__(bitfile_name, dtbo=None, download=True, ignore_version=False, device=None)
+        self.sendchannel = self.hier_0.axi_dma_0.sendchannel
+        self.recvchannel = self.hier_0.axi_dma_0.recvchannel
+        self.input_buffer = allocate(shape=x_shape, dtype=dtype)
+        self.output_buffer = allocate(shape=y_shape, dtype=dtype)
```
thesps, line 15: How much performance is gained from moving the buffer allocation to the `__init__`? We lose some flexibility this way since the size is fixed in the constructor, whereas previously it was sized based on the input size in `predict`.

thesps (approved):

Issue #423, opened 2021-10-26
qkeras_mnist_dense example fails to convert


siorpaes:
I am trying the 'qkeras_mnist_dense' example using the following script derived from the Getting Started section:

```
import hls4ml

#Fetch a keras model from our example repository
#This will download our example model to your working directory and return an example configuration file
config = hls4ml.utils.fetch_example_model('qkeras_mnist_dense.json')

print(config) #You can print the configuration to see some default parameters

#Convert it to a hls project
hls_model = hls4ml.converters.keras_to_hls(config)

# Print full list of example models if you want to explore more
hls4ml.utils.fetch_example_list()

print(hls_model)

#Use Vivado HLS to synthesize the model
#This might take several minutes
hls_model.build()

#Print out the report if you want
hls4ml.report.read_vivado_report('my-hls-test')
```

The model is being downloaded but the process stops almost immediately with this error:

```
2021-10-26 10:48:02.100919: W tensorflow/stream_executor/platform/default/dso_loader.cc:59] Could not load dynamic library 'libcudart.so.10.1'; dlerror: libcudart.so.10.1: cannot open shared object file: No such file or directory
2021-10-26 10:48:02.100967: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
Downloading example model files ...
Downloading configuration files ...
Traceback (most recent call last):
  File "qkeras_mnist_dense.py", line 5, in <module>
    config = hls4ml.utils.fetch_example_model('qkeras_mnist_dense.json')
  File "/home/siorpaed/anaconda3/envs/hls4ml-tutorial-0.5.0/lib/python3.7/site-packages/hls4ml/utils/example_models.py", line 145, in fetch_example_model
    config = _load_example_config(model_name)
  File "/home/siorpaed/anaconda3/envs/hls4ml-tutorial-0.5.0/lib/python3.7/site-packages/hls4ml/utils/example_models.py", line 92, in _load_example_config
    config = yaml.load(ymlfile)
TypeError: load() missing 1 required positional argument: 'Loader'
```
The Conda environment being used is the usual 'hls4ml-tutorial-0.5.0'

Thanks

David

thesps:
Thanks for this, I guess it's related to the `yaml` version. I'll take a proper look soon, but thanks for also providing the fix.

siorpaes:
Thanks to you. Let me know if you need any other input.

[...]

PR #424, opened 2021-10-26
Fixed 'qkeras_mnist_dense' example build problem #423

siorpaes:
Should fix issue #423


thesps (approved):
This change is needed due to the update from `pyyaml-5.4.1` to `pyyaml-6.0` that is picked up when doing `pip install hls4ml`.

PR #432, opened 2021-11-02
Rework to use Constant nodes in intermediate representation

jmitrevs:
Based on discussion with @vloncar, I introduced a Constant node. The idea is as follows. The initial conversion from onnx to the hls4ml model should be as simple as possible, and then we build things up with optimizers. Constants, including weights, come in through Constant nodes, and the Quantizers then more easily apply their quantization on the weight inputs, producing weight outputs, which with the MatMul node, can be combined to a Dense layer with weights. The scalings and biases also come in via Merge nodes with a constant. I think it simplifies things.

The merge request is not complete. The optimization steps are not implemented, and I haven't looked at convolution yet. It seems to be able to ingenst TFC_2W2A_clean.onnx (though of course the compilation step fails.)

Any thoughts on this implementation? Note it's a PR to the ingest-qonnx development branch, not the main branch.


jmitrevs:
There were no objects to this, so I will merge it with ingest-qonnx and continue development there.

PR #435, opened 2021-11-03
Update for pyyaml 6.0

thesps:
From our `setup.py` we pick up the latest `pyyaml` with `pip install hls4ml`. Since `pyyaml`'s `6.0` release a few weeks ago the `Loader` argument is required by `yaml.load`, causing a few issues in `hls4ml`. The pytests currently fail on master branch as a consequence.

This PR & #423 fix all the broken `yaml.load` occurrences in the code base.

### DIFF
--- a/hls4ml/converters/__init__.py
+++ b/hls4ml/converters/__init__.py
@@ -93,7 +93,7 @@ def parse_yaml_config(config_file):
 
     print('Loading configuration from', config_file)
     with open(config_file, 'r') as file:
-        parsed_config = yaml.load(file, Loader=yaml.SafeLoader)
+        parsed_config = yaml.safe_load(file)
     return parsed_config
 
 def convert_from_config(config):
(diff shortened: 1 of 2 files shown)


vloncar:
For the sake of consistency of used library calls, you may want to change loading of yaml in the converter [init](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/converters/__init__.py#L96)

thesps:
Done, but untested. Jenkins tests should probe that, but not pytests I think.

vloncar:
It will not break, `safe_load` is just a shortcut for that `load(..., Loader=yaml.SafeLoader)`. [According](https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/lib/yaml/__init__.py#L117) to the pyyaml code.

vloncar (approved):

Issue #415, opened 2021-10-12
Compilation error with concatenate layer


rdhen3:
Hi there, I have just started using hls4ml a couple of days ago with the hope of applying it to a simple SSD inspired CNN. 
The model involves some number >= 0 of convolutional feature extraction layers well as several parallel sets of predictor layers whose outputs needs to be concatenated together. 

A simple version of the model has the following keras summary:
![image](https://user-images.githubusercontent.com/88073332/136896961-f9a25ee4-3748-4bd3-a27a-c70fde3e268f.png)

The current stable build seems to have issues in processing concatenate layers in that i think the shape is incorrectly inferred leading to an IndexError in `hls4ml/model/hls_layers.py`:
```
   1380         axis = self.attributes['axis']
   1381         shape = inp1.shape[:]
-> 1382         shape[axis] += inp2.shape[axis]
   1383         rank = len(shape)
   1384         if rank > 1:

IndexError: list index out of range
```

I tried pulling the latest master upon seeing issue #361 and it seems that the shape/indexing problem was solved by this and is get the following hls4ml summary of the model:
![image](https://user-images.githubusercontent.com/88073332/136898001-38bc231a-eccb-477f-9596-0a747f9c19ca.png)
which seems to match up to the original model correctly. 

However, now I have hit a compilation error when building/compiling the model:
![image](https://user-images.githubusercontent.com/88073332/136898331-0f4d589b-0616-48c6-81a8-6ff3dcb0dc14.png)

I thought perhaps this related to the update going on in #402 so I tried swapping to that branch and trying again but received an identical error. Any thoughts?

jmduarte:
@rdhen3 can you provide a script to reproduce the error? I do believe the latest error is related to issues being addressed in #402.

Also, to confirm are you using `io_stream` or `io_parallel`?

rdhen3:
Indeed I did notice the behaviour of unrolling the output to 1D. Is this done in the same way as, say, a keras `Flatten` layer, or by extension the same as `numpy.reshape`? 
I believe I could indeed remove the `Reshape` layer but would still need to implement it outside the network the output at some point in order to decode the predictions correctly.

jmduarte:
Yes, in your particular case, I would do this:
```python
model = Model(inputs=x, outputs=classes_boxes)
...
keras_prediction = model.predict(X_input).reshape((X_input.shape[0], 16, 6))
...
[...]

Issue #408, opened 2021-10-02
Flatten layer creating a lot of logic


jmduarte:
Due to #378, Keras `Flatten` layers are now mapped to hls4ml `Reshape` layers instead of being skipped. While harmless for `io_parallel` implementations, for `io_stream`, these result in the creation of `repack_stream` through the [`ReshapeStream`](https://github.com/fastmachinelearning/hls4ml/blob/4564916cf837f7de89f232b37b3126003689f66e/hls4ml/model/optimizer/passes/repack_stream.py#L71) optimizer which changes the depth of the `hls::stream` being used.

The most common case I see this happening is for CNN classifiers with `Conv2D -> Flatten -> Dense`. It seems this creates a lot of (unnecessary?) logic, as the same model works when skipping the optimizer and produces a lot less logic.

As an example, the [TinyML Conv2D](https://github.com/fastmachinelearning/platform_ml_models/blob/master/eembc/CIFAR10_ResNetv1/tiny.yml) model with master gives (example [here](https://www.dropbox.com/s/5va00k55gafae7t/my-hls-test-quantized-tiny.tar.gz?dl=0)) :
```
+-----------------+---------+-------+--------+-------+-----+
|       Name      | BRAM_18K| DSP48E|   FF   |  LUT  | URAM|
+-----------------+---------+-------+--------+-------+-----+
|DSP              |        -|      -|       -|      -|    -|
|Expression       |        -|      -|       0|      2|    -|
|FIFO             |      163|      -|   16481|  54096|    -|
|Instance         |      100|      0|   47685|  79522|    0|
|Memory           |        -|      -|       -|      -|    -|
|Multiplexer      |        -|      -|       -|      9|    -|
|Register         |        -|      -|       1|      -|    -|
+-----------------+---------+-------+--------+-------+-----+
|Total            |      263|      0|   64167| 133629|    0|
+-----------------+---------+-------+--------+-------+-----+
|Available        |      280|    220|  106400|  53200|    0|
+-----------------+---------+-------+--------+-------+-----+
|Utilization (%)  |       93|      0|      60|    251|    0|
+-----------------+---------+-------+--------+-------+-----+
```

while if we skip the optimizer we get (example [here](https://www.dropbox.com/s/iyg6viogzvb8rvl/my-hls-test-quantized-tiny-skipopt.tar.gz?dl=0)):
```
+-----------------+---------+-------+--------+-------+-----+
[...]

PR #443, opened 2021-11-07
Reshape fixes: don't repack stream for flatten; remove final reshape

jmduarte:
2 reshape-related fixes:
- Fix to `ReshapeStream` optimizer: don't repack stream for flatten; Addresses #408
  -  Resources for TinyML model are consistent with skipping optimizer:
```
+-----------------+---------+-------+--------+-------+-----+
|       Name      | BRAM_18K| DSP48E|   FF   |  LUT  | URAM|
+-----------------+---------+-------+--------+-------+-----+
|DSP              |        -|      -|       -|      -|    -|
|Expression       |        -|      -|       0|      2|    -|
|FIFO             |      163|      -|    6241|  13136|    -|
|Instance         |       99|      0|   30015|  34856|    0|
|Memory           |        -|      -|       -|      -|    -|
|Multiplexer      |        -|      -|       -|      9|    -|
|Register         |        -|      -|       1|      -|    -|
+-----------------+---------+-------+--------+-------+-----+
|Total            |      262|      0|   36257|  48003|    0|
+-----------------+---------+-------+--------+-------+-----+
|Available        |      280|    220|  106400|  53200|    0|
+-----------------+---------+-------+--------+-------+-----+
|Utilization (%)  |       93|      0|      34|     90|    0|
+-----------------+---------+-------+--------+-------+-----+
```
- Introduce `RemoveFinalReshape` optimizer to remove reshape if it's the final layer; Addresses #415
  - Currently, this is in the same file `repack_stream.py`, but it could be moved
  - This script now works out of the box: https://gist.github.com/jmduarte/4c91f28ab570036aa4cef63db0eb2322
  - Should we add a pytest similar to the above script?


hls4ml/model/optimizer/passes/repack_stream.py
```diff
@@ -121,3 +122,14 @@ def transform(self, model, node):
...
+
+class RemoveFinalReshape(OptimizerPass):
+    ''' Remove reshape if final layer '''
+    def match(self, node):
+        # match if reshape is final node
+        return node.__class__.__name__ == 'Reshape' and not node.get_output_nodes()
+
+    def transform(self, model, node):
+        # remove, but don't rewire because it's the output layer
+        model.remove_node(node, rewire=False)
```
vloncar, line 134: I don't think we should always do this, `io_stream` should still reshape the tensor if the model explicitly requested it.

thesps: I think the idea is that when the `Reshape` is the output layer, it basically has no effect since the output is just a stream, which doesn't carry any inherent 'shape' information. But having the layer in seemed to cause the problem in #415

thesps: See the test in #444 which passes for `io_stream` whether or not there is a `Reshape` (`repack_stream`) as the final operation (executed on master vs this PR, or only on this PR skipping the new optimizer)

hls4ml/model/optimizer/__init__.py
```diff
@@ -41,6 +41,7 @@
 register_pass('optimize_pointwise_conv', OptimizePointwiseConv)
 register_pass('clone_output', CloneOutput)
 register_pass('reshape_stream', ReshapeStream)
+register_pass('remove_final_reshape', RemoveFinalReshape)
```
thesps, line 44: When using `io_stream`, the `Reshape` layer is replaced with a `Repack` so the `RemoveFinalReshape` doesn't pick it up. The model still works, but that final reshape is redundant and costs some unnecessary resources, as in the `Flatten` case. 

If we swap the order of these lines to run `RemoveFinalReshape` before `ReshapeStream`, then we can additionally remove `Reshape` as the final layer for `io_stream` and save some resources.

hls4ml/model/optimizer/passes/repack_stream.py
vloncar: We can't say output is "just a stream", streams hold packed data (along the innermost dimension). If we always discard the last Reshape layer, we effectively lose the option of controlling the output of the IP.

thesps: Okay I take the point about the packed data - so the `Reshape` is only a `nop` if the innermost dimension size doesn't change, right? But I do feel like controlling the model output dataflow with a `Reshape` layer is not a desirable feature

vloncar: It is more desirable that the tool does what it is expected as opposed to silently ignoring the user's wishes without even telling them "you don't want this, we know better than you".

thesps: Sure, but is it really reasonable to expect that something like reshaping `(3, 3, 2) -> (18)` would change the dataflow? It's true for Vivado backend io_stream, but not io_parallel, and generally would depend on the backend implementation. This doesn't have to be silent either, we could add a warning in the printout, and optimizer passes can be skipped.

thesps: Also 

> "you don't want this, we know better than you".

Isn't this an alternative definition of optimizer passes? 😇

jmduarte: Modified as discussed offline to just remove the final reshape layer for `io_parallel` and warn for both `io_parallel` and `io_stream`

hls4ml/model/optimizer/__init__.py
jmduarte: Swapped, just in case we want to remove it in the future for `io_stream`, but currently we are keeping it for `io_stream`

thesps:
I'm happy with the state of both developments:
- Don't repack stream for flatten: tested the `example-models/keras/qkeras_mnist_cnn` on pynq-z2, getting identical performance to master (NN accuracy & throughput), but for fewer resources
- `RemoveFinalReshape` works as expected

Only question: do you want the test case for the second item from #444 ? (It's a PR to this branch)

jmduarte:
Yes, now merged!

thesps (approved):

PR #444, opened 2021-11-08
Add a test for a model with Reshape as the final layer

thesps:
I added a very simple test model with a `Reshape` as the final layer. The only problem is... this works on both master branch and the PR #443 (ie either with or without `repack_stream` as the final operation in the `myproject.cpp`).

I don't know yet what's the significant difference wrt @jmduarte 's [gist](https://gist.github.com/jmduarte/4c91f28ab570036aa4cef63db0eb2322)

### DIFF
--- a/test/pytest/test_graph.py
+++ b/test/pytest/test_graph.py
@@ -1,6 +1,7 @@
 import hls4ml
 import numpy as np
 import pytest
+import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


thesps:
Correction: the test fails on master branch for `io_parallel` but works with `io_stream`.

Issue #437, opened 2021-11-04
QKeras bad predictions


HenningCode:
Hello Guys,
with the fix from yesterday I tried the example "qkeras_mnist_dense" example model, but the predictions are pretty bad. 

I then tried to create my own QKeras model and tried it for a short amount of time on the mnist dataset just to see if thats the also the case with my own model and the results were pretty bad again. Am I doing something wrong?

This is the code to test the example model: (for this I used the latest branch form the Github)

```
config = hls4ml.utils.fetch_example_model('qkeras_mnist_dense.json')
print_dict(config)
hls_model = hls4ml.converters.keras_to_hls(config)


(x_train, y_train), (x_test, y_test) = mnist.load_data()


RESHAPED = 28*28
NB_CLASSES = 10

x_train = x_train.astype("float32")
x_test = x_test.astype("float32")

x_train = x_train.reshape(x_train.shape[0],RESHAPED)
x_test = x_test.reshape(x_test.shape[0],RESHAPED)

#Tested with and without 
x_train /= 256 
x_test /= 256


y_train = to_categorical(y_train, NB_CLASSES)
y_test = to_categorical(y_test, NB_CLASSES)


hls_model.compile()
print('Ground truth\n',y_test[0:5])
print('HLS Predict:\n',hls_model.predict(x_test[0:5]))
```

This is the script I am using to test the QKeras Model. If I change the single QDense layer to a normal Dense layer the predictions of the HLS model are pretty close to the predictions of the keras model. Why is that the case? (For this I was using the pip installation)

```
def count_errors(x,y):
    error = 0
    for i in range(len(x)):
        if(not np.array_equal(x[i],y[i])):
            error += 1

    return error
        

print("============================================================================\n\n\n")

NB_EPOCH = 2
BATCH_SIZE = 64
VERBOSE = 1
NB_CLASSES = 10
OPTIMIZER = Adam(learning_rate=0.0001, decay=0.000025)
VALIDATION_SPLIT = 0.1
BUILDING = 0

(x_train, y_train), (x_test, y_test) = mnist.load_data()

RESHAPED = 784

x_test_orig = x_test

x_train = x_train.astype("float32")
x_test = x_test.astype("float32")


x_train = x_train.reshape(x_train.shape[0], RESHAPED)
x_test = x_test.reshape(x_test.shape[0], RESHAPED)

#Here only tested with, without the accuracy is bad even without QDense
x_train /= 256
x_test /= 256

[...]

Issue #375, opened 2021-08-12
Fix accuracy test for MNIST CNN with IOType=io_parallel and Strategy=resource 


jmduarte:
_Originally posted by @thesps in https://github.com/fastmachinelearning/hls4ml/issues/362#issuecomment-897581793_

`FAILED test_cnn_mnist.py::test_accuracy[resource-io_parallel] - assert (0.9885 > 0.95 and 0.8867981790591806 < 0.01)`
test_cnn_mnist.py fails for IOType=io_parallel and Strategy=resource.

jmduarte:
Should be solved with #448

PR #448, opened 2021-11-09
Fix Conv2D with `io_type = io_parallel` & `Strategy: Resource`

thesps:
Currently a `Conv2D` layer with `io_type = io_parallel` & `Strategy: Resource` gives wrong results. The loops of `im2col_2d_cl` needed reordering.

`test_cnn_mnist.py` probes the issue. On master branch the problematic combination is excluded, but if we run it we see:

```
Accuracy keras:      0.9838
Accuracy hls4ml:     0.112
Relative difference: 0.8861557227078675

>     assert acc_keras > 0.95 and rel_diff < 0.01
E     assert (0.9838 > 0.95 and 0.8861557227078675 < 0.01)

test_cnn_mnist.py:72: AssertionError
```

Now it can be included, passing with:

```
Accuracy keras:      0.9838
Accuracy hls4ml:     0.9834
Relative difference: 0.0004065867046147143
```

I've also reduced the number of MNIST images used in the test from 10000 to 5000, since that test was already the slowest and this PR adds another iteration of it.

Solves issue #375 , and also seen in issue #437.

For completeness: this doesn't touch `io_stream` implementations, they're all fine.
I haven't touched `im2col_2d_cf`, but is that even accessible?

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv2d_resource.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv2d_resource.h
@@ -154,20 +154,18 @@ void im2col_2d_cl(
     const int col)
 {
     int index = 0;
-    for (int channel = CONFIG_T::n_chan; channel--; data++) {
+    for (int kernel_row = 0; kernel_row < CONFIG_T::filt_height; kernel_row++) {
         #pragma HLS UNROLL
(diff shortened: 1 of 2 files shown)


vloncar (approved):
LGTM! Thanks @thesps

Issue #263, opened 2020-12-14
Reshape layer issue


Mmiglio:
Hi all, 

I'm having trouble while compiling a model similar to this

```
model = tf.keras.models.Sequential([
        tf.keras.layers.Input((10)),
        tf.keras.layers.Dense(10*3),
        tf.keras.layers.Reshape((10,3)),
        tf.keras.layers.Softmax()
])
```

These are the errors: 

![Screenshot 2020-12-14 at 11 59 57](https://user-images.githubusercontent.com/26923771/102073725-020b0500-3e04-11eb-9a6b-4e551878a087.png)

<img width="636" alt="Screenshot 2020-12-14 at 12 00 28" src="https://user-images.githubusercontent.com/26923771/102073750-09321300-3e04-11eb-8160-913dc2ebd511.png">

My guess is that the problem is someway related to the combination of Reshape + softmax. Thanks in advance for the help!

jmduarte:
@vloncar I was wondering if this would be solved by #384 but the bug seems to be still present. I think the issue is that Reshape is handled by just an InPlaceVariable, which should refer to `layer2_out`, but somehow the (nonexistent) `layer3_out` is used as input to softmax.

gemarcano:
Seeing that #384 has been merged and I'm still seeing this problem with a similar model construction, I don't think that was the solution. That said, I also arrived at the same general area of the code by noticing the `InPlaceVariable` definition in the reshape handler class while debugging this problem.

Prior to commit 7f75add50a5acd2a4335bde0ab98c9d4e79e1137 I was able to build the model, but that's because one of the optimization passes was replacing the Reshape with a Repack, which doesn't seem to have the same problem.

I'm going to see if I can make some sense as to what's going on. At a first glance, every other handler appears to be using `self.add_output_variable` instead of calling `self.model.register_output_variable` explicitly. Then again, Reshape is the only one using InplaceVariable, at least in the `hls4ml/model/hls_layers.py` file.

gemarcano:
I'm not sure, but I think the problem is that for `InplaceVariable`, method `variable_definition_cpp` in `hls4ml/writer/vivado_writer.py` returns `None`, and method `function_cpp` in `hls4ml/model/hls_layers.py` returns `None`. In other words, nothing is output for this layer in the project C++ file.

gemarcano:
[...]

PR #468, opened 2021-12-07
Inplace variable fix

jmitrevs:
This pull request fixes the reference of an `InplaceVariable` when optimizers remove the reference. It was reported in Issue #263.

The solution adds an `update_inplace_variables` member function to Layers, with the default implementation being just `pass`. If a layer does have an `InplaceVariable`, however, it must update the reference. Such a function is implemented for the `Reshape` layer.

The added pytest fails in the current main branch, but succeeds with this fix.

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -402,7 +402,7 @@ class Layer(object):
     def get_output_nodes(self, output_name=None):
         if output_name is None:
             output_name = self.outputs[0]
-        return [node for node in self.model.graph.values() if node.inputs[0] == output_name]
+        return [node for node in self.model.graph.values() if len(node.inputs) and node.inputs[0] == output_name]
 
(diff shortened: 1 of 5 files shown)


jmitrevs:
The cherry-pick included in this PR did remove trailing whitespace and add a test that inputs don't have length zero before taking inputs[0], which strictly speaking are not needed for this fix, but I think they are fine to include.

vloncar:
`InplaceVariable` is a leftover from old times when we had only `io_parallel` and arrays were flat, so `Reshape`/`Flatten` was a nop. A better approach is to remove the concept of `InplaceVariable` and handle in-place transformations in the optimizers. Adding a new function to all layer instances so that one particular class can make use of it just turns an old workaround into a feature we must support, which I think is the wrong approach.

jmitrevs:
That sounds good. I will close this PR and try to implement `Reshape` without an `InplaceVariable`.

PR #469, opened 2021-12-08
Remove InplaceVariables and the problems they cause

jmitrevs:
This attempts to fix the problem that removing nodes can cause to `InplaceVariable`s by removing `InplaceVariable`s altogether. These variables are only used for the `Reshape` layar. In `io_stream` we already replace `Reshape` nodes with `Repack` nodes; this pull request additionally removes those nodes in `io_parallel` (and flatten), changing the shape of the preceding variable. (We may want to rename the `ReshapeStream` optimizer pass to reflect on the additional duties it performs.)

I added a basic test, though it does not test the flatten case.

@vloncar, is this the direction we should go?

### DIFF
--- a/hls4ml/model/hls_layers.py
+++ b/hls4ml/model/hls_layers.py
@@ -194,23 +194,6 @@ class StreamVariable(Variable):
     def size_cpp(self):
         return '*'.join([str(k) for k in self.dim_names])
 
-class InplaceVariable():
-    def __init__(self, shape, dim_names, proxy, **kwargs):
-        self.shape = shape
-        self.dim_names = dim_names
-        self.type = proxy.type
-        self.name = proxy.name
-        self.size = proxy.size
-
-    def get_shape(self):
-        return zip(self.dim_names, self.shape)
-
-    def definition_cpp(self):
-        return None
-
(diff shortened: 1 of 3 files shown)


vloncar:
Thanks @jmitrevs this is what I meant. I'll check it later, basically we only need to see if it broke reshape/flatten (that you test) and multidimensional softmax (i.e., cnn without flatten)

thesps:
I pushed the PR branch to local branch PR/469 to trigger pytest CI. I notice the cnn_mnist test- that contains a flatten - is failing

jmitrevs:
Thanks. I'll try to fix the bugs.

jmitrevs:
This approach for a solution is more problematic than I originally thought. The output shape is often referenced in `function_cpp`, so changing it can break the produced code, while not changing it will make the input variable shape wrong in the subsequent layer. I wonder if we need to keep `InplaceVariable`s in some form, maybe defining it as a reference of the input variable. I will try that approach.

jmitrevs:
The logic from here now is included in #472 (with changes to pass the cnn_mnist test), so closing this PR.

PR #471, opened 2021-12-10
Pointwise conv1d resource

jmduarte:
- Introduce pointwise conv1d resource


thesps:
Can we have a test case to show it works?

vloncar:
I included the missing 2d implementation and a test case.

vloncar (approved):

PR #472, opened 2021-12-12
Implement InplaceVariables with a reference

jmitrevs:
This is an alternate of #469 to fix the issues caused by `InplaceVariable`s and optimzers removing their reference. This is how I see it:

* The implementation in the master branch creates the reference of an `InplaceVariable` when it is created. This runs into problems with optimizers changing the graph and making the reference no longer valid. Adding hooks to `remove_node` and `replace_node` seemed too much machinery for variables that are used only in `Reshape`  layers.
* The fix in #469 completely removes `InplaceVariable`s and changes the shape of the preceding node's output. However, the cnn_mnist test showed that this isn't really safe to do, because `function_cpp` calls often make use of the output variable shape, so changing the shape breaks the code generation.
* This attempt tries to localize the `InplaceVariable` code without adding more machinery. `InplaceVariable`s become more like regular variables, but they have a blank `definition_cpp`.  Instead the `Reshape` layer gets a `function_cpp` that makes the connection between the input at output variable at that stage, after the optimzers have run. The `function_cpp` basically becomes a typedef for the type and defines the output variable to be a referece to the input variable. Note, unlike the previous option, this allows for the shapes to differ while effectively becoming a nop in the code. Of course in `io_stream` cases the `Reshape` is often changed to a `Repack` node. This PR has no effect on that case.

Both the new `test_reshape` and the existing `test_cnn_mnist` tests succeed.

Let me know what you think. I left #469 open because I am not sure what is the best direction, though of course this and that are mutually exclusive.


jmitrevs:
One thing that slightly bothers me about this implementation is that `Reshape` is only for `io_parallel` reshapes and all flattens, while an `io_stream` reshape gets converted to a `Repack`. It makes me wonder if the `Repack` and `Reshape` layers should actually be merged, with different behavior implemented for the different cases. But even if it's logically cleaner that way, the implementation may be more complicated. I am not sure what is best.

vloncar:
This adds a hack to an existing workaround that basically says that reshape/flatten is a no-op by default, and the case when it is not should be handled in a special way. I argue that the opposite should be true. The internal representation should be flexible enough to support representing any model, without assumptions about how some layers will be implemented as that is backend-specific. If Vivado backend can't live without the `InplaceVariable` an optimizer should introduce it for the `Flatten` with `io_parallel` case and do the other special thing for the `io_stream`. I still feel that the approach in #469 is the way to go, it just needs more work on other areas to make it fit. There are really two issues here, the `InplaceVariable` and the way some layers assume that the shapes are 1D. The latter affects all layers derived from `Activation` class and is the true reason why tests fail with #469. The changes to `Softmax` were the first step, but the way it is implemented now, it is simply not possible to apply activations only on a given axis with `io_parallel` as the code assumes a flat input of `n_in` elements. Another example is `PReLU` in CNN models (`Conv -> PReLU -> ...`), which incorrectly parses in hls4ml and assumes the parameter array is the size of the input tensor itself and not just the last dimension of the tensor. This blocks progress on the SSD model.

jmitrevs:
I tend to agree, as I said in my above comment, that I didn't like the way that this treated the `io_parallel` and `io_stream` cases differently. One can either merge`Repack` and `Reshape` into one more complicated `Reshape` layer you go directly to, or the `Reshape` can be just a placeholder with `Repack` and `Nop` implementations you change into with the optimizers to split the two codes. I think you are suggesting the latter more, so I can try to do that tomorrow.

As for the `InplaceVariable`, the main issue with the #469 attempt was that lots of layers (conv, pooling, etc.) in `function_cpp` and `config_cpp` access properties of the output variable, which can produced wrong results if the shape is changed in place. This involves code similar to:

```python
   def config_cpp(self):
        params = self._default_config_params()
        if self.get_attr('data_format') == 'channels_last':
            params['in_height'] = self.get_input_variable().dim_names[0]
            params['in_width'] = self.get_input_variable().dim_names[1]
            params['n_chan'] = self.get_input_variable().dim_names[2]
            params['out_height'] = self.get_output_variable().dim_names[0]
            params['out_width'] = self.get_output_variable().dim_names[1]
            params['n_filt'] = self.get_output_variable().dim_names[2]
        else:
            params['n_chan'] = self.get_input_variable().dim_names[0]
            params['in_height'] = self.get_input_variable().dim_names[1]
            params['in_width'] = self.get_input_variable().dim_names[2]
            params['n_filt'] = self.get_output_variable().dim_names[0]
            params['out_height'] = self.get_output_variable().dim_names[1]
            params['out_width'] = self.get_output_variable().dim_names[2]
        params['dilation'] = self.get_attr('dilation', 1)
        params['nzeros'] = self.get_weights('weight').nzeros
```

The `InplaceVariable` just introduces another variable you can change the shape of safely. The only thing that's special about an `InplaceVariable` in this case is just that `definition_cpp` needs to return None, since you don't want to create extra variables; you want to create references.

jmitrevs:
I attempted to merge the ideas of #469 into this PR while still keeping the cnn_mnist tests from failing. Let me know what you think.

jmitrevs:
The logic here needs to change. Most likely this will be closed and replaced with a new PR.

PR #475, opened 2021-12-18
Support UpSampling1D

vloncar:
Some L1 models need `UpSampling1D` layer. The implementation relies on existing HLS, only some wiring from converter needed to be added. The included test also tests `UpSampling2D` which fails for `io_stream` due to a bug. @nicologhielmetti is working on it.


sznajder:
UpSampling1D implementation for GCN

jmitrevs (approved):

PR #481, opened 2021-12-26
skip BatchNorm fusion when input/output is used multiple times

### LINKED ISSUE #436: Dense Layer with Linear activation causes clone stream later not working
So I tried a simple model with a Dense layer with BatchNorm afterward that has multiple outputs. 
I know the model doesn't make sense, I just want to try out the clone stream function.

image

conca_model

So I print out the model's streams variables
image

Usually, it should create clone stream variables, but instead, the  2 BatchNorm layers after the dense layer are skipped.
Therefore comes out this error:
image

The problem is solved after I set the activation function of the Dense layer to RELU.

Any suggestion?

jmduarte:
Skip BatchNorm fusion when input/output is used multiple times

Partially resolves #436

### DIFF
--- a/hls4ml/backends/fpga/passes/clone.py
+++ b/hls4ml/backends/fpga/passes/clone.py
@@ -55,13 +55,7 @@ class CloneOutput(OptimizerPass):
         if model.config.get_config_value('IOType') != 'io_stream':
             return False
 
-        output_map = {}
-        for output in node.outputs:
-            output_map[output] = []
-            for layer in model.get_layers():
-                for inp in layer.inputs:
-                    if output == inp:
-                        output_map[output].append(layer)
+        output_map = node.get_output_use_map()
 
         transformed = False
         for output in node.outputs:
--- a/hls4ml/model/layers.py
+++ b/hls4ml/model/layers.py
@@ -131,10 +131,28 @@ class Layer(object):
         else:
             return self.model.get_layer_output_variable(self.inputs[0])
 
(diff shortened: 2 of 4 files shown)


hls4ml/model/optimizer/optimizer.py
```diff
@@ -44,3 +44,13 @@ def optimize_model(model, passes=None):
...
+
+def node_output_use_map(model, node):
+    output_map = {}
+    for output in node.outputs:
+        output_map[output] = []
+        for layer in model.get_layers():
+            for inp in layer.inputs:
+                if output == inp:
+                    output_map[output].append(layer)
+    return output_map
```
thesps, line 56: Maybe this should be a member function of `Layer` in `hls_layers.py` ? And at the same time, it would make sense to tidy up `Layer.get_output_nodes` to loop over outputs instead of picking just the first to make it consistent with this.

jmduarte: @thesps this is done in a4870f8d3e3ae81a01562141b84e7f91a4c9963d

jmduarte:
@thesps @vloncar i think this is ready to be merged

vloncar (approved):

vloncar:
This looks good to me. The main concern I had was if this breaks the qkeras optimizers, but it seems like it doesn't. The key enhancement (getting the output nodes that don't consider just the first input) will be very useful.

PR #492, opened 2022-01-19
Ingest qonnx scales

jmitrevs:
This is work to propagate scales. `execute_onnx` doesn't produce identical results when running on `CNV_2W2A.onnx`, which is why the downward scale propagation is not enabled, and why this is a draft pull request at the moment.


jmitrevs:
Since the criteria for merging into the ingest-onnx branch is lower than merging into master, I removed the draft criteria, and I propose that this be merged pretty quickly. The more major reviews will go when merging that into master.

One todo is to move the accumulation inferences out of the onnx-specific optimizers, but that is not specific to getting the propagation of scales working (and the related convolution developments), so I don't think that's required for this merge.

thesps:
I agree with the above, since this is working and the target is a development branch, I'll merge it. The type of thing I would like to check up more on later is whether the "flow" of optimizer passes disrupts anything else, and whether they're somehow shared across related things. (e.g. scale factors exist in QKeras, is that handled with the same passes already or are they somehow QONNX specific)

PR #499, opened 2022-02-22
Register `ApplyAlpha` layer templates

thesps:
As first noticed by @jmitrevs in #490, since the merge of #395 the `ApplyAlpha` layer is no longer written out. The issue was that the templates weren't registered, so I've added that here. I cherry picked the commit adding the test that @jmitrevs added to expose the problem. It fails on current master, passes on the PR.

I started a `quantized_templates.py` module anticipating that we may add more specialized functions / layers for handling scale factors etc for the QONNX work later, so their template registration might go there.


jmitrevs:
I was wondering, should there be a similar setup under the quartus backend?

thesps:
I think eventually, but at the moment the Quartus HLS code doesn't support all of the specialisations (like exponential, for example), so it wouldn't give correct results. But we could look later at synchronising that stuff and moving passes (also qonnx ones probably) into the path of both Vivado and Quartus.

vloncar (approved):

PR #508, opened 2022-03-15
VivadoAccelerator backend updates

thesps:
A few small things:
- Fix registration of `VivadoAcceleratorBackend` writer flow. In master branch, the writer flow for that backend is just `VivadoBackend`'s writer flow relabelled, so the extra files for `VivadoAcceleratorBackend` aren't written out.
- Replace `VivadoAcceleratorBackend.make_bitfile` method with `build(bitfile=True)`. This is an API change to make that workflow more consistent with the usual `build` API (rather than requiring one call to `build` then one call to `make_bitfile`)
- Add a `package` method to make a .tar.gz archive with all the necesary files for hardware inference - otherwise the user needs to do some finding and renaming of files themselves

### DIFF
--- a/hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
+++ b/hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
@@ -1,8 +1,4 @@
 import os
-import shutil
-import tarfile
-import time
-import numpy as np
 
 from hls4ml.backends import VivadoBackend
 from hls4ml.model.flow import register_flow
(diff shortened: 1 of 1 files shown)


hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
```diff
@@ -56,11 +96,17 @@ def create_initial_config(self, board='pynq-z2', part=None, clock_period=5, io_t
...
+        vivado_writer = ['vivado:write']
+        vivado_accel_writer = ['vivadoaccelerator:write_hls']
+        self._writer_flow = register_flow('write', vivado_accel_writer, requires=vivado_writer, backend=self.name)
+        self._default_flow = 'vivado:ip'
+
+def _make_tarfile(output_filename, source_dir):
+    with tarfile.open(output_filename, "w:gz") as tar:
+        tar.add(source_dir, arcname=os.path.basename(source_dir))
+
+def _copy_wait_retry(src, dst, sleep=60):
```
vloncar, line 108: Can you explain why this hack is necessary? Or, in other words, why isn't the build process blocking so that the `package` can start only when we're sure the files are there?

thesps:
I got rid of the `package` function for now. I can't replicate the bitfile writing issue for the moment, it seems to depend on targeting the larger FPGA which takes a long time to build and so getting to the bottom of it will take longer. I intend to add it back as a follow up.

So now the PR does:
- Fix registration of `VivadoAcceleratorBackend` writer flow. In master branch, the writer flow for that backend is just `VivadoBackend`'s writer flow relabelled, so the extra files for `VivadoAcceleratorBackend` aren't written out.
- Replace `VivadoAcceleratorBackend.make_bitfile` method with `build(bitfile=True)`. This is an API change to make that workflow more consistent with the usual `build` API (rather than requiring one call to `build` then one call to `make_bitfile`)

nicologhielmetti (approved):

PR #511, opened 2022-03-24
Add QKeras as a package dependency

vloncar:
So far we have avoided requiring QKeras installation and tried to only load it if available. The reasons for this are long gone now, and much of the advanced functionality is not available without QKeras, so we should make it a hard dependency. Since it will be installed by pip alongside hls4ml, the impact on users should be minimal. As for "why now?", well, GitHub changed the checkout policy for unauthenticated git access which broke Jenkins tests. While the fix for that particular issue is trivial it reveals that the current code-base simply doesn't work without QKeras, so instead of maintaining this feature of QKeras being optional, we should make sure it is installed in the first place.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
   agent {
     docker {
-      image 'vivado-el7:1'
+      image 'vivado-el7:2'
       args  '-v /data/Xilinx:/data/Xilinx'
     }
   }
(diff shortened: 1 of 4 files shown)


vloncar:
This required a little bit more elbow grease. QKeras requires Python 3.7, so does hls4ml from a while back, so I added this to the package config and also updated Jenkins and the image it uses at Fermilab to use the environment based on Python 3.7.

thesps (approved):

PR #449, opened 2021-11-10
Embedding layer

jmduarte:
Original implementation from @drewpersigehl: https://github.com/drewpersigehl/L1MET_EmbeddingLayer

Addresses #422

Simple test script: https://gist.github.com/jmduarte/302c1d629747d86b926fa13b485a1682

- [x] `io_parallel` implementation
- [x] `io_stream` implementation
- [x] add tests

A few peculiarities to note (not yet enforced):
* `weight_t` should always be equal to `res_t`
* `data_t` should always be `[u]int` or `ap_[u]int<X>`

Example HLS latency/resources for `n_in = 100`, `vocab_size = 13`, `n_out = 8`, `ap_uint<4>` for inputs, and `ap_fixed<16,6>` for outputs.
* `io_parallel`:
```
    +---------+---------+----------+----------+-----+-----+----------+
    |  Latency (cycles) |  Latency (absolute) |  Interval | Pipeline |
    |   min   |   max   |    min   |    max   | min | max |   Type   |
    +---------+---------+----------+----------+-----+-----+----------+
    |        1|        1| 5.000 ns | 5.000 ns |    1|    1| function |
    +---------+---------+----------+----------+-----+-----+----------+
+---------------------+---------+-------+---------+---------+-----+
|         Name        | BRAM_18K| DSP48E|    FF   |   LUT   | URAM|
+---------------------+---------+-------+---------+---------+-----+
|DSP                  |        -|      -|        -|        -|    -|
|Expression           |        -|      -|        0|        6|    -|
|FIFO                 |        -|      -|        -|        -|    -|
|Instance             |        0|      -|     5602|     4918|    -|
|Memory               |        -|      -|        -|        -|    -|
|Multiplexer          |        -|      -|        -|       45|    -|
|Register             |        -|      -|      403|        -|    -|
+---------------------+---------+-------+---------+---------+-----+
|Total                |        0|      0|     6005|     4969|    0|
+---------------------+---------+-------+---------+---------+-----+
|Available SLR        |     1440|   2280|   788160|   394080|  320|
+---------------------+---------+-------+---------+---------+-----+
|Utilization SLR (%)  |        0|      0|    ~0   |        1|    0|
+---------------------+---------+-------+---------+---------+-----+
|Available            |     4320|   6840|  2364480|  1182240|  960|
+---------------------+---------+-------+---------+---------+-----+
|Utilization (%)      |        0|      0|    ~0   |    ~0   |    0|
+---------------------+---------+-------+---------+---------+-----+
[...]

PR #521, opened 2022-04-04
RNN support (part 1)

vloncar:
As discussed before, this is the first stage of the RNN support. The PR includes the support for parsing RNN layers into the IR, as well as the Vivado-specific initialization. The same initialization specific to Quartus backend can be part of the second PR. After this is merged, parts 2 and 3 will add HLS implementations for Vivado and Quartus.

### DIFF
--- a/hls4ml/backends/fpga/fpga_backend.py
+++ b/hls4ml/backends/fpga/fpga_backend.py
@@ -59,28 +59,48 @@ class FPGABackend(Backend):
     def get_writer_flow(self):
         raise NotImplementedError
 
-    def get_valid_reuse_factors(self, layer):
-        n_in = 0
-        n_out = 0
+    def get_layer_mult_size(self, layer):
         if 'Dense' in layer.class_name:
             n_in = layer.get_attr('n_in')
             n_out = layer.get_attr('n_out')
-        elif 'Conv1D' in layer.class_name:
+            return n_in, n_out
+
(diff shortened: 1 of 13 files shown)


vloncar:
Now also including Embedding layer from #449. @drankincms @jmduarte Can you take a look so we can merge this soon? I'd like to finish the port of Vivado RNN branch (aka part 2).

drankincms:
So right now this crashes on a GRU I tried with because `get_valid_reuse_factors()` returns an empty list for the recurrent layers.
In the keras RNN branch I have something like this: https://github.com/drankincms/hls4ml/blob/keras-RNN-staticswitch/hls4ml/templates/vivado_template.py#L478-L501
There might be a better way to pass around `recr` for this purpose, but I want to make sure @vloncar you see this issue as something that this part 1  PR should solve first before I add a fix.

jmduarte (commented):
@vloncar looks good from my side! I checked and the embedding layer works and synthesizes as expected in my models. We may want to address @drankincms's issue though.

vloncar:
I added the RF fix from the ongoing port of the RNN branch. There will be more changes in PR2, namely the transposing of weights is moved to the optimizer, in line with the other layers, which in this PR is done in the initializer of Vivado RNN layers.

drankincms (approved):
All clear for me

PR #525, opened 2022-04-12
Update inplace variables to not lose connection when source node is optimized.

jmitrevs:
In the original implementation, inplace variables are created when initializing an object, which can run into problems if the source node is changed due to optimizers. This PR moves the linking to a later stage, after the optimizations have run.

This is a replacement for #472, making use of the new features in the master branch.


jmitrevs:
I think I need to copy the type from the `input_var` in the `InplaceTensorVariable.__init__`. Let me test it a bit more extensively, though.

jmitrevs:
Some quartus-specific updates are in the QONNX PR #591. Should they be cherry-picked and added here, or is it fine to have them as part of that?

PR #527, opened 2022-04-21
Optimized look-up table

nemerchiedde:
This pull request adds the new optimized LUT, as presented in https://indico.cern.ch/event/1145909/

### DIFF
--- a/hls4ml/writer/quartus_writer.py
+++ b/hls4ml/writer/quartus_writer.py
@@ -4,7 +4,6 @@ import yaml
 from shutil import copyfile, copytree, rmtree
 import numpy as np
 import os
-import re
 import glob
 from collections import OrderedDict
 
@@ -234,7 +233,6 @@ class QuartusWriter(Writer):
             for weights in layer.get_weights():
                 self.print_array_to_cpp(weights, layer, model.config.get_output_dir())
 
-
     def write_test_bench(self, model):
         ###################
         ## test bench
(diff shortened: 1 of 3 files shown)


hls4ml/writer/quartus_writer.py
```diff
@@ -509,18 +509,21 @@ def write_activation_tables(self, model):
...
         ###################
-        ## sigmoid_table
+        # sigmoid_table
         ###################
+        CENTERED = True
+        MAX_VALUE = 8
+        MIN_VALUE = 0
         for layer in model.get_layers():
-            if(layer.get_attr('activation') == 'sigmoid'):
+            if layer.get_attr('activation') == 'sigmoid' and layer.get_attr('table_size') is not None:
```
vloncar, line 518: Why not just `table_size = layer.get_attr('table_size', 1024)`?

hls4ml/writer/quartus_writer.py
```diff
@@ -534,28 +537,35 @@ def write_activation_tables(self, model):
...
         ###################
-        ## tanh_table
+        # tanh_table
         ###################
+        CENTERED = True
+        MAX_VALUE = 4
+        MIN_VALUE = 0
         for layer in model.get_layers():
-            if(layer.get_attr('activation') == 'dense_tanh'):
+            if layer.get_attr('activation') == 'dense_tanh' and layer.get_attr('table_size') is not None:
```
vloncar, line 560: Same as above.

hls4ml/writer/quartus_writer.py
```diff
@@ -509,18 +509,21 @@ def write_activation_tables(self, model):
         h_file.close()
 
         ###################
-        ## sigmoid_table
+        # sigmoid_table
         ###################
+        CENTERED = True
```
vloncar, line 514: Can you remind me if there is any reason to keep the non-centered versions? There seems to be no advantage to that. If so, the non-centered code path should be removed.

vloncar (changes requested):
Thanks Nemer, this is really nice. Can you address the comments below?
Also, can you switch to using 4 spaces as indentation in the HLS code? Right now, you mix 2 and 4 spaces. It would help with readibility

hls4ml/writer/quartus_writer.py
nemerchiedde: I had left it in case someone would like to use or test the non-centralized functionality, but particularly I don't see any advantages in using it. I will remove it so.

hls4ml/writer/quartus_writer.py
nemerchiedde: This line allows me to set a different table_size for each activation function, otherwise I would have the same value for both activation functions.

hls4ml/writer/quartus_writer.py
nemerchiedde: Same reason as above

vloncar (approved):

PR #528, opened 2022-04-21
Extension API

vloncar:
One of the pitfalls of using hls4ml with new applications is that there is little support for extending hls4ml with custom objects (layers, hls code, optimizers etc). This forces people to create forks and make changes there. These forks often include useful bugfixes and improvements that never go back into the main branch. As a result, we often fix bugs multiple times across different branches. If the core is extendable, there is no need for forks, and bugfixes can be applied once, on the main branch, and we wouldn't waste effort.

This PR extends the core with a few missing features and documents the extension points in depth. It is not the end of extension API, just the beginning, but it should be enough so that most of the new applications targeting existing features don't need to fork.

The included test case demonstrates a flow of extending hls4ml to support a custom Keras layer.

### DIFF
--- a/hls4ml/backends/backend.py
+++ b/hls4ml/backends/backend.py
@@ -43,21 +43,58 @@ class Backend(object):
         return [name for name in get_backend_passes(self.name) if isinstance(get_optimizer(name), Template)]
 
     def create_initial_config(self, **kwargs):
+        """Create the minimal conversion config for the backend.
+
+        Subclasses should implement this method to provide the initial configuration for the conversion.
+        """
         raise NotImplementedError
 
     def create_layer_class(self, layer_class):
+        """Wrap the original layer class into the backend-specific layer class.
+
+        Backends should extend base layer classes with new attributes and variables as needed. These new classes are then
+        used within the model.
+
+        Args:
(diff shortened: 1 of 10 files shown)


thesps:
Nice! Does this already support having "custom configuration" for these custom layers? i.e. some config parameters with names not in the standard ones like "n_in", "n_out" etc? It looks like it may just work, but thought I'd check.

vloncar:
No, it won't just work ATM unfortunately. It was in a way a conscious omission :smile:. We need an overhaul of the config and I'd rather not add further hacks now. I'm planning a proper fix for June. The `config_from...` function could go over the layers and get a list of configurable attributes (similarly to how the validation works) and build a base config from that. I have code that implements this concept, but it is just not elegant with the current way of dealing with the config (model is essentially parsed twice, first time just to get the config). So I don't want to replace one ugly hack with another.

thesps:
Returning to this... I wasn't referring to the config generation utility, but the reading of the config. I think it's safe to assume that if the user can add a custom layer then they can write the config for it by hand. Since it's been a while, how about pulling master branch and then we merge?

thesps (approved):

Issue #497, opened 2022-02-17
BatchNormalization scale and bias array size


jmitrevs:
The `normalize` function call in both `nnet_batchnorm.h` and `nnet_batchnorm_stream.h` has an interface of:

```c++
template<class data_T, class res_T, typename CONFIG_T>
void normalize(
    data_T    data[CONFIG_T::n_in],
    res_T     res[CONFIG_T::n_in],
    typename CONFIG_T::scale_t  scale[CONFIG_T::n_in],
    typename CONFIG_T::bias_t   bias[CONFIG_T::n_in]
)
```
However, in the case when `CONFIG_T::n_filt == -1`, only the first `CONFIG_T::n_filt` of the arrays is used. This value is usually much smaller. The code should be updated to reflect this (potentially by me).

Two options that I can think of are to either introduce another constant, say `CONFIG::n_scale_bias`, that is set appropriately to the size of the scale and bias, or to have separate `normailze` variants depending on the value of `CONFIG_T::n_filt`.

vloncar:
`CONFIG::n_scale_bias` approach seems far simpler, but are we sure we won't have to split the functions anyway for optimal QoR?

jmitrevs:
This was merged on the Vivado side but not on the Quartus side. Leaving it open till it's fixed there.

jmitrevs:
This was fixed in #749 for Quartus, so closing.

PR #532, opened 2022-04-28
Make the size of bn scale and bias what they really are

jmitrevs:
As discussed in issue #497, the normalize scale and bias are declared too large in the case when `n_filt != -1`, which sometimes affects syntesizability.  This PR declares them to be the needed size.

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_batchnorm.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_batchnorm.h
@@ -36,6 +36,7 @@ struct batchnorm_config
     // Layer Sizes
     static const unsigned n_in = 10;
     static const unsigned n_filt = -1;
+    static const unsigned n_scale_bias = 10;
     
     // Resource reuse info
     static const unsigned io_type = io_parallel;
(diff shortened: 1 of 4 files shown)


hls4ml/templates/vivado/nnet_utils/nnet_batchnorm.h
```diff
@@ -36,6 +36,7 @@ struct batchnorm_config
     // Layer Sizes
     static const unsigned n_in = 10;
     static const unsigned n_filt = -1;
+    static const unsigned n_scale_bias = 10;
```
vloncar, line 39: Can't you just put `n_scale_bias = (n_filt == -1) ? n_in : n_filt;` here and then you don't need changes to the python templates?

jmitrevs: That might be too early to do it, since n_in and n_filt are rewritten. I don't think a reference would work, either, since by basically they get hidden by the inheritance. What might work is if line 60 in core_templates changed to have the ? : there. Then 'n_scale_bias' would never have to be on the python parameters.

jmitrevs: I just pushed such a change.

vloncar (approved):

Issue #536, opened 2022-05-02
Issue with hlsmodel.predict


JochiSt:
When I try to use ` hls_model.predict( np.array(x) )`

I get the error:
```
  File "/hls4ml/hls4ml/model/graph.py", line 645, in predict
    for i in range(n_samples):
TypeError: 'numpy.float64' object cannot be interpreted as an integer
```
I fixed this by changing the return of  `_compute_n_samples(self, x):` to `return int(n_sample)`

The versions I used:
```
Python 3.8.10 (default, Mar 15 2022, 12:22:08)
[GCC 9.4.0]
Numpy 1.22.3
TensorFlow 2.8.0
Keras 2.8.0
```
For HLS4ML I used the latest git version.

PR #537, opened 2022-05-02
returning integer from _compute_n_samples

JochiSt:
I changed the return type of `_compute_n_samples` to `int`. 

See issue #536 for the error message I get otherwise.


jmduarte (approved):
LGTM

PR #538, opened 2022-05-04
Ingest qonnx jettagging

jmitrevs:
Allow to parse QKeras model by treating power of 2 scales as special cases.


jmitrevs:
The bias fusing doesn't seem to be happening. Let me fix that.

jmitrevs:
The fusing actually affects the precision, so let me take this opportunity to move the precision calculation to a separate optimization. I therefore added WIP to this PR

PR #543, opened 2022-05-06
use `unsigned int` instead of `unsigned short` for number of inputs and outputs

jmduarte:
- to allow models that have more than 65535 inputs or outputs

@nickshey


jmduarte:
closed by #559

PR #547, opened 2022-05-11
Enable SeparableConv2d

jmduarte:
This PR enables SeparableConv2D and adds a simple test, however, it'd be useful to discuss further what features we want to support/add (currently it seems the implementation is bit brittle)

@nickshey


vloncar:
I'm fine with this as is, but one thing I noticed is that if we use kernel_size=stride the padding layer still gets inserted, but does nothing. This could be optimized away (either as part of this PR, or a futue one).

jmduarte:
@vloncar added the requested change in fb6004f

vloncar (approved):
Looks good to me

PR #548, opened 2022-05-13
Quartus Embedding Layer

bo3z:
Added support for `Embbeding` layer in Quartus, similar to work in this commit: https://github.com/fastmachinelearning/hls4ml/pull/521/commits/5040f75c095a92559f124e092f145849e0c815ca#diff-7dd4e5f5df1fd1a2cf2974ce0982818da375ff46be0eda702ba60b4f6188d928

Will be useful once work on `SimpleRNN` and `LSTM` is completed.

### DIFF
--- a/hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed.h
+++ b/hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed.h
@@ -1,0 +1,47 @@
+#ifndef NNET_EMBED_H_
+#define NNET_EMBED_H_
+
+#include "nnet_common.h"
+#include "nnet_helpers.h"
+
+namespace nnet {
+
+    struct embed_config {
+        // Internal data type definitions
+        typedef float embeddings_t;
+
+        // (Default layer sizes, overwritten form the backend
+        static const unsigned n_in = 10;
+        static const unsigned n_out = 16;
+        static const unsigned vocab_size = 50;
+
+        // Resource reuse info
+        static const unsigned io_type = io_parallel;
+        static const unsigned reuse_factor = 1;
+    };
+
+    template<class data_T, class res_T, typename CONFIG_T>
+    void embedding(
+        data_T data[CONFIG_T::n_in],
+        res_T  res[CONFIG_T::n_in * CONFIG_T::n_out],
+        const typename CONFIG_T::embeddings_t embeddings[CONFIG_T::vocab_size * CONFIG_T::n_out]) {
+
+        /*
+        * Can store embeddings[] in a register, but a large multiiplexer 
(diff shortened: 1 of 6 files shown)


hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed_stream.h
```diff
@@ -0,0 +1,10 @@
+/*
+*   PLACEHOLDER - The common pass embedding.py includes both parallel and streaming implementations; streaming is currently not supported in Quartus
+*/
+
+#ifndef NNET_BATCHNORM_STREAM_H_
+#define NNET_BATCHNORM_STREAM_H_
```
vloncar, line 6: This can trip you up :wink:

hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed_stream.h
```diff
@@ -0,0 +1,10 @@
...
+
+namespace nnet {}
```
vloncar, line 8: Can we add a stub that raises an exception while we still don't have the implementation? Otherwise the compilation fails with a cryptic message (cannot substitute template argument or something like that). Or does that require the more streaming infrastructure in Quartus (#557)?

hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed.h
```diff
@@ -0,0 +1,47 @@
...
+        */
+       
+        InputSequence:
+        #pragma ii CONFIG_T::reuse_factor 
+        #pragma unroll
+        for (int j = 0; j < CONFIG_T::n_in; j++) {
+            DenseEmbedding: 
+            #pragma unroll
+            for (int i = 0; i < CONFIG_T::n_out; i++) {
+                res[j * CONFIG_T::n_out + i] = embeddings[data[j].to_uint() * CONFIG_T::n_out + i];
```
vloncar, line 41: Remind me, why is the `to_uint()` needed here?

vloncar (changes requested):

hls4ml/templates/quartus/firmware/nnet_utils/nnet_embed_stream.h
bo3z: This file should never be invoked, as soon as Quartus attempts to use io_stream an exception is thrown from Python (although https://github.com/fastmachinelearning/hls4ml/pull/557 attempts to add infrastructure for streaming, which means for some layers we should throw exceptions and for some not - this should be handled by Python and can be a part of the streaming PR)

vloncar (approved):

PR #551, opened 2022-05-17
fix insertion before a node with multiple inputs + support additional broadcasting

jmduarte:
This PR has a few related fixes
- fix insertion before a node with multiple inputs
- correctly insert broadcast stream before a node with multiple inputs (e.g. Add or Concatenate), e.g. see figure below. The current code would behave incorrectly and replace the wrong input (always the first input) for the following layer... this would result in an additional output incorrectly being inferred
- don't insert broadcast before concatenate (figure below incorrectly does this)

Screen Shot 2022-05-16 at 6 59 26 PM

To do
- [x] Add a pytest

@nickshey

### DIFF
--- a/hls4ml/backends/vivado/passes/repack_stream.py
+++ b/hls4ml/backends/vivado/passes/repack_stream.py
@@ -44,8 +44,10 @@ broadcast_function_template = 'nnet::broadcast_stream<{input_t}, {output_t}, {co
 broadcast_config_template = """struct config{index} : nnet::broadcast_config {{
     static const unsigned in_width = {in_width};
     static const unsigned in_height = {in_height};
-    static const unsigned n_chan = {n_chan};
-    static const unsigned n_dupl = {n_dupl};
+    static const unsigned in_chan = {in_chan};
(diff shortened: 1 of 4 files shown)


thesps:
Pushed to local branch pr/551 to run Gitlab CI, looks like the graph tests still pass. An addition to `test_graph.py` that touches this new `input_idx` argument would be good.

jmduarte:
@thesps I added the test you requested. It was failing though because I realized the only stream broadcasting we currently supported as of the form: `(1, 1, C) -> (H, W, C)`, and I was testing a different form, namely `(H, W, 1) -> (H, W, C)`.

So I updated our HLS code to support both of these (easy) cases. I also added tests for both cases, and added a `RuntimeError` if a user tries to use a model with broadcasting that is not one of these two cases.


In principle, it should be straightforward(ish) to also support `(1, W, C) -> (H, W, C)` and `(H, 1, C) -> (H, W, C)`, but this can be a future PR

jmduarte:
@thesps I think this is ready to be merged unless you have any other comments!

PR #552, opened 2022-05-18
Providing support for Alveo boards

selwyn96:
This PR provides backend support for Alveo boards (U50, U200, U250 and U280) by leveraging the VivadoAccelerator backend. 
Reference: https://github.com/fastmachinelearning/hls4ml/issues/496.

## Changes:
In `hls4ml/templates/vivado_accelerator/alveo`
Added RTL files which act as AXI stream to AXI master adapter for the Alveo boards.
The axi-stream tcl scripts for all boards.
The python drivers.

In` /hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py`
Added a make_xclbin function that generates the xclbin container by linking the project and object files.

## Tests
Modified the `test/hls4ml-keras-test.sh`  script to test the functionality of the branch. 
A more comprehensive tutorial for training and deploying on the Alveo boards can be found at  https://github.com/selwyn96/Alveo-tutorial

## Slides
(https://docs.google.com/presentation/d/1D9tryRyFp6kSb2ytc8kBLJ-El4CB_giDGnOQ2lFuLzs/edit?usp=sharing)


thesps:
Thanks a lot for this! And thanks @Gabriele-bot for getting it rolling. I did a build today with the QKeras jet tagging example model targeting my U200, and it worked, which is awesome!
I have one general question: did you try to create an AXI master interface in HLS and skip the master-to-stream RTL? It's great that it works as is, but it would still be nice if we could simplify things.

I'll leave some more detailed review comments inline in the code.

thesps (changes requested):
`axi_stream_design.tcl`: now there are 4 copies of basically the same file, differing only by the device part (as far as I can tell). Can we reduce that to one copy with a paremeter? It will be better for maintenance. We could write the part into the project.tcl, for example. For the Python driver there's already just a single file.

hls4ml/templates/vivado_accelerator/alveo/python_drivers/axi_stream_driver.py
```diff
@@ -0,0 +1,109 @@
+from datetime import datetime
+
+import numpy as np
+from pynq import Overlay
+from pynq import allocate
+
+
+class NeuralNetworkOverlay(Overlay):
```
thesps, line 8: I think we can hide a bit of the buffer handling from the user. For example, make `input_buffer` and `output_buffer` members of the class so the user doesn't have to pass them around between function calls. And `predict` could call `allocate_mem` itself.
It would be nice if the user code could be as simple as:
```
nn = NeuralNetworkOverlay('my.cxlbin')
y = nn.predict(X, y_shape)
```

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -311,11 +311,15 @@ def write_wrapper_test(self, model):
...
-        Write the tcl scripts to create a Vivado IPI project for the VivadoAccelerator
+        Write the tcl scripts and kernel sources to create a Vivado IPI project for the VivadoAccelerator
         '''
         filedir = os.path.dirname(os.path.abspath(__file__))
         copyfile(os.path.join(filedir, self.vivado_accelerator_config.get_tcl_file_path()),
                  '{}/design.tcl'.format(model.config.get_output_dir()))
+        # Generic alveo board
+        if self.vivado_accelerator_config.get_board().startswith('alveo'):
+            copytree(os.path.join(filedir, self.vivado_accelerator_config.get_krnl_rtl_src_dir()),
+                 '{}/src'.format(model.config.get_output_dir()), dirs_exist_ok=True)
```
thesps, line 322: This `dirs_exist_ok` option was introduced in Python 3.8, but hls4ml has the minimum Python version 3.7 ([here](https://github.com/fastmachinelearning/hls4ml/blob/master/setup.cfg#L5)). It would be okay to increase that to 3.8.

scripts/hls4ml
```diff
@@ -140,7 +140,7 @@ def _build(args, extra_args):
         print('Project directory (-p or --project) must be provided.')
         sys.exit(1)
 
-    try:
+    try:
```
thesps, line 143: Can you revert this so we don't add a whitespace only change?

test/keras-models.txt
```diff
@@ -29,9 +29,10 @@ KERAS_3layer_batch_norm
 KERAS_3layer_binary_smaller
 KERAS_3layer_ternary_small
 
-KERAS_3layer b:pynq-z2 B:VivadoAccelerator x:xc7z020clg400-1 s:Resource
+# Pynq backend
+-b pynq-z2 -B VivadoAccelerator -x xc7z020clg400-1 KERAS_3layer
```
thesps, line 33: I think the syntax is wrong here

test/keras-models.txt
```diff
@@ -29,9 +29,10 @@ KERAS_3layer_batch_norm
...
 
-garnet_1layer x:xcku115-flvb2104-2-i y:garnet_1layer_config 
+-x xcku115-flvb2104-2-i -y garnet_1layer_config garnet_1layer
```
thesps, line 35: and here

thesps (changes requested):

selwyn96:
@thesps Thanks a lot for the detailed suggestions. 
Regarding your general question: I have not been able to make much progress in moving the AXI master interface to hls. I'm still looking into it though and will make an update if there's any progress.

hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
```diff
@@ -24,6 +24,34 @@ def build(self, model, reset=False, csim=True, synth=True, cosim=False, validati
 
         return parse_vivado_report(model.config.get_output_dir())
 
+    def make_xclbin(model, platform='xilinx_u250_gen3x16_xdma_3_1_202020_1'):
```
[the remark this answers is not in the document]
thesps: I think `bitfile=True` is fine. For sure it should be the same argument for either the Zynq or Alveo flow, basically an argument that means "do all the synthesis, place and route, and produce the output products". Do you have an idea for a different name that captures that well for both?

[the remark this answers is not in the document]
selwyn96: If we want to keep the same argument for both then as you said `bitfile` should be fine.  I don't think there is a general word that encapsulates the meaning of both, I'll stick to  `bitfile` for now.

hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
```diff
@@ -24,6 +24,34 @@ def build(self, model, reset=False, csim=True, synth=True, cosim=False, validati
...
+        os.chdir(model.config.get_output_dir())
+        os.makedirs('xo_files', exist_ok=True)
+        try:
+            os.system('vivado -mode batch -source design.tcl')
+        except:
+            print("Something went wrong, check the Vivado logs")
+        # TODO improve the line below
+        ip_repo_path = model.config.get_output_dir() + '/myproject_prj/solution1/impl/ip'
+        os.makedirs('xclbin_files', exist_ok=True)
+        os.chdir(model.config.get_output_dir() + '/xclbin_files')
```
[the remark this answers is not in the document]
selwyn96: It should work if  `model.config.get_output_dir()` is set to the absolute path. Can you tell me what you had to do to get it to work?

hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py
```diff
@@ -52,6 +88,7 @@ def create_initial_config(self, board='pynq-z2', part=None, clock_period=5, io_t
         config['AcceleratorConfig']['Board'] = board
         config['AcceleratorConfig']['Interface'] = interface  # axi_stream, axi_master, axi_lite
         config['AcceleratorConfig']['Driver'] = driver
+        config['AcceleratorConfig']['Platform'] = platform
```
thesps, line 91: Could we only set this if the `board` requires it (i.e. is Alveo)?
Otherwise it could return something confusing like:
`{'Board' : 'pynq-z2',
   'Platform' : 'xilinx_u250_xdma_201830_2'}

thesps (changes requested):

thesps (approved):
Thanks for the fixes, looks very nice now

Issue #248, opened 2020-11-24
Specifying qkeras activation by keyword parameter


veyron8800:
Hello,

Small convenience issue here. When I specify a quantized activation for a qkeras layer through a keyword parameter, hls4ml throws an error during model conversion. This can be easily overcome by specifying no activation function within the layer and a separate QActivation following it. However, this may be confusing for those working with qkeras/hls4ml for the first time.

I receive the following error from hls4ml when specifying quantized_relu as a keyword parameter:
```
Traceback (most recent call last):
  File "/snap/pycharm-community/214/plugins/python-ce/helpers/pydev/pydevd.py", line 1448, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/snap/pycharm-community/214/plugins/python-ce/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/areis/Desktop/HLS4ML_Master/q_model/convert.py", line 7, in <module>
    hls4ml.converters.keras_to_hls(cfg)
  File "/home/areis/Desktop/HLS4ML_Master/venv/lib/python3.8/site-packages/hls4ml/converters/keras_to_hls.py", line 243, in keras_to_hls
    act_layer['name'] = layer['name'] + '_' + layer['activation']
TypeError: can only concatenate str (not "dict") to str
```

And it is because `layer['activation']` is indeed a dictionary and not a string. In my case it contains the following data:
```
{
    "class_name": "quantized_relu",
    "config": {
        "bits": 8,
        "integer": 4,
        "negative_slope": 0.0,
        "relu_upper_bound": null,
        "use_sigmoid": 0,
        "use_stochastic_rounding": false
    }
}
```

The fix is switching
```
model.add(qkeras.QConv2D(4, (3,3), kernel_quantizer='quantized_bits(8,4)', bias_quantizer='quantized_bits(8,4)', activation='quantized_relu(8,4)'))
```
with
```
model.add(qkeras.QConv2D(4, (3,3), kernel_quantizer='quantized_bits(8,4)', bias_quantizer='quantized_bits(8,4)'))
model.add(qkeras.QActivation('quantized_relu(8,4)'))
```

[Here are the example models I was working with.](https://github.com/fastmachinelearning/hls4ml/files/5587081/ExampleQActivationModels.zip)

Thanks

PR #553, opened 2022-05-23
Fix for QActivations passed as an argument

AdrianAlan:
Fix for `QKeras` layers with `QActivations` passed as an argument.

Right now these models fail to compile with `TypeError` in the `keras_to_hls` method as the method expects the `layer['activation']` to be a string, not a dict.

### DIFF
--- a/hls4ml/converters/keras_to_hls.py
+++ b/hls4ml/converters/keras_to_hls.py
@@ -316,16 +316,32 @@ def keras_to_hls(config):
         layer_list.append( layer )
         if 'activation' in layer and layer['class_name'] not in activation_layers + recurrent_layers:# + qkeras_layers:
             act_layer = {}
-            act_layer['name'] = layer['name'] + '_' + layer['activation']
-            act_layer['activation'] = layer['activation']
-            if 'activ_param' in layer:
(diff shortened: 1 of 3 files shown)


thesps:
Thanks for this! Could you adapt the test to do some check on the output more than just compilation? I would suggest testing this functionality in existing `test_keras_api.py`, following the pattern of constructing a model, compiling, running prediction with some random inputs, checking the output Keras vs hls4ml, and doing some explicit checking that the converted model makes sense (ie here that the activation is what you expect).

thesps:
Thanks for this! Could you adapt the test to do some check on the output more than just compilation? I would suggest testing this functionality in existing `test_keras_api.py`, following the pattern of constructing a model, compiling, running prediction with some random inputs, checking the output Keras vs hls4ml, and doing some explicit checking that the converted model makes sense (ie here that the activation is what you expect).

AdrianAlan:
I moved it to `test_qkeras.py`, I thought it may belong there, rather than `test_keras_api.py`. Let me know if you have some other `quantizers` in mind. BTW, it solves #248

thesps (approved):

PR #559, opened 2022-05-31
Remove the in/out size from top function

vloncar:
The size of input and output that is part of the top function is never used and can be removed. Since the data type used is `short` large models emit warnings. #543 suggested a fix, but after some thought, we should get rid of this feature altogether.

Basically, from  `myproject(input_t input_1, result_t result, unsigned short &const_size_in_1, unsigned short &const_size_out_1)` we do `myproject(input_t input_1, result_t result)`.


jmduarte (approved):
ran tests and locally and it looks good to me!

PR #560, opened 2022-06-02
Vivado Backend GRU/LSTM support

drankincms:
This PR adds support for GRU/LSTM layers to the Vivado backend.

### DIFF
--- a/hls4ml/backends/vivado/passes/recurrent_templates.py
+++ b/hls4ml/backends/vivado/passes/recurrent_templates.py
@@ -108,9 +108,14 @@ class LSTMConfigTemplate(LayerConfigTemplate):
 
         params['n_in'] = node.get_input_variable().dim_names[1]
         params['n_sequence'] = node.get_input_variable().dim_names[0]
-        params['n_sequence_out'] = node.get_output_variable().dim_names[0]
-        params['n_state'] = node.get_output_variable().dim_names[1]
-        params['n_out'] = node.get_output_variable().dim_names[1]
+        if node.get_attr('return_sequences'):
+            params['n_sequence_out'] = node.get_output_variable().dim_names[0]
+            params['n_state'] = node.get_output_variable().dim_names[1]
+            params['n_out'] = node.get_output_variable().dim_names[1]
+        else:
+            params['n_sequence_out'] = 1
+            params['n_state'] = node.get_output_variable().dim_names[0]
+            params['n_out'] = node.get_output_variable().dim_names[0]
         params['config_mult_t1'] = 'config{}_1'.format(node.index)
         params['config_mult_t2'] = 'config{}_2'.format(node.index)
         params['lstm_act_t'] = '{}_config{}_recr'.format(node.get_attr('recurrent_activation'), node.index)
@@ -124,9 +129,13 @@ class LSTMConfigTemplate(LayerConfigTemplate):
         recr_act_params = self._default_config_params(node)
 
         act_params['type'] = node.get_attr('activation')
-        act_params['n_in'] = node.get_output_variable().dim_names[1]
         recr_act_params['type'] = node.get_attr('recurrent_activation')
-        recr_act_params['n_in'] = node.get_output_variable().dim_names[1] + ' * 3'
+        if node.get_attr('return_sequences'):
+            act_params['n_in'] = node.get_output_variable().dim_names[1]
+            recr_act_params['n_in'] = node.get_output_variable().dim_names[1] + ' * 3'
+        else:
+            act_params['n_in'] = node.get_output_variable().dim_names[0]
+            recr_act_params['n_in'] = node.get_output_variable().dim_names[0] + ' * 3'
 
         act_config = self.act_template.format(**act_params)
         recr_act_config = self.recr_act_template.format(**recr_act_params)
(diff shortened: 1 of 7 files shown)


hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -241,7 +241,7 @@ def init_lstm(self, layer):
         reuse_factor = layer.model.config.get_reuse_factor(layer)
         layer.set_attr('recurrent_reuse_factor', reuse_factor)
 
-        recurrent_bias = np.zeros(layer.weights['recurrent_weight'].shape[1])
+        recurrent_bias = np.zeros(layer.weights['recurrent_weight'].shape[0 if layer.model.config.is_resource_strategy(layer) else 1])
```
vloncar, line 244: Why? Also handling resource strategy should happen later in the flow.

hls4ml/model/layers.py
```diff
@@ -850,6 +850,7 @@ class SimpleRNN(Layer):
         Attribute('return_sequences', value_type=bool, default=False),
         Attribute('return_state', value_type=bool, default=False),
         ChoiceAttribute('direction', ['forward', 'backward'], default='forward'),
+        Attribute('static', value_type=bool, default=True),
```
vloncar, line 853: This is an implementation detail, so it should belong to the backend. [Here](https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/backends/vivado/vivado_backend.py#L25).

hls4ml/model/layers.py
```diff
@@ -863,10 +864,9 @@ class SimpleRNN(Layer):
     def initialize(self):
         if self.attributes['return_sequences']:
             shape = [self.attributes['n_timesteps'], self.attributes['n_out']]
-            dims = ['N_TIME_STEPS_{}'.format(self.index), 'N_OUT_{}'.format(self.index)]
         else:
-            shape = [self.attributes['n_out']]
-            dims = ['N_OUT_{}'.format(self.index)]
+            shape = [1, self.attributes['n_out']]
```
vloncar, line 868: Is there any way to handle this differently? If we always return sequence(s), that means any HLS/C++ implementation will have to be done like that, which is counter-intuitive.

hls4ml/model/layers.py
```diff
@@ -917,10 +917,15 @@ def initialize(self):
             self.add_output_variable(state_shape, state_dims, out_name=self.outputs[1], var_name='layer{index}_h', type_name='layer{index}_h_t')
             self.add_output_variable(state_shape, state_dims, out_name=self.outputs[2], var_name='layer{index}_c', type_name='layer{index}_c_t')
 
-        self.add_weights()
+        data = self.model.get_weights_data(self.name, 'kernel')
+        if self.model.config.is_resource_strategy(self) and self.model.config.backend.name in ['Vivado', 'VivadoAccelerator']:
```
vloncar, line 921: This should be moved later in the flow.

hls4ml/model/optimizer/passes/multi_dense.py
```diff
@@ -5,7 +5,7 @@
 class ReplaceMultidimensionalDenseWithConv(OptimizerPass):
     def match(self, node):
         return isinstance(node, Dense) and \
-            len(node.get_input_variable().shape) > 1
+            len(node.get_input_variable().shape) - sum(d==1 for d in node.get_input_variable().shape) > 1
```
vloncar, line 8: Surely there is a more descriptive way of excluding RNN layers from this optimization... :smiley:

hls4ml/templates/vivado/nnet_utils/nnet_recr_activations.h
```diff
@@ -0,0 +1,63 @@
...
+
+template<class data_T, class res_T, typename CONFIG_T>
+class relu : public Activation<data_T, res_T, CONFIG_T>{
+    public:
+    // *************************************************
+    //       Relu Activation
+    // *************************************************
+    static void activation(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_in])
+    {
+        nnet::relu<data_T, res_T, CONFIG_T>(data, res);
```
vloncar, line 31: Out of curiosity, does Vivado inline this call automatically, or can we save a cycle here with a simple pragma?

vloncar (changes requested):
Thanks Dylan, this is looking good. Can we address the first round of comments and then we can look at the HLS in more detail?

hls4ml/backends/vivado/passes/recurrent_templates.py
```diff
@@ -0,0 +1,249 @@
...
+    static const unsigned reuse_factor = {reuse};
+    static const bool store_weights_in_bram = false;
+    static const bool use_static = {static};
+}};\n"""
+
+gru_function_template = 'nnet::gru_stack<{input_t}, {output_t}, {config}>({input}, {output}, {w}, {wr}, {b}, {br});'
+
+recr_include_list = ['nnet_utils/nnet_recurrent.h']
+
+class LSTMConfigTemplate(LayerConfigTemplate):
```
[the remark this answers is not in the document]
drankincms: How would I go about handling multiple cases for say: `super().__init__(GRU)`? Also, are you suggesting to also merge the template strings? Or just the classes?

hls4ml/model/layers.py
drankincms: Ok, I've moved it there.

hls4ml/model/optimizer/passes/multi_dense.py
drankincms: Indeed :) Although it seemed to me like a multidimensional dense layer with all axis lengths but one equal to 1 should not really be automatically transformed into a pointwise conv. I can change this to explicitly pick out the dense layers connected to LSTM/GRU if you would prefer.

hls4ml/model/layers.py
drankincms: Got it, fixed.

hls4ml/backends/vivado/passes/recurrent_templates.py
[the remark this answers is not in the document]
drankincms: @vloncar Just bumping this. It's not clear to me how I should go about merging these classes.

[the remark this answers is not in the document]
vloncar: Sorry, I missed this. Why not `super().__init__((LSTM, GRU))`? You can also merge the two template strings and handle the small differences in `format()`. Since that function takes `node` as a parameter, with `node.class_name` you can get whether it is `LSTM` or `GRU`.

vloncar (approved):

PR #562, opened 2022-06-02
remove cppname

jmitrevs:
Variable `name` and `cppname` are often used interchangeably in the code, not consistently. My understanding is that `cppname` was meant to be a cleaned up C-legal version of name. Instead of looking at each case where `name` is used and trying to see if it should be `cppname` and vice versa, it probably makes sense to just get rid of `cppname` altogether, and just use `name` everywhere.

One question is whether we want to include this name cleaning in the `Variable.__init__` function:
```python
self.name = re.sub(r'\W|^(?=\d)','_', self.name)
```
I have not included it. The standard pytests all pass without it.

### DIFF
--- a/hls4ml/backends/fpga/fpga_types.py
+++ b/hls4ml/backends/fpga/fpga_types.py
@@ -184,11 +184,11 @@ class VariableDefinition(object):
 
 class VivadoArrayVariableDefinition(VariableDefinition):
     def definition_cpp(self, name_suffix='', as_reference=False):
-        return '{type} {name}{suffix}[{shape}]'.format(type=self.type.name, name=self.cppname, suffix=name_suffix, shape=self.size_cpp())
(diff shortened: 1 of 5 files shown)


vloncar:
Cleaning the name at that point would in fact break the model if a change actually happened. `cppname` dates back to when the output variable names followed the layer names, so naming a layer "my layer" would generate invalid C++. Since we switched to using `layer{index}_out` as the name of the output variable in generated C++, this is no longer relevant so `cppname` can be removed. I forgot to do it as part of the backend restructuring. Thanks for this.

jmitrevs:
For Quartus we rename name for structure variables (the input and output). Is that safe?

jmitrevs:
I would like if possible to have this merged fairly soon since it effects new developments, which have to consider this PR.

vloncar:
This looks fine to me, I tried it on my Vivado developments and nothing breaks. But Quartus worries me a bit. I remember (ab)using `cppname` there for something. @bo3z Can you try this out and if there are no issues we merge it ASAP?

vloncar (approved):

PR #567, opened 2022-06-10
Don't override precision directly in the QKeras optimizer

vloncar:
`output_rounding_saturation_mode` optimizer directly manipulates precision object of the output variable of the node, which can trip up the later optimizers and make them skip the type conversion. It was observed in testing of accelerator backend with fifo depth optimization (by @thesps) as well as some GNN models (by @sznajder) This change ensures the update of precision also updates the entire type. Eventually we will make the `precision` a read-only property to ensure this type of error doesn't affect future optimizers, but this kind of change requires a bit more work.


thesps:
Can you push to a local branch to trigger the Gitlab CI? The convention so far has been to push to, in the case, `pr/567`, but any local name will do.

vloncar:
['tis done](https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/pipelines/4096442)

thesps (approved):

PR #569, opened 2022-06-16
Add quantized sigmoid, fix quantized tanh for QKeras

jmitrevs:
This is an attempt to fix quantized tanh and add quantized sigmoid activation support for QKeras ingestion. In Quartus the tanh handling probably needs an optimizer to rename it to dense_tanh (and that is why the tests fail--the hacky fix in the writer doesn't work), but for Vivado I think this generally works. Overall, though, I think this fix is only temporary. The method breaks if `granularity='name'` is not used in `hls4ml.utils.config_from_keras_model`, though this feature is not just for these quantizers.

Any suggestions?


test/pytest/test_qkeras.py
```diff
@@ -229,6 +229,45 @@ def test_quantizer(randX_1000_1, quantizer, backend):
...
+
+  output_dir = str(test_root_path / 'hls4mlprj_qkeras_quantizer_alt_{}_{}_{}'.format(quantizer.__class__.__name__,
+                                                            quantizer.bits, backend))
+  hls_model = hls4ml.converters.convert_from_keras_model(model,
+                                                       hls_config=config,
+                                                       output_dir=output_dir,
+                                                       backend=backend)
+  hls_model.compile()
+
+  y_calc = func(X)
```
thesps, line 267: Why compare to a floating point version of the function on not the quantized one here?
i.e. you could do `y_calc = quantizer(X)`.
And then, on a related note, is there a problem just including these quantizers in `test_quantizer`?

jmitrevs: The reason for the different test is to do the different comparison. You have to do a try catch on the integer field, but in general, even if doing `use_qkeras_tanh=True` there were occasional differences that I could not solve by setting rounding and saturation modes. I can look more.

jmitrevs:
I agree about moving the logic to an optimizer pass, but that would be for all the quantizers, not just for quantized tanh and sigmoid. Basically all QKeras info is lost if `granularity='name'` is not used, so it is a more fundamental change we should discuss.

test/pytest/test_qkeras.py
thesps: I think it's quite likely that the `sigmoid` and `tanh` C++ needs a refresh to get bit-exactness wrt QKeras, they both have some hard coded values still (the `init_{sigmoid | tanh}_table` functions in particular)

hls4ml/utils/config.py
```diff
@@ -32,12 +32,24 @@ def _get_precision_from_quantizer(quantizer):
...
+    overflow = "AP_SAT"
+
+    if quantizer['class_name'] == 'quantized_tanh':
+        overflow = "AP_SAT_SYM" if quantizer['config']['symmetric'] else "AP_SAT"
+        bits = int(quantizer['config']['bits'])
+        integer = 1
+    elif quantizer['class_name'] == 'quantized_sigmoid':
+        bits = int(quantizer['config']['bits'])
+        integer = 0
+        signed = False
```
thesps, line 51: Maybe move these into the `if quantizer['class_name'] in supported_quantizers:` statement, as an `elif` after `if quantizer['class_name'] == 'quantized_relu'` - just to improve readability so all the special cases after the logic that applies to them all.

hls4ml/utils/config.py
```diff
@@ -32,12 +32,24 @@ def _get_precision_from_quantizer(quantizer):
...
+        else:
             quantizer['class_name'] = quantizer_obj.__name__
 
-    supported_quantizers = ['quantized_bits', 'quantized_relu', 'quantized_tanh', 'quantized_po2', 'quantized_relu_po2']
+    supported_quantizers = ['quantized_bits', 'quantized_relu', 'quantized_tanh',
+                            'quantized_sigmoid', 'quantized_po2', 'quantized_relu_po2']
     signed = True
-    if quantizer['class_name'] in supported_quantizers:
+    rnd = "AP_RND_CONV"
+    overflow = "AP_SAT"
```
thesps, line 42: These will apply to weights as well as activations. It may not matter, since the 'rounding' for weights actually happens in Python and not HLS, but we should make sure it doesn't have any implications on the HLS results.

thesps:
Also can you push to a local branch (like `pr/569`) to trigger Gitlab CI for pytests?

jmitrevs:
Let me do a few more tests first, then I'll push to a local branch.

hls4ml/utils/config.py
jmitrevs: I was wondering whether I should put the ap_fixed defaults (AP_TRN and AP_WRAP) or these as the default, and didn't really have convincing arguments one way or the other. The QONNX route tends to have saturation and rounding, so I chose that at the end, but it was not a strong argument.

thesps: I think we just need to see that:
- no QKeras tests that were passing now fail
- resources are the same for some synthesized QKeras models on master and on this branch, with all else being the same

jmitrevs:
The issue that we are seeing is that the hls4ml sigmoid really seems to approximate `1/1+exp(-x)`, but the QKeras one is quite different in a way I don't currently understand.

jmitrevs:
So qkeras uses the hard sigmoid by default:
https://github.com/google/qkeras/blob/master/qkeras/quantizers.py#L192-L237
I have to see if hls4ml implements it; otherwise, it's easy to do:  `clip(0.5 * x + 0.5, 0.0, 1.0)`.

hls4ml/utils/config.py
jmitrevs: Done

jmitrevs:
This resulted in more changes than I expected. Please take a look to see if this is reasonable.

jmitrevs:
I am not sure how the matplotlib import came into this pull request. It should not be there. I will fix that.

jmitrevs:
Merged in master to fix pytest setup.

test/pytest/test_qkeras.py
jmitrevs: This has changed so I will resolve this conversion.

hls4ml/utils/config.py
jmitrevs: I was wondering if we should defer this to the OutputRoundingSaturationMode optimizer. That might be more consistent with what is done in other cases. I will try that.

jmitrevs: I decided to go back to using the ap_fixed defaults.

jmitrevs:
I pushed the changes to pr/569

thesps:
I notice in the qkeras pytest we get:
`E   TypeError: __init__() got an unexpected keyword argument 'use_real_tanh'`
It's because that option is in QKeras main branch but not in a release.

hls4ml/converters/keras/qkeras_layers.py
```diff
@@ -72,26 +75,38 @@ def parse_qactivation_layer(keras_layer, input_names, input_shapes, data_reader,
...
     if activation_config['class_name'] == 'ternary_tanh':
         layer['class_name'] = 'TernaryTanh'
         layer['threshold'] = activation_config.get('config', {}).get('threshold', 0.33)
         if layer['threshold'] is None:
             layer['threshold'] = 0.33 # the default ternary tanh threshold for QKeras
+        layer['activation'] = 'ternary_tanh'
+    elif ((activation_config['class_name'] == 'quantized_sigmoid'
+          and not activation_config['config']['use_real_sigmoid'])
+          or (activation_config['class_name'] == 'quantized_tanh'
+          and not activation_config['config']['use_real_tanh'])):
```
thesps, line 94: Since these options are not in a QKeras release yet, maybe we should access them with a `.get` returning a default value of `False` for robustness

jmitrevs: I just committed the fix.

thesps:
So a few tests are still failing, including the existing `hard_sigmoid` tests in `test_activations`, some of which fail and at least one of which now creates a segmentation fault. The test with these settings produces the segfault:

```
backend, shape, io_type, activation, name = 'Quartus', (8, ), 'io_stream', Activation('hard_sigmoid'), 'hard_sigmoid'
```

Apart from that, I think we should remove the CNN fixes to a separate PR to keep this one on topic.

jmitrevs:
The CNN fixes are actually in #610. You are right that it should not be here. I am not sure if it's best to try to get that in first or to revert the commit.

jmitrevs:
I fixed up #610 so maybe it's easier to merge this after that PR (and I fix the remaining issues).

jmitrevs:
I think this is ready, but I would like to stage it after #610 (which is also ready). That way the encoded fix, that's partially included here, does not need to be removed. After #610 is merged, I would need to glance one more time at this PR.

hls4ml/converters/keras_to_hls.py
```diff
@@ -257,7 +257,17 @@ def parse_keras_model(model_arch, reader):
...
+    activation_layers = [
+        'Activation',
+        'LeakyReLU',
+        'ThresholdedReLU',
+        'ELU',
+        'PReLU',
+        'Softmax',
+        'TernaryTanh',
+        'HardActivation',
+        'QActivation',
```
jmitrevs, line 269: Double-check if QActivation should be here.

jmitrevs:
It still passes with those values. I do wonder if it's the qkeras version. I noticed after doing git pull on the master branch the pytests no longer work for unrelated reasons (`TypeError: Got an unexpected keyword argument 'use_legacy_format'`), and checkout out v0.9.0 fails for compatibility reasons. I probably need to target v0.9.0 and make sure that the tests succeed with that version (and disable quantized sigmoid, which is not supported in that version.)

vloncar:
`use_legacy_format` comes from keras serialization functions. You need the latest TF installation for QKeras main. That's a bummer, we need them to make a release and fix on that. 0.9.0 is getting a bit old.

jmitrevs:
So the test fails with tensorflow 2.4.1 (used by the CI) but passes with 2.9.1 and 2.11.0. I don't know why. Should I increase the allowed difference?

jmitrevs:
The pytests now succeed. I am not sure why there are the CI failures, though. Is it because we couldn't build the docker image with Vivado?

vloncar:
Looks like a real error, probably something was generated with a syntax error.

jmitrevs:
I was able to manually reproduce it. It is in the vivado_accelerator_writer not liking the pre-commit changes:
```python
            elif 'void myproject(' in line:
                newline = f'void {model.config.get_project_name()}_axi(\n'
```
This basically removes that arguments that were added to the same line by the pre-commit:
```C++
void myproject(input_axi_t in[N_IN], output_axi_t out[N_OUT]);
```
The fix should be easy. I'll try to fix it also in vivado_writer where it currently doesn't cause a problem but could.

vloncar (approved):

PR #570, opened 2022-06-16
Softmax LUT Optimization

bo3z:
This PR exploits the following properties of the current Softmax implementation:
1. For all inputs, the maximum is subtracted, before calculating the exponential. Therefore the exponential lookup table need only have entries for negative inputs.
2. The sum of exponential is always positive. Therefore, the invert table need only have positive values, not negative.

The following plot shows the accuracy of hls4ml (old vs. new) compared to Keras at identifying argmax - the accuracy is identical for both implementations:
acc

The following plot shows the mean absolute percentage error of hls4ml (old vs. new) compared to Keras's output - new version performs slightly better compared to the old one due to more "relevant" elements in the LUT:
mape

Finally, the arrays holding the differences between the elements and the maximum is removed (as no significant accuracy was achieved), so resources were saved, as seen below:
softmax_results

### DIFF
--- a/hls4ml/utils/fixed_point_utils.py
+++ b/hls4ml/utils/fixed_point_utils.py
@@ -1,8 +1,5 @@
 import sys
 import math
-from unicodedata import decimal
-
-from numpy import integer
 
 '''
 A helper class for handling fixed point methods
@@ -70,7 +67,6 @@ class FixedPointEmulator:
                 self.integer_bits[i] = bits[i]
             elif i >= self.I and i<self.N:
                 self.decimal_bits[i-self.I] = bits[i]
-        # print('Len bits ' + str(len(bits)) + ' Inside FPU ' + str(self.integer_bits) + str(self.decimal_bits))
 
     '''
     Returns e^x, where x is the current fixed point number
--- a/hls4ml/utils/fixed_point_utils.py
+++ b/hls4ml/utils/fixed_point_utils.py
@@ -1,5 +1,8 @@
 import sys
 import math
+from unicodedata import decimal
+
+from numpy import integer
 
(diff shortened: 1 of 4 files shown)


bo3z:
@thesps maybe you could review this PR, since you reviewed the first Quartus Softmax implementation?

hls4ml/utils/fixed_point_utils.py
```diff
@@ -1,5 +1,8 @@
 import sys
 import math
+from unicodedata import decimal
```
vloncar, line 3: why do you need these imports?

hls4ml/utils/fixed_point_utils.py
```diff
@@ -65,8 +68,9 @@ def set_msb_bits(self, bits):
         for i in range(0, len(bits)):
             if i < self.I:
                 self.integer_bits[i] = bits[i]
-            elif i >= self.I and i<self.F:
+            elif i >= self.I and i<self.N:
                 self.decimal_bits[i-self.I] = bits[i]
+        # print('Len bits ' + str(len(bits)) + ' Inside FPU ' + str(self.integer_bits) + str(self.decimal_bits))
```
vloncar, line 73: Remove?

hls4ml/utils/fixed_point_utils.py
bo3z: Don't think I do, will remove.

hls4ml/utils/fixed_point_utils.py
bo3z: Yes.

vloncar (approved):

PR #575, opened 2022-06-19
Implementation and optimizations linked to Simple-RNN and LSTM for qu…

nemerchiedde:
This PR explores the integration of two RNNs for quartus. The current implementation has integrated the Simple-RNN and LSTM.
Data can be passed in two different formats (single-cell and sliding-window):

- Single-cell:
The LSTM and Single RNN cell are computed at every instance, and the same RNN operation is applied for each value combining the past state and the new value (this procedure is repeated until the end of the data)

- Sliding-Window:
Full sequence split into overlapping subsequences with a sliding window, used for time date acquisition.

The activation functions for Simple RNN and LSTM were recovered directly from the trained model and then were automatically overwritten for the RNN, so no changes in the code are longer necessary. This is also true for the recurrent activation function of LSTM.

Concerning the LSTM, the weights (kernel, recurrent kernel and bias) have been subdivided into their respective gates to allow bit_width independence.


hls4ml/backends/quartus/passes/transform_types.py
```diff
@@ -27,7 +27,7 @@ def transform(self, model, node):
                 elif node.name in node.model.outputs:
                     new_var = self.struct_var_converter.convert(var, pragma='hls_register', struct_name='outputs')
                 else:
-                    new_var = self.array_var_converter.convert(var, pragma='hls_register')
+                    new_var = self.array_var_converter.convert(var, pragma=' ')
```
bo3z, line 30: Should avoid removing this pragma as it affects other layers in Quartus as well. 

On the other hand, having this pragma causes HLS compilation to fail with a message along the lines of "Failed to place variable in registers, you might be taking a reference or a pointer to a variable". I've seen this error before, when implementing streams, and I think (this should be easy to verify) a likely cause here is the fact that the LSTM and SimpleRNN methods take arguments as pointers, rather than static C++ arrays of fixed size (even though these two are usually interchangeable) which is how other layers implemented. So, should try changing all method argument pointers to statics arrays.

If this doesn't work, we should further debug this, as it likely decreases performance in other layers

hls4ml/backends/quartus/quartus_backend.py
```diff
@@ -188,3 +201,56 @@ def init_softmax(self, layer):
             layer.set_attr('exp_table_t', layer.get_attr('table_t'))
         if 'inv_table_t' not in layer.attributes:
             layer.set_attr('inv_table_t', layer.get_attr('table_t'))
+
```
bo3z, line 204: HLS code for `SimpleRNN` is included but it's `init` function isn't here?

hls4ml/backends/quartus/quartus_backend.py
```diff
@@ -188,3 +201,56 @@ def init_softmax(self, layer):
...
+            layer.weights['recurrent_weight_i'].data = np.transpose(layer.weights['recurrent_weight_i'].data)
+
+            layer.weights['weight_f'].data = np.transpose(layer.weights['weight_f'].data)
+            layer.weights['recurrent_weight_f'].data = np.transpose(layer.weights['recurrent_weight_f'].data)
+
+            layer.weights['weight_c'].data = np.transpose(layer.weights['weight_c'].data)
+            layer.weights['recurrent_weight_c'].data = np.transpose(layer.weights['recurrent_weight_c'].data)
+
+            layer.weights['weight_o'].data = np.transpose(layer.weights['weight_o'].data)
+            layer.weights['recurrent_weight_o'].data = np.transpose(layer.weights['recurrent_weight_o'].data)
```
bo3z, line 250: Transposing could be moved to a resource strategy optimizer, as is done for many layers in Vivado. I added the basic optimizer in this commit: https://github.com/fastmachinelearning/hls4ml/pull/561/commits/068dcc7086cd966625006726b04a5689537cc587 for Conv layers (still unmerged)

hls4ml/model/layers.py
```diff
@@ -880,11 +880,17 @@ def initialize(self):
...
+        self.add_weights(quantizer=self.get_attr('weight_quantizer'), compression=self.model.config.get_compression(self))
+        self.add_bias(quantizer=self.get_attr('bias_quantizer'))
+
+        data  = self.model.get_weights_data(self.name, 'kernel')
+        data2 = self.model.get_weights_data(self.name, 'recurrent_kernel')
+        data3 = self.model.get_weights_data(self.name, 'bias')
+
+        self.add_weights_variable(name='weight', var_name='kernel_{index}' , data=data[0][0:self.get_attr('n_out')], quantizer=self.get_attr('weight_quantizer'), compression=None)
+        self.add_weights_variable(name='recurrent_weight', var_name='recurrent_kernel_{index}', data=data2[0:self.get_attr('n_out'),0:self.get_attr('n_out')], quantizer=self.get_attr('weight_quantizer'), compression=None)
+        self.add_weights_variable(name='bias', var_name='bias_{index}', data=data3[0:self.get_attr('n_out')], quantizer=self.get_attr('weight_quantizer'), compression=None)
```
bo3z, line 893: Possible duplication? Lines 884-885 already add the 'weight' and 'bias' variables, so 887 & 889 would simply be re-doing the same operation? However, 888 is still needed, as the recurrent weights are not included yet.

hls4ml/backends/quartus/passes/transform_types.py
nemerchiedde: Solved!

hls4ml/backends/quartus/quartus_backend.py
```diff
@@ -188,3 +201,56 @@ def init_softmax(self, layer):
...
+        layer.add_weights_variable(name='recurrent_bias_i', var_name='br_i_{index}', data=recurrent_bias_i)
+
+        recurrent_bias_f = np.zeros(layer.weights['recurrent_weight_f'].shape[1])
+        layer.add_weights_variable(name='recurrent_bias_f', var_name='br_f_{index}', data=recurrent_bias_f)
+
+        recurrent_bias_c = np.zeros(layer.weights['recurrent_weight_c'].shape[1])
+        layer.add_weights_variable(name='recurrent_bias_c', var_name='br_c_{index}', data=recurrent_bias_c)
+
+        recurrent_bias_o = np.zeros(layer.weights['recurrent_weight_o'].shape[1])
+        layer.add_weights_variable(name='recurrent_bias_o', var_name='br_o_{index}', data=recurrent_bias_o)
```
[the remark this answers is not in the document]
nemerchiedde: I removed.

hls4ml/model/layers.py
nemerchiedde: Correct, removed lines 884 and 885

vloncar (approved):
This has been rebased, cleaned up and tested as part of a mini hackathon in Marseille. Many thanks to Nemer, Benjamin and the CPPM people for this contribution!

Issue #573, opened 2022-06-16
LayerName configuration capitalization


jmitrevs:
I noticed that functions accessing the configuration use `layer.name.lower()` as the key, for example in:
```python
    def get_layer_config(self, layer):
        hls_config = self.config['HLSConfig']
        layer_config = {}

        type_config = hls_config.get('LayerType', {}).get(layer.class_name, None)
        if type_config is not None:
            layer_config.update(type_config)

        name_config = hls_config.get('LayerName', {}).get(layer.name.lower(), None)
        if name_config is not None:
            layer_config.update(name_config)

        return layer_config
```
But in `config_from_keras_model` when you use the granularity='name' feature, the name is used directly, without being made `.lower()`:  https://github.com/fastmachinelearning/hls4ml/blob/master/hls4ml/utils/config.py#L299-L305. So the generated configuration does not pass the information correctly because it could well have capital letters in the keys. This needs to be made consistent. I think there are two reasonable options:
1.  Make the comparisons case-sensitive
2. Make the name comparison case-insensitive, even if they are entered in mixed case in the configuration

One could add a `.lower()` in `config_from_keras_model` as well, but I think that's a worse option, because it would be hard to enforce in hand-generated configurations. For the second option, I think there could a key "fixing" step in the `HLSConfig` constructor to make the keys lower case.

What do people think?

PR #577, opened 2022-06-21
Make layer names case sensitive in config.

jmitrevs:
This implements option 1 of #573:  all keys in the HLSConfig are made case sensitive. I think it is logically simpler to make the config case sensitive or insensitive everywhere, not just in some places, and the first option is easier. This PR fixes being able to turn on tracing if layer names have capitalizations, where part assumed case sensitive, and another part assumed case insensitive.


jmduarte (approved):
LGTM!

PR #585, opened 2022-06-27
Softsign optimization

nemerchiedde:
Storing only the positive part of the Softsign is possible due to its symmetry in the Y axis. Same theory when presented in pull request: https://github.com/fastmachinelearning/hls4ml/pull/527


bo3z:
Looks good to me. Would be nice to add a test, just to show correct behaviour. We could do it in a similar way to other activations, in `test/test_keras_api.py`

nemerchiedde:
> Looks good to me. Would be nice to add a test, just to show correct behaviour. We could do it in a similar way to other activations, in `test/test_keras_api.py`

Added one test file.

test/pytest/test_softsign.py
```diff
@@ -0,0 +1,58 @@
...
+    hls_model.compile()
+   
+    y_keras = model.predict(X)
+    y_hls4ml = hls_model.predict(X).reshape(y_keras.shape)
+
+    acc_hls4ml = accuracy_score(np.argmax(y_keras, axis=-1).ravel(), np.argmax(y_hls4ml, axis=-1).ravel())
+
+    print('Accuracy hls4ml relative to keras: {}'.format(acc_hls4ml))
+
+    assert acc_hls4ml >= 0.7
```
vloncar, line 56: Why only 0.7?

nemerchiedde: Changed to 0.98. New tests have also been added to the pytest file.

vloncar (approved):

PR #503, opened 2022-03-11
MLPerf Tiny developments

jmduarte:
Not to be merged but simply to open discussion about which developments we should cherry-pick and rebase to hls4ml main branch. Based on a quick look, I can imagine a few separate PRs:

- [x] FIFO depth profiling @nicologhielmetti #509 
- [x] ReLU merge optimizer @anmeza @oliviaweng #586 
- [x] QDenseBatchnorm support @julesmuhizi #718 (requires the corresponding QKeras PR to be merged google/qkeras#74)
- [ ] Arty support for VivadoAccelerator backend @GiuseppeDiGuglielmo
- [ ] Ultra96 support for VivadoAccelerator backend @GiuseppeDiGuglielmo #646 (to be split up)
- [ ] AXI Master interface for VivadoAccelerator backend @GiuseppeDiGuglielmo #646 (to be split up)
- [ ] Reprogrammable weights for VivadoAccelerator backend @GiuseppeDiGuglielmo #646 (to be split up)

PR #586, opened 2022-06-29
Relu merge optimizer pass

oliviaweng:
# Description

We introduce an hls4ml optimizer pass for merging the ReLU layer into the Dense/Conv2D layers when ReLU immediately follows them---a frequently encountered pattern in neural networks (NNs). NNs in hls4ml are spatially laid out using dataflow stages to implement each layer, which are linked together by FIFOs. These FIFOs can cost BRAMs, LUTs, and/or FFs. By default in hls4ml, each rectified linear unit (ReLU) is implemented as its own dataflow stage. Because each additional dataflow stage costs extra logic and FIFOs, we reduce the resource utilization by merging the ReLU activation function into the layer preceding it. Although the layers with the newly merged ReLU functionality use more logic than before, there is still a net decrease in resources.This optimization was introduced in hls4ml's MLPerf TinyML Benchmark 2022 submission and written up in [this paper](https://arxiv.org/abs/2206.11791). Resource reductions introduced by this optimization are reported in the paper.

## Type of change

This optimization pass was first mentioned in the MLPerf TinyML PR https://github.com/fastmachinelearning/hls4ml/pull/503. 

## Tests

[This repo](https://github.com/oliviaweng/hls4ml_dense_relu_merge) contains two test models (a fully-connected NN and a CNN) that can be trained on MNIST, converted into Vivado HLS, and synthesized using Vivado HLS 2020.1.

## Checklist

### DIFF
--- a/hls4ml/model/optimizer/passes/merge_relu.py
+++ b/hls4ml/model/optimizer/passes/merge_relu.py
@@ -5,8 +5,12 @@ class MergeRelu(OptimizerPass):
         supported_layers = ['Conv2D', 'Conv2DBatchnorm', 'Dense']
         is_match = node.get_input_node().__class__.__name__ in supported_layers
 
-        # hls4ml names ReLU activations 'Activation'
+        # hls4ml names ReLU activations 'Activation' TODO: Node class name isn't
+        # Activation anymore.. it can change and in our test case is called
+        # VivadoAcceleratorActivation
         is_match = is_match and (node.__class__.__name__ == 'Activation') 
+        print(f"Node class name = {node.__class__.__name__}")
+        print(f"Does layer {node.__class__.__name__} match Relu merge? {is_match}")
         return is_match
 
     def transform(self, model, node):
--- a/hls4ml/model/optimizer/passes/merge_relu.py
+++ b/hls4ml/model/optimizer/passes/merge_relu.py
@@ -1,22 +1,18 @@
 from hls4ml.model.optimizer import OptimizerPass
+from hls4ml.model.layers import Activation, Dense, Conv2D, Conv2DBatchnorm
 
 class MergeRelu(OptimizerPass):
     def match(self, node):
-        supported_layers = ['Conv2D', 'Conv2DBatchnorm', 'Dense']
-        is_match = node.get_input_node().__class__.__name__ in supported_layers
+        supported_layers = (Dense, Conv2D, Conv2DBatchnorm)
 
-        # hls4ml names ReLU activations 'Activation' TODO: Node class name isn't
-        # Activation anymore.. it can change and in our test case is called
-        # VivadoAcceleratorActivation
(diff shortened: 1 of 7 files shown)


jmduarte:
Hi @oliviaweng thanks a lot for the contribution, it looks great!

It seems some tests are failing: https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/pipelines/4158655

I think most of the failures are relatively easy to fix, e.g. `CONFIG_T::out_t` is missing. Could you take a look and see if you can make them pass? We can also discuss if something is not clear with the error

hls4ml/model/optimizer/passes/merge_relu.py
```diff
@@ -0,0 +1,48 @@
...
+from hls4ml.model.layers import Activation, Dense, Conv2D, Conv2DBatchnorm
+
+class MergeRelu(OptimizerPass):
+    def match(self, node):
+        supported_layers = (Dense, Conv2D, Conv2DBatchnorm)
+
+        is_match = issubclass(node.get_input_node().__class__, supported_layers)
+        # ReLU layers are of class Activation
+        is_match = is_match and issubclass(node.__class__, Activation)
+        return is_match
```
oliviaweng, line 11: I see. For the missing `CONFIG_T::out_t` error, it looks like this `match()` function is too generous. Since all activation layers are of subclass `Activation`, any Dense/Conv2D layer that is followed by any activation function returns True, which is wrong. I'll look into tightening this up.

vloncar:
A question on the approach itself: This will only work for ReLU and will require duplicate overrides for every possible combination of activation, layer and HLS implementation if we want to expand it. Instead of extending the kernel of the matrix-vector multiplication to tack on ReLU computation and then create duplicate function calls for that, why don't we introduce a new operation that is nop by default that sits at the end of the dense function? Basically the where the `cast` function sits now. Then the config can include a proper implementation of activation if we choose to merge it in. Like in config class we add `template<class data_T, class CONFIG_T> using activation = nnet::some_activation_or_nop<data_T, CONFIG_T>;`

This probably sounds unclear, so if you have 15 minutes we can chat over zoom.

jmitrevs:
Should we discuss @vloncar's suggestion? It seems like there is quite a bit of interest in this PR so it will be good to get it in.

vloncar:
It was discussed offline and we converged on the proper approach to this.

jmitrevs:
Any more news on this?

oliviaweng:
@abijithYayavaram is currently building a more generic version. We aim to push some updates within the next several weeks.

jmitrevs:
What is the plan for this in general? Is this where we'll attack with the new code generation framework?

vloncar:
I believe we should revisit this once we have a better way of generating functions in which activations would be merged. There's little gain in general if FIFO depth optimization is applied (and none for io_parallel).

PR #509, opened 2022-03-21
FIFO depth optimization

nicologhielmetti:
The PR optimize the depth of the FIFOs implemented in `io_stream` based models by estimating the actual usage of the FIFOs through cosimulation. For more details consider this presentation -> https://docs.google.com/presentation/d/1ItM-8EAlNRdjRk4Cu1LY7gVvCavD90wrE95SsH2wQQw/edit?usp=sharing

How to use it:
https://gist.github.com/nicologhielmetti/3a268be32755448920e9f7d5c78a76d8

PR #588, opened 2022-07-03
Update flow dependencies

vloncar:
## Description

This PR shuffles the order of flows a bit to help resolve the issues observed in #509 and elsewhere. Essentially it makes writer flows depend on the ip flows and expands functionality of `apply_flows()` to control how the flows are reapplied if they were already applied before. Another effect of the change is that the flows aren't applied multiple times, an issue that many people have observed.

## Type of change
- [x] Kind of a bug fix since it fixes multiple flows problem
- [x] New feature

## Tests
No extra tests are needed. To reproduce the old behavior, just convert any model and call `compile()` and `build()` and see how many times the same optimizers will be applied (in `model._applied_flows`).

## Checklist
- [x] My changes generate no new warnings. Yeah, right, as if I could prove this.

### DIFF
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -67,10 +67,10 @@ class VivadoBackend(FPGABackend):
         template_flow = register_flow('apply_templates', self._get_layer_templates, requires=[init_flow], backend=self.name)
 
         writer_passes = [
+            'make_stamp',
             'vivado:write_hls'
         ]
-        writer_flow_requirements = ['optimize', vivado_types_flow, template_flow]
-        self._writer_flow = register_flow('write', writer_passes, requires=writer_flow_requirements, backend=self.name)
+        self._writer_flow = register_flow('write', writer_passes, requires=['vivado:ip'], backend=self.name)
 
         all_passes = get_backend_passes(self.name)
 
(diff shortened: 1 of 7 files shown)


hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -173,6 +173,7 @@ def init_sepconv1d(self, layer):
 
     @layer_optimizer(Conv2D)
     def init_conv2d(self, layer):
+        print('here in conv2d')
```
thesps, line 176: Left over `print`

hls4ml/model/optimizer/passes/stamp.py
```diff
@@ -0,0 +1,19 @@
...
+    def transform(self, model):
+        def _make_stamp():
+            """ Create a unique identifier for the generated code. This identifier is used to
+            compile a unique library and link it with python. """
+            from string import hexdigits
+            from random import choice
+            length = 8
+            return ''.join(choice(hexdigits) for m in range(length))
+
+        model.config.config['Stamp'] = _make_stamp()
```
thesps, line 17: I know this isn't the point of this PR, but would it be safer to make the stamp a (hidden?) attribute of the model rather than a config entry?

thesps:
Perhaps some tests with dummy models and dummy flows, with asserts on the `_applied_flows`? That could be quite useful also as a way to express and bake-in the intended behaviour for any future developments.

hls4ml/backends/vivado/vivado_backend.py
vloncar: Removed, thanks!

hls4ml/model/optimizer/passes/stamp.py
vloncar: Yes! I was thinking that since layers have attributes, so should the model. Then all the config stuff that must be refactored can be implemented this way. But it's for the next PR.

vloncar:
I can add the test, but it would be pretty naive. It will test only the `apply_flow()` function, which is pretty simple.

thesps (approved):
Added a simple pytest doing dummy flows on a dummy project and checking that the order of applied passes matches expectation. Everything looks good, merging!

PR #596, opened 2022-07-11
Quartus GRU

bo3z:
Description
> :memo: Gated Recurrent Units (GRUs) for Quartus backend
>  
> * Adds support for GRU Units in Quartus backend
> * Similar to work done on the Vivado backend (#576) and Quartus LSTM & SimpleRNN (#575)

## Type of change

## Checklist

## Implementation

- HLS code for GRU layers consist of two main functions:

1. `gru_cell(t, h, weights, recurrent_weights, bias, recurent_bias)` - which takes in the input vector, x, and hidden state, h, at time step t, and produces the new hidden state according to GRU logic (reset, update, candidate state gate) - this function has several loops over the number of GRU units/states; therefore, those loops are often unrolled with the appropriate reuse factor. For results on resource usage and latency, see below.
2. `gru(data, res, weights, recurrent_weights, bias, recurent_bias)` - makes use of the previously mentioned function, by traversing through the data at each time step and obtaining the new state, until the final output is obtained. Note, it is not possible to pipeline this function, because there is a loop dependency (LD). Namely, the at every iteration, the state needs to be available so that the new state can be calculated.

- The backend containes a layer initialiser and the appropriate templates. Matrix multiplication and bias addition is done through the `Dense` layer. Finally, a resource strategy optimizer handles matrix transposes needed for `Dense` multiplication, rather than being done in layer initialising procedures.

## Results

Below are latency, DSP, REG and ALM usage results of a GRU layer with a 5-dimensional input, 8 time steps and a variable number of units. 

As expected, the latency remains approximately constant when increasing the number of units, while DSPs, REGs and ALM increase at a linear rate. This occurs because the implementation contains several loops unrolled over the number of units/states. Therefore, such an implementation is time-invariant, but resource-ineffficient.
latency - units
dsp - units
reg - units
alm - units

Finally, with the units fixed to 8 and the input size to 5, similar plots are obtained. As the time loop has pipelining disabled (due to loop dependencies), the use of DSPs remains approximately constant. ALMs and REGs increase slightly, because a larger input needs to be stored. The latency increases at a linear rate, as expected. 
latency-time steps
dsp - time steps
reg - time steps
alm - time steps

### DIFF
--- a/hls4ml/templates/quartus/firmware/nnet_utils/nnet_recurrent.h
+++ b/hls4ml/templates/quartus/firmware/nnet_utils/nnet_recurrent.h
@@ -1,0 +1,141 @@
+#ifndef NNET_RECURRENT_H_
+#define NNET_RECURRENT_H_
+
+#include "nnet_common.h"
+#include "nnet_dense.h"
+#include "nnet_recurrent_activation.h"
+
+namespace nnet {
+
+struct gru_config {
+    // Internal data type definitions
+    typedef float weight_t;
+    typedef float bias_t;
+    typedef float accum_t;
+
+    // Layer Sizes
+    static const unsigned n_in =  1;
+    static const unsigned n_out = 1;
+    static const unsigned n_units = 1;
+    static const unsigned n_timesteps = 1;
+    static const unsigned n_outputs = 1;
+    static const bool return_sequences = false;
+
+    // Resource reuse info
+    static const unsigned io_type = io_parallel;
+    static const unsigned reuse_factor = 1;
+    static const bool store_weights_in_bram = false;
+    
+    template<class x_T, class y_T, class config_T>
+    using activation_recr = nnet::activation::relu<x_T, y_T, config_T>;
+    
+    template<class x_T, class y_T, class config_T>
+    using activation = nnet::activation::relu<x_T, y_T, config_T>;
+};
+
+template<class data_T, class res_T, typename CONFIG_T>
+void gru_cell(
+    data_T x[CONFIG_T::n_in],
+    res_T  h[CONFIG_T::n_units],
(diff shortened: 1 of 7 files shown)


jmitrevs:
The pytest failure was related to running out of disk space. It probably is unrelated to this PR.

hls4ml/templates/quartus/firmware/nnet_utils/nnet_recurrent.h
```diff
@@ -0,0 +1,143 @@
...
+
+    // Activation on z(t) and r(t)
+    hls_register typename CONFIG_T::accum_t z_r_act [2*CONFIG_T::n_units]; 
+    CONFIG_T::template activation_recr<typename CONFIG_T::accum_t, typename CONFIG_T::weight_t, typename CONFIG_T::ACT_CONFIG_RECURRENT_T>::activation(z_r, z_r_act);
+
+    // A matrix containing the values of Hadamard product between r(t) = z_r_act[n_units:2*n_units] and h(t-1) = h
+    hls_register typename CONFIG_T::accum_t hadamard_r_h[CONFIG_T::n_units];
+    #pragma unroll recurrent_unroll_factor
+    for(int i = 0; i < (CONFIG_T::n_units); i++) {
+      hadamard_r_h[i] = z_r_act[i + CONFIG_T::n_units] * mat_mul_h_wr[i + 2 * CONFIG_T::n_units];
```
jmitrevs, line 72: Very minor point:  I think the indentation is not consistent, being 2 in some places, 4 in others.

bo3z: Fixed.

vloncar (approved):

PR #597, opened 2022-07-13
Override parent backend optimizer passes with derived backend passes

thesps:
Opening this PR for discussion, with the solution if agreed.

In #509 we noticed that if a base and derived `Backend` class both define an optimizer pass of the same name, it is the base backend pass that remains registered to the derived backend set of passes. This is a bit counterintuitive - I think you'd expect the derived class to override passes in the same way it might override methods. I put the change in a separate branch since it touches something deeper in the framework.

A concrete example from #509:
```
print(hls4ml.model.optimizer.get_optimizer('vivadoaccelerator:fifo_depth_optimization'))
<hls4ml.backends.vivado.passes.fifo_depth_optimization.FifoDepthOptimization object at 0x7f6f353b5820>
```
Notice I get the pass registered to `vivadoaccelerator` backend, which returns an object in `hls4ml.backends.vivado`.

With this PR I changed the behaviour to register the base class passes first, which are then updated with the current class. 
So now it yields:
```
print(hls4ml.model.optimizer.get_optimizer('vivadoaccelerator:fifo_depth_optimization'))
<hls4ml.backends.vivado_accelerator.passes.fifo_depth_optimization.FifoDepthOptimization object at 0x7f35d77f65e0>
```
As expected, getting the pass from `vivadoaccelerator` returns the pass from `hls4ml.backends.vivado_accelerator`.

I think this case is our only example of the derived class defining a pass of the same name as the parent class. The backends that derive from `FPGABackend` only add passes, they don't override (I think).


thesps:
@bo3z since you added this optimizer pass inheritance for #523, maybe you can take a look if this makes sense

thesps:
> Although it seems that one of the tests fails.
I think it's unrelated to this, but to #602. I'm happy to wait until that's resolved in any case.

 I think all that you added makes sense, and then I think this PR is needed to get the intended inheritance behaviour

bo3z:
> > Although it seems that one of the tests fails.
> 
> I think it's unrelated to this, but to #602. I'm happy to wait until that's resolved in any case.
> 
> I think all that you added makes sense, and then I think this PR is needed to get the intended inheritance behaviour

Sounds good to me, happy to merge this.

bo3z (approved):

PR #605, opened 2022-07-14
Enforce function pipelining when using io_parallel with Resource strategy

vloncar:
# Description

`Resource` strategy implies using `DATAFLOW` at the top-level function, which doesn't cause loop unrolling and function pipelining. In this case, functions that don't explicitly call `#pragma HLS PIPELINE II=...` won't be pipelined and will suffer from long latency. We pipeline most functions, but there were a few missing. This PR adds this directive to the functions missing it.

While I was at it, I also cleaned up the pooling implementation and fixed the `concatenate1d` for `io_stream`.

The `transpose_3d` function (corresponding to the `Permute` Keras layer) is left out, because this function needs to be replaced with something synthesizable. This will be addressed in a follow up PR.

## Type of change

## Tests

The included pytest only tests the validity of the merge functions. To test the synthesis difference, same functions can be used, just use `Resource` strategy and call `build` to see it before and after the change.

For example:
```
import hls4ml
from tensorflow.keras.models import Model
from tensorflow.keras.layers import Input, Add, Average, Maximum, Minimum, Multiply, Subtract

def test_merge(merge_layer):
    input_shape = (10, 10, 3)

    in1 = Input(shape=input_shape)
    in2 = Input(shape=input_shape)
    out = merge_layer()([in1, in2])
    
    model = Model(inputs=[in1, in2], outputs=out)
    model.compile(optimizer='adam', loss='mse')

    config = hls4ml.utils.config_from_keras_model(model, default_precision='ap_fixed<32,16>')
    config['Model']['Strategy'] = 'Resource'
    output_dir = 'hls4mlprj_merge_{}'.format(merge_layer.__name__.lower())
    hls_model = hls4ml.converters.convert_from_keras_model(model, hls_config=config, output_dir=output_dir, io_type='io_parallel')
    report = hls_model.build(synth=True)
    print(report['CSynthesisReport'])

for cls in [Add, Average, Maximum, Minimum, Multiply, Subtract]:
    test_merge(cls)
```

## Checklist

### DIFF
--- a/hls4ml/backends/vivado/passes/pooling_templates.py
+++ b/hls4ml/backends/vivado/passes/pooling_templates.py
@@ -18,7 +18,7 @@ pooling1d_config_template = """struct config{index} : nnet::pooling1d_config {{
     static const unsigned stride_width = {stride_width};
     static const nnet::Pool_Op pool_op = nnet::{pool_op};
     static const nnet::conv_implementation implementation = nnet::conv_implementation::{implementation};
-    static const unsigned reuse = {reuse};
+    static const unsigned reuse_factor = {reuse};
     typedef {accum_t.name} accum_t;
 }};\n"""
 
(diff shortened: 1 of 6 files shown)


jmitrevs:
I made pr/605 to trigger the tests.

jmitrevs (approved):
I commented on the logic change that I saw, which looks correct. Given the many differences in nnet_pooling.h it's possible I missed something, but overall things look good to me.

vloncar:
The logic change had to be done, because the previous implementation didn't work correctly for the way we pack data into streams

PR #606, opened 2022-07-15
WIP Add custom KL loss layer HLS implementation

katyagovorkova:
Adds an implementation of the **KL loss layer** used for _CMS Anomaly detection at L1_.
Adds an example of usage of **KL layer** is in `contrib/kl_layer.py` and the `HLS` part is in `hls4ml/templates/vivado/nnet_utils/nnet_distance.h`.
The original implementation of the **KL layer** is available on the `AE_L1_paper` branch, this PR updates the implementation for the new layer API. 

## Type of change
- [x] Documentation update

## Tests
The test creates a dummy Keras model which includes the **KL loss layer**, converts it to an `hls4ml` model 
and synthesises it.

**Test Configuration**:
To run the test do `python contrib/kl_layer.py`

## Checklist


jmitrevs:
I think maybe all of this should go into contrib, since there is no parser. Potentially make a kl_loss directory in there with these two files + a readme to explain.

contrib/kl_layer/README.md
```diff
@@ -0,0 +1 @@
+This folder contains a self-contained example
```
jmitrevs, line 1: Can you add a bit more of a description--a sentence or two about the KL loss--and what the two files are?

katyagovorkova: yeah,sorry, I was still working on it. Will finish asap

jmitrevs:
When I run `python kl_layer.py`, I get:
```
Interpreting Model
  File "/Users/jmitrevs/work/hls4ml/contrib/kl_layer/kl_layer.py", line 189, in <module>
    test_extensions(test_root_path)
  File "/Users/jmitrevs/work/hls4ml/contrib/kl_layer/kl_layer.py", line 168, in test_extensions
    hmodel = hls4ml.converters.convert_from_keras_model(
  File "/Users/jmitrevs/work/hls4ml/hls4ml/converters/__init__.py", line 241, in convert_from_keras_model
    return keras_to_hls(config)
  File "/Users/jmitrevs/work/hls4ml/hls4ml/converters/keras_to_hls.py", line 384, in keras_to_hls
    layer_list, input_layers, output_layers = parse_keras_model(model_arch, reader)
  File "/Users/jmitrevs/work/hls4ml/hls4ml/converters/keras_to_hls.py", line 298, in parse_keras_model
    raise Exception('ERROR: Unsupported layer type: {}'.format(keras_layer['class_name']))
Exception: ERROR: Unsupported layer type: TFOpLambda
```
Does it succeed for you?

vloncar:
@jmitrevs You're probably using TF 2.8 or newer where the information about the custom layer is not embedded in the model when saving it to disk, but rather its computation graph is embedded. So when loading the model back you get these lambda ops. You can try saving and then loading the model back and printing its `summary()` to see if this is the cause. I found no solution for this and was hoping TF reverts to the old functionality (because this one is of dubious utility) in newer release but I didn't check if the latest TF did so.

jmitrevs:
The model that the test seems to parse is:
```
 Layer (type)                   Output Shape         Param #     Connected to                     
==================================================================================================
 input_1 (InputLayer)           [(None, 19, 3, 1)]   0           []                               
                                                                                                  
 dense_1 (Dense)                (None, 19, 3, 10)    20          ['input_1[0][0]']                
                                                                                                  
 dense (Dense)                  (None, 19, 3, 10)    20          ['input_1[0][0]']                
                                                                                                  
 tf.__operators__.add (TFOpLamb  (None, 19, 3, 10)   0           ['dense_1[0][0]']                
 da)                                                                                              
                                                                                                  
 tf.math.square (TFOpLambda)    (None, 19, 3, 10)    0           ['dense[0][0]']                  
                                                                                                  
 tf.math.subtract (TFOpLambda)  (None, 19, 3, 10)    0           ['tf.__operators__.add[0][0]',   
                                                                  'tf.math.square[0][0]']         
                                                                                                  
 tf.math.exp (TFOpLambda)       (None, 19, 3, 10)    0           ['dense_1[0][0]']                
                                                                                                  
 tf.math.subtract_1 (TFOpLambda  (None, 19, 3, 10)   0           ['tf.math.subtract[0][0]',       
 )                                                                'tf.math.exp[0][0]']            
                                                                                                  
 tf.math.reduce_mean (TFOpLambd  (None, 19, 3, 1)    0           ['tf.math.subtract_1[0][0]']     
 a)                                                                                               
                                                                                                  
 tf.math.multiply (TFOpLambda)  (None, 19, 3, 1)     0           ['tf.math.reduce_mean[0][0]']    
                                                                                                  
==================================================================================================
```

jmitrevs:
So I think it is as @vloncar says. Let's see what options we have to proceed.

vloncar:
Maybe add `get_config()` to the Keras implementation and try like that? Also try decorating the function with `tf.keras.utils.register_keras_serializable`

jmitrevs:
Is there anything we can get from the `test_extensions.py`, which still works?

vloncar:
Upon digging a bit more, turns out this implementation is problematic, not the TF version itself. Apparently we shouldn't use `tensorflow.python.keras.*` anymore and go with `tensorflow.keras.*`. I don't even remember why we used the former. With that change the issue is gone. But now the issue is how to import the required base class `_Merge` which has changed locations sometime between 2.7 and the current version (too lazy to investigate exactly when). So I propose we do something like:

```
try:
    from keras.layers.merge import _Merge as Merge
except Exception:
    from keras.layers.merging.base_merge import _Merge
```

contrib/kl_layer/README.md
```diff
@@ -0,0 +1,18 @@
+This folder contains the implementation of custom KL divergence layer.
+This is a custom implementation and not a built-in layer in any deep learning framework.
+It was developed specifically for [AD@L1 CMS paper](https://www.nature.com/articles/s42256-022-00441-3) and works spesifically for the input of size `(19, 3, 1)`.
```
vloncar, line 3: Does it really work `spesifically for the input of size (19, 3, 1)`? I see no hardcoded shapes in the implementation

contrib/kl_layer/kl_layer.py
```diff
@@ -0,0 +1,189 @@
...
+import tensorflow as tf
+from tensorflow.python.keras.layers.merge import _Merge as Merge
+from tensorflow.python.keras.utils import tf_utils
+from tensorflow.python.ops import math_ops
+
+import hls4ml
+from hls4ml.converters.keras_to_hls import parse_default_keras_layer
+from hls4ml.model.types import FixedPrecisionType, NamedType
+
+test_root_path = Path(__file__).parent
```
vloncar, line 21: This entire file is structured like a test, for which there is no reason. Perhaps we remove references to "test" and keep this script clutter-free?

contrib/kl_layer/kl_layer.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+# Keras implementation of a KL layer
+class Distance(Merge):
+    def _check_inputs(self, inputs):
+        if len(inputs) not in [2, 3]:
```
vloncar, line 27: In the end we never used the other losses, so there's no need for `Distance` class itself, nor the check for the number of inputs.

contrib/kl_layer/kl_layer.py
```diff
@@ -0,0 +1,189 @@
...
+    assert 'KLLoss' in keras_layer['class_name']
+
+    layer = parse_default_keras_layer(keras_layer, input_names)
+
+    output_shape = [input_shapes[0][0], 1]
+
+    return layer, output_shape
+
+
+def test_extensions(tmp_path):
```
vloncar, line 128: No reason to call this an "extension test"

contrib/kl_layer/nnet_distance.h
```diff
@@ -0,0 +1,87 @@
...
+    static const unsigned n_out = 1;
+
+    // Internal data type definitions
+    typedef float accum_t;
+    typedef float sum_t;
+    typedef ap_fixed<18, 8> exp_table_t;
+
+    // Internal info
+    static const unsigned table_size = 1024;
+    static constexpr float exp_range = 1024;
```
vloncar, line 23: Does this really need to be float? It would likely have bad QoR if it is not a power-of-two integer.

vloncar (changes requested):

contrib/kl_layer/kl_layer.py
katyagovorkova: ok, test removed

contrib/kl_layer/kl_layer.py
katyagovorkova: ok, `Distance` class removed

contrib/kl_layer/kl_layer.py
katyagovorkova: renamed

contrib/kl_layer/nnet_distance.h
katyagovorkova: what should it be instead of float?

contrib/kl_layer/README.md
katyagovorkova: right. fixed now, thanks

jmitrevs:
I think you need these changes:
```
(fastml39) mac-137349:hls4ml jmitrevs$ git diff
diff --git a/contrib/kl_layer/kl_layer.py b/contrib/kl_layer/kl_layer.py
index 198fb012..318c2f46 100644
--- a/contrib/kl_layer/kl_layer.py
+++ b/contrib/kl_layer/kl_layer.py
@@ -14,7 +14,7 @@ import tensorflow as tf
 try:
     from keras.layers.merge import _Merge as Merge
 except Exception:
-    from keras.layers.merging.base_merge import _Merge
+    from keras.layers.merging.base_merge import _Merge as Merge
     
 from tensorflow.python.keras.utils import tf_utils
 from tensorflow.python.ops import math_ops
@@ -110,7 +110,7 @@ class HKLLossFunctionTemplate(hls4ml.backends.template.FunctionCallTemplate):
 
 
 # Parser for converter
-def parse_klloss_layer(keras_layer, input_names, input_shapes, data_reader, config):
+def parse_klloss_layer(keras_layer, input_names, input_shapes, data_reader):
     assert 'KLLoss' in keras_layer['class_name']
 
```
However, I still get a `KeyError: 'accum_t'` at the format in:
```python
    def format(self, node):
        params = self._default_config_params(node)
        params['n_in'] = node.get_input_variable(node.inputs[0]).shape[0]
        params['n_out'] = 1
        return self.template.format(**params)
```
around line 90 of `kl_layer.py`.

katyagovorkova:
> I think you need these changes:
> 
Fixed, thanks!

> However, I still get a `KeyError: 'accum_t'` at the format in:
> 
> ```python
>     def format(self, node):
>         params = self._default_config_params(node)
>         params['n_in'] = node.get_input_variable(node.inputs[0]).shape[0]
>         params['n_out'] = 1
>         return self.template.format(**params)
> ```
> 
> around line 90 of `kl_layer.py`.

Ah I see, that's most likely because I have removed the Distance class as Vladimir suggested. But seems like it required more changes than just removing the class.. Also I can not test it locally since I have different error when running: `Exception: Optimization pass vivado:clone_output already registered`

vloncar:
I fixed the outstanding issues. Unfortunately, `pre-commit` broke for me so I couldn't run it, hence the noise until I manually made it compliant. So annoying...

Anyway, it is ready now.

vloncar (approved):
Looks good, thanks Katya!

PR #618, opened 2022-07-25
Quartus Synthesis Flow Improvement

bo3z:
# Description

> :memo: Improves the Quartus synthesis flow with more logging, storing resource usage and latency results to files and more robust compilation flags

## Type of change

## Tests

> :memo: A few models were synthesized, in order to verify more logging and the presence of files containing synthesis results.

This PR attempts to do three things in the Quartus synthesis flow:

1. Add a variable level of logging during HLS synthesis, which allows the user to see the exact stage of HLS synthesis, alongside any warnings, intermediate fails, reasons of failure etc.
2. Write HLS synthesis results (resource usage and timing analysis) to a text file, so these can be reused later. Also, a common issue with Quartus synthesis is an exception encountered when reading verification results (latency and ii), which are obtained by running the binary produced by HLS synthesis. If i++ is not available on the host system natively (such as when it is installed in a container, e.g. singularity), running the binary from hls4ml will throw an exception after which the execution of the entire program halts, because verification results are not obtained. This PR addresses this problem by placing the verification inside a try-catch clause as well as providing clear instructions why an exception is encountered.
3. Adds a flag to HLS synthesis to continue compiling if a large area is estimated. Sometimes, the HLS compiler overestimate the resources needed, and fails compilation. Therefore, we allow the compilation to continue (even if it can't fit on the desired FPGA) to perform some analysis. This is a configurable flag and not used by default.

### DIFF
--- a/hls4ml/backends/quartus/quartus_backend.py
+++ b/hls4ml/backends/quartus/quartus_backend.py
@@ -115,35 +115,43 @@ class QuartusBackend(FPGABackend):
         layer.weights['weight'].data_length = layer.weights['weight'].data.size
         return
 
-    def build(self, model, synth=True, fpgasynth=False):
+    def build(self, model, synth=True, fpgasynth=False, log_level=1, cont_if_large_area=False):
         """
         Builds the project using Intel HLS compiler.
 
-        Users should generally not call this function directly but instead use `ModelGraph.build()`.
-        This function assumes the model was written with a call to `ModelGraph.write()`
-
         Args:
             model (ModelGraph): The model to build
-            synth, optional: Whether to run synthesis
-            fpgasynth, optional:  Whether to run fpga synthesis
-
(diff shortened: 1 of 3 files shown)


hls4ml/backends/quartus/quartus_backend.py
```diff
@@ -115,35 +115,43 @@ def gen_quartus_weight_array(self, layer):
         layer.weights['weight'].data_length = layer.weights['weight'].data.size
         return
 
-    def build(self, model, synth=True, fpgasynth=False):
+    def build(self, model, synth=True, fpgasynth=False, logging=1, cont_if_large_area=False):
```
vloncar, line 118: Can we name it `log_level` or something more descriptive? Not to be confused that it is enabled/disabled with 0/1.

bo3z: Agreed.

Issue #624, opened 2022-07-31
convert_from_keras_model with project_name set to nondefault value fails to build


jmitrevs:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out. (62046d79)
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x] If there are steps to reproduce the problem, make sure to write them down below.
- [x] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
convert_from_keras_model with project_name set to nondefault value fails to build, because in `build.tcl` the variable ${mynewprojname} is not defined. I will look to fix this, so just adding issue to not forget.

## Details
If I set project_name to a non_default value, e.g.:
```python
hls_model = hls4ml.converters.convert_from_keras_model(model,
                                                       hls_config=config,
                                                       output_dir='model_1_copy/hls4ml_prj',
                                                       part='xcu250-figd2104-2L-e',
                                                       project_name="mynewmodel")
```
in the hls4ml-tutorial (part 1), the build strep produces:

```
Sourcing Tcl script 'build_prj.tcl'
can't read "mynewmodel": no such variable
    while executing
"open_project ${mynewmodel}_prj"
    invoked from within
"if {$opt(reset)} {
  open_project -reset ${mynewmodel}_prj
} else {
  open_project ${mynewmodel}_prj
}"
    (file "build_prj.tcl" line 151)
    invoked from within
"source build_prj.tcl"
    ("uplevel" body line 1)
    invoked from within
"uplevel \#0 [list source $arg] "
```

Please add to the following sections to describe the bug as accurately as possible.

### Steps to Reproduce

The description above describes the change to part 1 of the hls4ml-tutorial that is legal but causes the project to not build

PR #626, opened 2022-08-01
fix nondefault project name handling

jmitrevs:
# Description

As discussed in #624, convert_from_keras_model with project_name set to nondefault  value (e.g. mynewprojname) fails to build, because in build.tcl the variable ${mynewprojname} is not defined. This fixes the problem my always leaving the variable called `myproject`, but just changing the value of the variable. Some changes in the tcl code were requried since now since the variable name does not always match the value it contains. The build.tcl indentation was fixed, and the parsing of the Vivado report was fixed. Note that this is broken in main even in the default case currently.

## Type of change

## Tests

Testing was done offline by modifying and running part1 of the hls4ml-tutorial, as mentioned in #624. Nevertheless, we probably should add a test. The complication is that the test requires synthesis, and generally our pytests don't do synthesis. Plus, synthesis is long. Do we want to add another synthesis test for this case or modify an existing one?

## Checklist

### DIFF
--- a/hls4ml/report/vivado_report.py
+++ b/hls4ml/report/vivado_report.py
@@ -39,7 +39,7 @@ def _parse_project_script(path):
 
     with open(project_path, 'r') as f:
         for line in f.readlines():
-            if 'set myproject' in line:
+            if 'set project_name' in line:
                 top_func_name = line.split('"')[-2]
                 prj_dir = top_func_name + '_prj'
 
--- a/hls4ml/report/vivado_report.py
+++ b/hls4ml/report/vivado_report.py
@@ -12,8 +12,8 @@ def read_vivado_report(hls_dir, full_report=False):
     prj_dir = None
     top_func_name = None
 
-    if os.path.isfile(hls_dir + '/build_prj.tcl'):
-        prj_dir, top_func_name = _parse_build_script(hls_dir + '/build_prj.tcl')
+    if os.path.isfile(hls_dir + '/project.tcl'):
+        prj_dir, top_func_name = _parse_build_script(hls_dir)
 
     if prj_dir is None or top_func_name is None:
         print('Unable to read project data. Exiting.')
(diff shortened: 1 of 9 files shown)


jmitrevs:
The qkeras pytest failure is a known problem that exists in the main branch even without this change.

thesps:
Thanks for the fix, I don't think we need a new test for this

thesps (approved):

hls4ml/templates/vivado/build_prj.tcl
```diff
@@ -149,19 +149,19 @@ set CSIM_RESULTS "./tb_data/csim_results.log"
...
-  open_project -reset ${myproject}_prj
+    open_project -reset ${myproject}_prj
 } else {
-  open_project ${myproject}_prj
+    open_project ${myproject}_prj
 }
-set_top myproject
-add_files firmware/myproject.cpp -cflags "-std=c++0x"
-add_files -tb myproject_test.cpp -cflags "-std=c++0x"
+set_top ${myproject}
```
vloncar, line 156: If we're using a variable now, why are we still calling it `myproject` and not something more descriptive, like `project_name`?

hls4ml/report/vivado_report.py
```diff
@@ -12,8 +12,8 @@ def read_vivado_report(hls_dir, full_report=False):
     prj_dir = None
     top_func_name = None
 
-    if os.path.isfile(hls_dir + '/build_prj.tcl'):
-        prj_dir, top_func_name = _parse_build_script(hls_dir + '/build_prj.tcl')
+    if os.path.isfile(hls_dir + '/project.tcl'):
+        prj_dir, top_func_name = _parse_build_script(hls_dir)
```
vloncar, line 16: We don't really parse the build script anymore? Can we rename this?

hls4ml/writer/vivado_writer.py
```diff
@@ -529,22 +512,27 @@ def write_build_script(self, model):
         f.write('set myproject "{}"\n'.format(model.config.get_project_name()))
```
vloncar, line 512: `myproject` variable name still survives...

jmitrevs:
The last push was not a complete fix. I'll send a message again when I think it's ready.

hls4ml/report/vivado_report.py
jmitrevs: Done

jmitrevs:
I still need to test vsynth=True and also that I didn't mess up the FIFO optimizations.

jmitrevs:
Actually, looking in the main branch as a starting point, I see 
```
 'VivadoSynthReport': {'LUT': '0',
  'FF': '0',
  'BRAM_18K': '0',
  'URAM': '0',
  'DSP48E': '0'}
```
when I do part1 of the hls4ml-tutorial with `hls_model.build(vsynth=True)` so vsynth might be broken in the main branch. Will try to investigate.

vloncar:
This has been broken for some time now, I observed it as well while playing with Vitis. Thanks for looking into it.

jmitrevs:
This seems to break the VivadoAccelerator backend.

jmitrevs:
I believe this is ready for merging.

vloncar (approved):

PR #628, opened 2022-08-03
Quartus Extensions

bo3z:
A# Description

> :memo: Support for custom Keras and HLS layers
> * Simple re-ordering of optimizers in the Quartus backend allows to register a new layer function/config template, alongside any optimizers linked to it.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) - This should have worked before in a similar manner way to Vivado, but was never tested. This PR simply adds a PyTest and fixed flow requirements to allow custom Keras layers to be introduced.

## Checklist
- [] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/test/pytest/test_extensions.py
+++ b/test/pytest/test_extensions.py
@@ -1,13 +1,12 @@
-import hls4ml
-import tensorflow as tf
-import numpy as np
 import pytest
+import hls4ml
+import numpy as np
+import tensorflow as tf
 from pathlib import Path
 
 test_root_path = Path(__file__).parent
 
 # Keras implementation of a custom layer
-
 class KReverse(tf.keras.layers.Layer):
     ''' Keras implementation of a hypothetical custom layer '''
     def __init__(self):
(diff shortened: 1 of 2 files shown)


test/pytest/test_extensions.py
```diff
@@ -74,8 +97,6 @@ def format(self, node):
     data_T input[CONFIG_T::n_in],
     data_T reversed[CONFIG_T::n_in]
 ) {
-    #pragma HLS PIPELINE
```
vloncar, line 77: Does this line cause problems for Intel HLS compiler? What if we wrap it in `#ifdef __SYNTHESIS__` and do similar for Intel compiler

bo3z: I think it won't (I think there is a flag for ignoring unknown pragmas), but PyTests do not run HLS synthesis anyway, so this line does not really do anything. Only reason I removed it is to make it "technically" correct.

vloncar: That entire file used to be an example of extensions API that also served as a test, so that block is an example of (Vivado) HLS code that users should write themselves. I don't see how distilling it to pure C++ code makes it "tecnhically" correct ;-) But it doesn't matter, since now it is a test that serves as a (slightly convoluted) example.

vloncar (approved):

PR #635, opened 2022-08-10
Fix conv1d stream implementation hls directives

Jonathan-Shoemaker:
# Description

Fixed compiler directives for conv1d io_stream implementation. A loop was given the unroll directive, when it should have been pipelined. This loop is used to shift the kernel data when a new input is read by the stream. The new directives match those used in the conv2d implementation. 

## Type of change

## Tests

The existing tests for conv1d still pass

**Test Configuration**:

To see the improvements of the fix, I just ran synthesis on a large conv1d layer. The layer's parameters in this test are 8 channels, 16 filters, length 4096, with a kernel size of 7 and a stride of 2. The fix reduces the latency from 635,968 cycles to 53,342 cycles when using a reuse factor of 1 and resource strategy.

## Checklist

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv_stream.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv_stream.h
@@ -156,13 +156,13 @@ void kernel_shift_1d(
     typename data_T::value_type kernel_window[CONFIG_T::filt_width * CONFIG_T::n_chan]
 ) {
     #pragma HLS inline
-    #pragma HLS PIPELINE II = 1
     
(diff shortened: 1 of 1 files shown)


vloncar:
Thanks Jonathan. Can you add the before and after synthesis results to demonstrate the change?

Jonathan-Shoemaker:
I attached the synthesis printouts for the example conv layer before and after the change. I also attached a python file used to construct the model in case it is of interest.
[conv_stream_update.zip](https://github.com/fastmachinelearning/hls4ml/files/9310475/conv_stream_update.zip)

vloncar (approved):

PR #639, opened 2022-08-16
Fix parsing of logic synthesis reports

vloncar:
# Description

Due to changes in the format of the Vivado (logic) synthesis, the parsing function is broken and zeros are returned. This fix ensures only the appropriate sections are parsed.

## Type of change

## Tests

```
report = hls4ml.report.parse_vivado_report('/path/to/my/hls/outout/dir')
print(report['VivadoSynthReport'])
```
Observe before and after.


jmitrevs:
I think this needs some of the fixes in #626 in order for part 1 of the hls4ml-tutorial to work. So I would approve it, but only after #626.

jmitrevs (approved):

PR #641, opened 2022-08-22
append applied_flows container before filling instead of after

jmitrevs:
# Description

This change simply appends `applied_flows` to `ModelGraph._applied_flows` in the beginning, before it is filled, so that  `ModelGraph._applied_flows` always has an accurate description of what has been applied. This is important in the FIFO depth optimization, because the `FifoDepthOptimization` runs the writer flow outside of the main flow, before the flow is finished and before `ModelGraph._applied_flows` is updated, so currently it sees `ModelGraph._applied_flows == []`, and since the writer depends on `'vivado:ip'`, all the optimizers are run again, which sometimes messes things up. For example, running on a ResNet sample, rerunning the optimizers changes a type from being `PackedType` to just being `FixedPrecisionType`, causing it to fail. This way,   `ModelGraph._applied_flows` is always current so the out of main flow writer does not run `'vivado:ip'` again.

I also updated a few empty lists to empty sets so that applied_flows is always a dictionary of sets, not a dictionary of sets or empty lists. 

## Type of change

## Tests

The main test is to see that this doesn't break the current pytests. The FIFO optimization is too long to run as a standard test.

## Checklist

### DIFF
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -348,7 +348,7 @@ class ModelGraph(object):
             applied_flows = {}
 
             for flow_group in self._applied_flows:
-                applied_flows.update({flow: [] for flow in flow_group.keys()})
+                applied_flows.update({flow: set() for flow in flow_group.keys()})
 
             return applied_flows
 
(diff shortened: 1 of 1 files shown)


jmitrevs:
There is a change in the fifo depth optimization discussed in in in #642, basically getting rid of the explict, out of flow write, called from the optimizer, that would make this change unnecessary in the case I mentioned, though this change would still not cause any harm.

jmitrevs:
I think this change should go in since it fixes the current fifo depth optimization, and even if we change how we do the optimization, this will have no ill effect.

jmitrevs:
Any more thoughts on this? I have to tell people to use it if they are doing fifo depth optimization, which is a bit annoying.

vloncar (approved):
Looks good to me. Let's hope this doesn't blow up into our faces.

PR #646, opened 2022-08-26
External weights vivado accelerator

GiuseppeDiGuglielmo:
Extend the `VivadoAccelerator` backend and add programmable weights. An ideal setup is a Xilinx Zynq/ZynqMP board (ARM core + programmable logic).

The backend generates code for Vivado HLS, Vivado, and Vivado SDK:
- the **hls4ml IP** with AXI master interfaces to move input features, output predictions, and _weights_ between off-chip RAM and FPGA chip
- a complete **Vivado project** that integrates the hls4ml IP (for the target board/chip)
- a **complete baremetal application** to control and program the accelerator

## Type of change

Some changes to the signature of the function `convert_from_keras_model()`:

```
    hls_model =  hls4ml.converters.convert_from_keras_model(
            model=model,
            clock_period=CLOCK_PERIOD,
            backend='VivadoAccelerator',
            board=BOARD_NAME,
            part=FPGA_PART,
            io_type='io_stream',
            interface='axi_master',
            driver='c',
            input_data_tb=DATA_DIR+'/X_test.npy',
            output_data_tb=DATA_DIR+'/y_test.npy',
            hls_config=config,
            output_dir=OUTPUT_DIR)
```

and a new function `write_header_file()` to write a header file with an harcoded dataset:

```
hls4ml.writer.vivado_accelerator_writer.VivadoAcceleratorWriter.write_header_file(hls_model, X_test, y_test, y_qkeras, y_hls, 64, OUTPUT_DIR + '/sdk/common/data.h')
```

## Tests

You can test it with the example at this repo: https://github.com/GiuseppeDiGuglielmo/test-hls4ml-backend

Right now, we support Ultra96v2, but more Zynq/ZynqMP boards can be added.

### DIFF
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -606,7 +606,7 @@ class ModelGraph(object):
     def _get_top_function(self, x):
 
         io_type = self.config.get_config_value('IOType')
-        interface = self.config.get_config_value('AcceleratorConfig')['Interface']
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface'] if self.config.get_config_value('AcceleratorConfig') else None
         config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
 
         if self._top_function_lib is None:
@@ -662,7 +662,7 @@ class ModelGraph(object):
     def predict(self, x):
 
         io_type = self.config.get_config_value('IOType')
-        interface = self.config.get_config_value('AcceleratorConfig')['Interface']
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface'] if self.config.get_config_value('AcceleratorConfig') else None
         config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
 
         top_function, ctype = self._get_top_function(x)
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -604,6 +604,11 @@ class ModelGraph(object):
         self._top_function_lib = ctypes.cdll.LoadLibrary(lib_name)
 
     def _get_top_function(self, x):
+
+        io_type = self.config.get_config_value('IOType')
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface']
+        config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
+
         if self._top_function_lib is None:
             raise Exception('Model not compiled')
         if len(self.get_input_variables()) == 1:
@@ -611,7 +616,8 @@ class ModelGraph(object):
         else: 
             xlist = x
         n_outputs = len(self.get_output_variables())
-        
+        n_weights = len(self.get_weight_variables())
+
         for xi in xlist:
             if not isinstance(xi, np.ndarray):
                 raise Exception('Expected numpy.ndarray, but got {}'.format(type(x)))
@@ -628,9 +634,9 @@ class ModelGraph(object):
         else:
             raise Exception('Invalid type ({}) of numpy array. Supported types are: single, float32, double, float64, float_.'.format(x0.dtype))
 
-
         top_function.restype = None
-        top_function.argtypes = [npc.ndpointer(ctype, flags="C_CONTIGUOUS") for i in range(len(xlist) + n_outputs)]
+        top_function.argtypes = [npc.ndpointer(ctype, flags="C_CONTIGUOUS") \
+                for i in range(len(xlist) + (n_weights if config_weights else 0) + n_outputs)]
 
         return top_function, ctype
 
@@ -654,10 +660,16 @@ class ModelGraph(object):
         return int(n_sample)
 
     def predict(self, x):
+
+        io_type = self.config.get_config_value('IOType')
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface']
+        config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
+
         top_function, ctype = self._get_top_function(x)
         n_samples = self._compute_n_samples(x)
         n_inputs = len(self.get_input_variables())
         n_outputs = len(self.get_output_variables())
+        n_weights = len(self.get_weight_variables())
 
         curr_dir = os.getcwd()
         os.chdir(self.config.get_output_dir() + '/firmware')
@@ -675,10 +687,16 @@ class ModelGraph(object):
                     inp = [np.asarray(xj[i]) for xj in x]
                 argtuple = inp
                 argtuple += predictions
+                if config_weights:
+                    for j in range(n_weights):
+                        weights = [float(w) for w in self.get_weight_variables()[j]]
+                        argtuple += [np.asarray(weights)]
                 argtuple = tuple(argtuple)
                 top_function(*argtuple)
-                output.append(predictions)
-
+                if config_weights and n_samples == 1 and n_inputs:
+                    output.append([predictions])
+                else:
+                    output.append(predictions)
 
             # Convert to list of numpy arrays (one for each output)
             output = [np.asarray([output[i_sample][i_output] for i_sample in range(n_samples)]) for i_output in range(n_outputs)]
--- a/hls4ml/model/optimizer/passes/nop.py
+++ b/hls4ml/model/optimizer/passes/nop.py
@@ -6,7 +6,7 @@ class EliminateLinearActivation(OptimizerPass):
         cast = False
         if isinstance(node, Activation):
             cast = node.get_input_variable().type.precision != node.get_output_variable().type.precision
-        return isinstance(node, Activation) and node.get_attr('activation') == 'linear' and not cast
+        return isinstance(node, Activation) and node.get_attr('activation') == 'linear' # and not cast
     
     def transform(self, model, node):
         model.remove_node(node)
--- a/hls4ml/templates/vivado/myproject_test.cpp
+++ b/hls4ml/templates/vivado/myproject_test.cpp
@@ -22,8 +22,9 @@
 #include <vector>
 #include <map>
 #include <stdio.h>
-#include <stdlib.h>
-#include <math.h>
+#include <cstdlib>
+#include <cmath>
+#include <cfloat>
 
 #include "firmware/myproject.h"
 #include "firmware/nnet_utils/nnet_helpers.h"
@@ -55,6 +56,10 @@ int main(int argc, char **argv)
   std::string iline;
   std::string pline;
   int e = 0;
+
+  //hls-fpga-machine-learning insert weights
+
+  //hls-fpga-machine-learning insert load weights
 
   if (fin.is_open() && fpr.is_open()) {
     while ( std::getline(fin,iline) && std::getline (fpr,pline) ) {
--- a/hls4ml/templates/vivado_accelerator/ultra96v2/c_drivers/sdk/Makefile
+++ b/hls4ml/templates/vivado_accelerator/ultra96v2/c_drivers/sdk/Makefile
@@ -1,0 +1,33 @@
+DESIGN := design_1
+
+help:
+	@echo "INFO: make <TAB> to show targets"
+.PHONY: help
+
+--setup:
+	xsct ./setup.tcl $(DESIGN)
+.PHONY: --setup
+
+sdk: --setup
+	rm -f $(DESIGN)_standalone/src/helloworld.c
+	cd  $(DESIGN)_standalone/src && ln -s ../../common/main.c main.c
+	cd  $(DESIGN)_standalone/src && ln -s ../../common/data.h data.h
+.PHONY: sdk
+
+gui:
+	xsdk --workspace . &
+.PHONY: gui
+
+clean:
+	rm -rf $(DESIGN)_platform
+	rm -rf $(DESIGN)_standalone
+	rm -rf $(DESIGN)_standalone_bsp
+	rm -rf RemoteSystemsTempFiles
+	rm -rf .Xil
+	rm -rf .metadata
+	rm -f *.log
+.PHONY: clean
+
+ultraclean: clean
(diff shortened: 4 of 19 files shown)


thesps (changes requested):
Thanks for this, that's a nice feature! I have some general feedback here and some specific points that I'll address inline.

- Can we decouple ‘repgrommable weights’ from ‘AXI Master interface’? I’d like to be able to do both:
— AXI Master interface without reprogrammable weights
— AXI Stream interface with reprogrammable weights (maybe only the NN in/out on AXI Stream interface while weights use AXI Master?)
- Can we have a C driver that has some function taking the data as argument and returning the NN prediction? And some other function to reprogram the weights. We can call those from the standalone example that puts the data in header files, but they would be helpful for anyone integrating an hls4ml NN in a real application
- Can we have a Python driver for this? - I had some pynq Python for an AXI Master interface before, so I can help with this
- Can we have the design for pynq-z2 as well? - I can help with this also

It might be easier to split these contributions into different PRs, roughly:
- AXI Master interface
- C Driver
- Ultra96 support
- Reprogrammable weights

Also, Xilinx SDK is a legacy tool, superseded by Vitis. Can we do it with Vitis instead?

hls4ml/model/graph.py
```diff
@@ -604,14 +604,20 @@ def compile(self):
         self._top_function_lib = ctypes.cdll.LoadLibrary(lib_name)
 
     def _get_top_function(self, x):
+
+        io_type = self.config.get_config_value('IOType')
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface'] if self.config.get_config_value('AcceleratorConfig') else None
+        config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
```
thesps, line 610: Two things:
- I don't think IOType of io_stream and Interface of axi_master should be the trigger for reprogrammable weights, I think it would be better as a new configuration parameter (of the AcceleratorConfig). Then there would probably need to be some `assert` somewhere to allow only combinations of those other parameters that have been implemented for (ie right now it also depends on the board)
- As written this can't go here in `ModelGraph` as it's backend specific

hls4ml/model/graph.py
```diff
@@ -654,10 +660,16 @@ def _compute_n_samples(self, x):
         return int(n_sample)
 
     def predict(self, x):
+
+        io_type = self.config.get_config_value('IOType')
+        interface = self.config.get_config_value('AcceleratorConfig')['Interface'] if self.config.get_config_value('AcceleratorConfig') else None
+        config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
```
thesps, line 666: Same comment as above

hls4ml/model/optimizer/passes/nop.py
```diff
@@ -6,7 +6,7 @@ def match(self, node):
         cast = False
         if isinstance(node, Activation):
             cast = node.get_input_variable().type.precision != node.get_output_variable().type.precision
-        return isinstance(node, Activation) and node.get_attr('activation') == 'linear' and not cast
+        return isinstance(node, Activation) and node.get_attr('activation') == 'linear' # and not cast
```
thesps, line 9: ? I don't think this should be included in this PR

hls4ml/templates/vivado/myproject_test.cpp
```diff
@@ -22,8 +22,9 @@
 #include <vector>
 #include <map>
 #include <stdio.h>
-#include <stdlib.h>
-#include <math.h>
+#include <cstdlib>
+#include <cmath>
+#include <cfloat>
```
thesps, line 27: ?

hls4ml/templates/vivado_accelerator/ultra96v2/c_drivers/sdk/Makefile
```diff
@@ -0,0 +1,33 @@
...
+.PHONY: help
+
+--setup:
+	xsct ./setup.tcl $(DESIGN)
+.PHONY: --setup
+
+sdk: --setup
+	rm -f $(DESIGN)_standalone/src/helloworld.c
+	cd  $(DESIGN)_standalone/src && ln -s ../../common/main.c main.c
+	cd  $(DESIGN)_standalone/src && ln -s ../../common/data.h data.h
```
thesps, line 14: I didn't try running anything yet, but it seems like these files might not have the same names in the driver & standalone example

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -17,6 +17,10 @@ def write_axi_wrapper(self, model):
         inp_axi_t, out_axi_t, inp, out = self.vivado_accelerator_config.get_corrected_types()
         indent = '    '
 
+        io_type = model.config.get_config_value('IOType')
+        interface = model.config.get_config_value('AcceleratorConfig')['Interface'] if model.config.get_config_value('AcceleratorConfig') else None
+        config_weights = (io_type == 'io_stream') and (interface == 'axi_master')
```
thesps, line 22: Same comment as before that I think a different trigger for reprogrammable weights is needed

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -87,6 +99,11 @@ def write_axi_wrapper(self, model):
                 newline = 'void {}_axi(\n'.format(model.config.get_project_name())
             elif '//hls-fpga-machine-learning insert include' in line:
                 newline = '#include "{}_axi.h"\n'.format(model.config.get_project_name())
+            elif config_weights and '//hls-fpga-machine-learning insert weights' in line:
+                newline = ''
+                for v in model.get_weight_variables():
+                    newline += indent + ', model_axi_t {name} [{shape}]\n'.format(name=v.name, shape=v.data_length)
+                newline += indent + ', char load_weights'
```
thesps, line 106: Why not `bool` for `load_weights`?

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -348,14 +416,129 @@ def write_board_script(self, model):
...
+            copytree(srcfiles, dstfiles, dirs_exist_ok=True)
+        else:
+            copyfile(srcfiles, dstfiles)
+
     def write_new_tar(self, model):
         os.remove(model.config.get_output_dir() + '.tar.gz')
         super(VivadoAcceleratorWriter, self).write_tar(model)
 
-        
+    def write_standalone_app(self, model):
```
thesps, line 430: Maybe the standalone app should be put in the examples instead?

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -348,14 +416,129 @@ def write_board_script(self, model):
...
+                newline += indent + 'XMyproject_axi_Start(&accelerator); /* TODO: design-dependent name */\n'
+                newline += indent + 'while (!XMyproject_axi_IsDone(&accelerator)); /* TODO: design-dependent name */\n'
+            else:
+                newline = line
+            fout.write(newline)
+
+        f.close()
+        fout.close()
+
+    def write_header_file(model, X, y, y_keras, y_hls, n_samples, filename='data.h'):
```
thesps, line 471: Related to the comment above - maybe this should be an example?

thesps (changes requested):

hls4ml/model/optimizer/passes/nop.py
jmitrevs: That was a quick and dirty hack to get some models to optimize better, but it really was meant only for that branch, where we had checked the correctness. I agree that it should not be in the PR.

GiuseppeDiGuglielmo:
@thesps thank you for all of the comments that help a lot!

I will reply both here and to the other inlined questions.

>  Can we decouple ‘repgrommable weights’ from ‘AXI Master interface’? I’d like to be able to do both:
>    — AXI Master interface without reprogrammable weights
>    — AXI Stream interface with reprogrammable weights (maybe only the NN in/out on AXI Stream interface while weights use AXI Master?)

We need the greater flexibility that you suggest. I agree.

Do we have a matrix or list of the existing possible combinations (_what goes with what_)?

We may also define how to pass this configuration information to the function `convert_from_keras_model()`. I am unsure if we should keep adding function parameters or if some nested struct/dictionary would be better. We may discuss this as a very first step because you have a better global and local view of the current status.

>  Each weight of the model will have its own top-level interface/port. How about a mode with a single 'weights' port with one type that can be configured (so could be some ap_fixed or float, double for example) with casting to the proper weights types in the HLS. This would be similar to what we allow now for inputs/outputs with VivadoAccelerator

We can fine-tune the weights to properly-sized ports. Not sure if that would significantly increase the logic. As you noticed, for simplicity, we bring in all the weights as `float` (32b) and convert them in the accelerator to the expected fixed precision.

>  Can we have a C driver that has some function taking the data as argument and returning the NN prediction? And some other function to reprogram the weights. We can call those from the standalone example that puts the data in header files, but they would be helpful for anyone integrating an hls4ml NN in a real application

Definitely. The current software application was mostly for proof-of-concept and debugging. We can break it down into a more organic library.

>  Can we have a Python driver for this? - I had some pynq Python for an AXI Master interface before, so I can help with this

Yes. Python would require an OS (or a lighter-weight RTOS). We are not booting an OS right now, so a C/C++ bare-metal application was sufficient as we had for the MLPerf Tiny submission.

I will comment a little more about this question.

>  Can we have the design for pynq-z2 as well? - I can help with this also

Definitely.

I have a few more boards in my working directory that I did not push yet. Essentially there are three classes of chips/boards:
- "Pure FPGA" that for us are Microblaze-based systems (see MLPerf Tiny submission ort ARTY)
- Zynq-based systems (like pynq-z2, pynq-z1, etc)
- ZynqMP-based systems (like Ultra96v2, ZCU106, ZCU102, etc)

> It might be easier to split these contributions into different PRs, roughly:
>
>    AXI Master interface
>    C Driver
>    Ultra96 support
>    Reprogrammable weights

I like the breakdown.

> Also, Xilinx SDK is a legacy tool, superseded by Vitis. Can we do it with Vitis instead?

This is a good question, and @jmitrevs brought it up as well. 

I did not use Vitis seriously enough, and I assume it may come with a similar flow and equivalent tools (Vivado HLS -> Vivado -> Vivado SDK). Given Vitis, I am unsure how much we should push on the software side. Do you have any experience with the software stack in Vitis? In the longer term, I would also like to close the loop and create software applications/library for Linux. I started [something](https://github.com/GiuseppeDiGuglielmo/hls4ml-udma) based on Petalinux etc. but that once again may overlap what Vitis is doing.

Similarly, there is an ongoing effort to use the PYNQ software stack and overlays. I am wondering if we should look at that as well (I do not have experience with it, but I guess there should a good degree of compatibility).

hls4ml/model/graph.py
GiuseppeDiGuglielmo: I agree. This was mostly a placeholder, and I was going to ask what should be better.

As in the main thread, do we already have a matrix or list of the existing possible combinations (what goes with what)? I am not sure how to properly set up the configuration/trigger. Do you have suggestions?

hls4ml/writer/vivado_accelerator_writer.py
GiuseppeDiGuglielmo: It came from a previous iteration. For a cleaner/safer C++ code, `bool` is the right choice. In hardware, it would become a single-bit register if that was just in the module, while on an AXI interface you can only use 8, 16, 32, etc sized words.

hls4ml/writer/vivado_accelerator_writer.py
GiuseppeDiGuglielmo: This function and the `write_header_file` can be either part of the example or come with the hls4ml code base.

hls4ml/writer/vivado_accelerator_writer.py
GiuseppeDiGuglielmo: I am good with moving those functions to the examples, and later on, we can "promote" them to the hls4ml codebase if necessary.

hls4ml/model/graph.py
GiuseppeDiGuglielmo: I started putting together a [document](https://fermicloud-my.sharepoint.com/:w:/g/personal/gdg_services_fnal_gov/ESO16EampBpBm27vlWeX2jwB8WNf1V25lRKyQOIXkY2EDQ?e=6fMzqV), which is far from being final. @thesps and @jmitrevs should be able to edit. Not sure if that may help or if it is better to keep it to the comments here. I bit of both I guess.

If [I understand it right](https://github.com/fastmachinelearning/hls4ml/blob/main/docs/api/configuration.rst) the additional configuration parameters (to enable programmable weights) should be passed via the function [create_initial_config](https://github.com/fastmachinelearning/hls4ml/blob/563c84c1ddfe5d8298a38ce2cdd74881ac8b955d/hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py#L62) for the `VivadoAccelerator` backend.

If you agree, then I would add:
- `configurable_weights`
    - `bool`, optional
    - weights are configurable at run time and thus they are exposed to the wrapper interface 
    - defaults to `False`
- `weight_type`
    - `dict`, optional 
    - a dictionary that specifies the data type for set of layer weights (can be `float` or an `ap_type`); if the type is not specified for a layer, then it defaults to `float`
    - defaults to an empty dictionary that is all of the weights are `float`
    - note: VivadoAcceleratorBackend will round the number of bits used to the next power-of-2 value. 

Please let me know if that makes sense.

GiuseppeDiGuglielmo:
> * AXI Master interface
> 
> * C Driver
> 
> * Ultra96 support
> 
> * Reprogrammable weights

Breaking down this draft in 4 PRs.

Let's start with [PR653](https://github.com/fastmachinelearning/hls4ml/pull/653)

PR #660, opened 2022-09-30
Symbolic expressions in hls4ml

vloncar:
# Description

This PR adds the ability to generate HLS code to evaluate symbolic expressions in hls4ml for people wanting to play with symbolic regression. It is very simple, it is implemented as a simple backend, leveraging everything from Vivado. The math operations come from `hls_math.h`. Expression handling itself and producing code is done with SymPy. There are still many things to improve, like the precision handling, accumulator types, using our lookup table-based implementations of math functions etc. But most of the expressions produced by PySR already work.

## Type of change

## Tests

The main function to convert is called `convert_from_symbolic_expression`, it takes a sympy expression, or a string from which a sympy expression can be built. This can also come from a `PySRRegressor`. For example:
```
# Create expression for conversion:
# 1) From PySRRegressor model
model = PySRRegressor(...)
model.fit(...)
expr = model.sympy()
# 2) From string
expr = sympy.parsing.sympy_parser.parse_expr('x0**2 + 2.5*cos(x1) - 0.5')
# 3) Build from sympy symbols
x0, x1 = sympy.symbols('x0 x1')
expr = x0**2 + 2.5*cos(x1) - 0.5
# Convert to hls4ml ModelGraph
hls_model = hls4ml.converters.convert_from_symbolic_expression(expr, n_symbols=2, precision='ap_fixed<x,y>')
# Now it's the usual
hls_model.write()
hls_model.compile()
hls_predictions = hls_model.predict(...)
hls_model.build(...)
```

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas. - I actually didn't do this
- [ ] I have made corresponding changes to the documentation. -> Will come
- [ ] I have added tests that prove my fix is effective or that my feature works. -> Will come


test/pytest/test_sr.py
```diff
@@ -0,0 +1,73 @@
...
+    y_hls = y_hls.reshape(y.shape)
+
+    np.testing.assert_allclose(y, y_hls, rtol=1e-2, atol=1e-2, verbose=True)
+
+
+def test_pysr_luts(data):
+    try:
+        from pysr import PySRRegressor
+    except ImportError:
+        warnings.warn(UserWarning('Failed to import PySR, test will be skipped.'), stacklevel=1)
```
jmitrevs, line 50: I am worried that this makes this test too easy to ignore.

jmitrevs: Note in the pytest:

```
=============================== warnings summary ===============================
test/pytest/test_sr.py::test_pysr_luts
  /builds/fastmachinelearning/hls4ml/test/pytest/test_sr.py:50: UserWarning: Failed to import PySR, test will be skipped.
    warnings.warn(UserWarning('Failed to import PySR, test will be skipped.'), stacklevel=1)
test/pytest/test_sr.py::test_pysr_luts
  /home/hls4ml/miniconda/envs/hls4ml-testing/lib/python3.8/site-packages/_pytest/python.py:199: PytestReturnNotNoneWarning: Expected None, but test/pytest/test_sr.py::test_pysr_luts returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    warnings.warn(
```

vloncar: The alternative is to have a full installation of PySR. This can be tricky, since we would need to update the underlying docker image to have it (and Julia and all dependencies). Furthermore, this test actually takes time to produce meaningful results, it's basically training. And we need to do it, since we would ideally have to check if the result uses the LUTs we supplied (an error in the implementation may make the function output nonsense, thus making PySR avoid using it)

as for the warning, that's a good catch, i'll fix the test.

vloncar: I changed the test to skip it if pysr cannot be imported. There's a nice feature of pytest specifically for these things, I didn't know about it. Now there's a clean message: `SKIPPED [1] test_sr.py:49: Failed to import PySR, test will be skipped.` which doesn't affect the overall outcome, but still flags the test in a special way in the final report.

jmitrevs: Do we want to change the setup so that it's not skipped in our standard testing? But that can be in a different PR. I'll look more closely tomorrow since today is a holiday.

vloncar: If we don't want it skipped, we need to change the environment of the container to include this dependency, which IMO should be a separate PR.

jmitrevs: I would recommend making a separate PR to change the environment, but for this PR, this is good.

jmitrevs (approved):

PR #668, opened 2022-10-24
update development version

jmitrevs:
# Description

The fact that the development branch has a lower version number than our release seems to confuse people. Therefore, I suggest we update the version number of the main branch. The actual version number I am requesting is just a proposal, not necessary the best choice. (I noticed qonnx automates this--maybe something like that would be better.) I just wanted to bring it up.

## Type of change

- [x] Other (Specify)  -- metadata

## Tests

None, other than trying to do `pip install -e .` in my work area.

## Checklist

### DIFF
--- a/hls4ml/__init__.py
+++ b/hls4ml/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = '0.5.1'
+__version__ = '0.7.dev0'
 
 from hls4ml import converters
 from hls4ml import report


thesps:
In `conifer` I started using `setuptools_scm` ([here](https://pypi.org/project/setuptools-scm/)) that extracts a version number from git. So that will include tags, but also each commit on main branch will automagically get a unique version number incrementing from the previous tag (like `.devN` at the end). It's quite nice and easy to use.

e.g. https://github.com/thesps/conifer/blob/master/setup.py#L9-L10 and https://github.com/thesps/conifer/blob/master/conifer/__init__.py

jmduarte:
Just a reminder about this PR: https://github.com/fastmachinelearning/hls4ml/pull/479

I can rebase it if needed

jmitrevs:
I'll close this and switch to #479, which solves this issue in a better way.

PR #669, opened 2022-10-26
Fix version extraction in Sphinx config

vloncar:
# Description

With the merging of #479 the location of `__version__` changed and Sphinx config still tries to extract it from `hls4ml/__init__.py`. This changes the config of Sphinx to extract the version information from the new `_version.py`.

## Type of change

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,6 +14,7 @@ import os
 import sys
 sys.path.insert(0, os.path.abspath('../'))
 
+import datetime
 from setuptools_scm import get_version
 
 # -- Project information -----------------------------------------------------
@@ -23,7 +24,7 @@ copyright = str(datetime.datetime.now().year)+', Fast Machine Learning Lab'
 author = 'Fast Machine Learning Lab'
 
 # The full version, including alpha/beta/rc tags
-release = get_version()
+release = get_version(root='..', relative_to=__file__)
 
 # -- General configuration ---------------------------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,14 +14,7 @@ import os
 import sys
 sys.path.insert(0, os.path.abspath('../'))
 
-import datetime
-def get_version(rel_path):
-    for line in open(rel_path):
-        if line.startswith('__version__'):
-            delim = '"' if '"' in line else "'"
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -30,7 +30,7 @@ def get_version(rel_path):
 author = 'Fast Machine Learning Lab'
 
 # The full version, including alpha/beta/rc tags
-release = get_version("../hls4ml/__init__.py")
+release = get_version("../hls4ml/_version.py")
```
jmduarte, line 33: @vloncar do you know if this works when building the docs in the GitHub action? `_version.py` is only generated when installing hls4ml. I'm wondering if we need to `pip install -e .` in the workflow yaml.

Actually, do you know if the version has any effect on the generated docs? I thought it may be shown on the sidebar

vloncar: I was under the impression that the workflow would install the package it checked out, given that the [pre-build](https://github.com/fastmachinelearning/hls4ml/blob/main/.github/workflows/build-sphinx.yml#L17) command installs the dependencies. I could be wrong though.

vloncar: After reading a bit more, I realized the proper way would be to use `get_version()` from `setuptools_scm` itself, rather than rolling our own function for extracting version.

jmduarte:
@vloncar what a nightmare trying to get the `sphinx-action` GitHub action to work. Let me summarize:

- We need `git` available in the image because that's how `setuptools_scm` gets the version (through `git tag` etc.)
- It turns out that because https://github.com/ammaraskar/sphinx-action is not really being maintained right now, you can't install `git` via `apt install git` as part of the `pre-build-command` step due to a variety of other issues (outdated image; trying to install using `os.system()` from python, etc. see e.g. https://github.com/ammaraskar/sphinx-action/issues/33).** There are proposed solutions, but because it's not being maintained, nobody is accepting the PRs, etc.
- I forked it and just installed `git` in the `Dockerfile`

Now it works.

jmduarte (approved):

vloncar:
@jmduarte I fixed the failing test of reading h5 file. It failed in quartus for unrelated reason, due to the name of the project not being propagated. Ready to (squash and) merge now?

PR #678, opened 2022-11-02
Adiabatically turn on `pre-commit`

jmduarte:
Adiabatically turn on `pre-commit`. The idea is that `pre-commit` is only run on files touched in a PR. Simultaneously we can open PRs to run `pre-commit` on files not often updated (or that just aren't touched by any open PRs). Eventually, we can run it on the whole repo.

## Changes
* Add `pre-commit` config
* Add `clang-format` config (@vloncar which style do you prefer? This is basically LLVM style stolen from [scikit-hep/boost-historgram](https://github.com/scikit-hep/boost-histogram))
* Use [`p-clang-format`](https://github.com/MedicineYeh/p-clang-format) to keep pragmas indented at the right level
* Add `pre-commit` GitHub action that should only operate on files touched in the PR

Note the con that this may make it harder to review PRs (as a lot of changes will just be style). To be discussed if that's an issue.

To see what it would look like if we ran this on the whole repo in one go, you can check out this branch and do (on Linux):
```bash
pre-commit install 
pre-commit run --all-files
```
I say on Linux, because there seems to be some `sed` issue with OSX.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
     stages: [manual]
 
 - repo: https://github.com/jmduarte/p-clang-format
-  rev: "v1.0.0"
+  rev: "v1.0.2"
   hooks:
     - id: p-clang-format
       types_or: [c++, c, cuda]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,8 @@ repos:
   hooks:
   - id: black
     language_version: python3
-    args: [--line-length=125]
+    args: ['--line-length=125',
+           '--skip-string-normalization']
 
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
     stages: [manual]
 
 - repo: https://github.com/jmduarte/p-clang-format
-  rev: "v1.0.2"
+  rev: "v1.0.3"
   hooks:
     - id: p-clang-format
       types_or: [c++, c, cuda]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,58 @@
+repos:
+- repo: https://github.com/psf/black
+  rev: 22.10.0
+  hooks:
+  - id: black
+    language_version: python3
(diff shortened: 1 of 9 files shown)


jmduarte:
@vloncar what do you think about this approach?

vloncar:
I think this is great. I still didn't get around to check how does the formatted C++ look like.

I guess we should also note in the documentation about contributing that users are advised to run `pre-commit` before making PRs, it will save us a fixing commits on many new PRs.

jmduarte:
@vloncar to see the diff if we do `pre-commit --run all` on this PR you can look here: https://github.com/jmduarte/hls4ml/compare/pre-commit...jmduarte:hls4ml:pre-commit-run-all

vloncar:
This looks great! Is there any way to keep some of the current conventions? I understand black's preference, but the codebase uses single quotes everywhere in python, so majority of the changes are related to that. So maybe we add `--skip-string-normalization`? I have a slight preference for that. What do others think?

In HLS code, can we use 4 spaces instead of two (`IndentWidth`) and the line limit (`ColumnLimit`) that matches the Python one (125)? The rest of the C++ formatting looks ok.

jmduarte:
@vloncar updated as you requested and new diff is here: https://github.com/jmduarte/hls4ml/compare/pre-commit...jmduarte:hls4ml:pre-commit-run-all

vloncar:
Thanks, I'm liking this one a lot. Is there a way to exclude some C++ files? I'm thinking the contents of ap_fixed and ac_fixed. These are (mostly) copied from the vendors, perhaps we should keep them intact, as it would be easier to track any future changes to them.

jmduarte:
yes, let me add that. i was thinking the same.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,61 @@
+exclude: ^hls4ml\/templates\/(vivado|quartus)\/(ap_types|ac_types)\/
```
jmduarte, line 1: @vloncar now excluding `ap_types` and `ac_types`

vloncar:
Anyone else with their style preferences? This is the perfect bikeshedding PR :smile: If not, I propose we merge it.

jmduarte:
@vloncar no complaints, so i say we merge :)

vloncar (approved):

vloncar:
And now, the avalanche of changes begins :mountain_snow:

PR #682, opened 2022-11-17
Delete `example-prjs` directory

jmduarte:
Delete `example-prjs` directory as I don't believe we use it for anything.

## Type of change

- [x] Other (Specify): cleanup

## Checklist


vloncar:
YES! This has remained in the codebase for no useful reason and should be deleted. I do think that we should have examples, perhaps something that goes beyond the tutorial (like custom models with extension API, examples of tweaks etc)

vloncar (approved):

PR #688, opened 2022-11-28
Add causal padding

cgutsche:
This code implements the "causal" padding for Keras Conv1D layers. This padding only adds leading zeros at the left of the input. This means, CNNs only look into the past but can not see into the future for temporal ordered data points.

The implementation is rather easy:
Instead of setting the left and right padding to (kernel size - 1) // 2 just set the left padding to (kernel size - 1) and the right padding to 0. To enable causal padding, also a .py file in backends is changed so that padding is activated if padding is "causal"

Also a simple test file is added.


vloncar (approved):
Thanks Christian!

PR #690, opened 2022-12-02
New config_from_keras_model

vloncar:
# Description

This PR introduces the new `config_from_keras_model` function that fixes the shortcomings of the old one, namely the need to keep in sync the supported layers and their configurable attributes. It relies on the attribute system to automatically discover tunable attributes of each layer and include them in the config. Config is now more expressive, and includes backend-specific parameters.

## Type of change

## Tests

Current tests should suffice

## Checklist

- [x] I have installed and run `pre-commit` on the files I edited or added. -> As discussed, this will be done after review

### DIFF
--- a/hls4ml/backends/fpga/fpga_backend.py
+++ b/hls4ml/backends/fpga/fpga_backend.py
@@ -1,21 +1,29 @@
-import numpy as np
 import math
 import os
+import re
 from bisect import bisect_left
 from collections.abc import Iterable
-import re
+
+import numpy as np
 
 from hls4ml.backends.backend import Backend
-from hls4ml.model.layers import Layer, Dense, BatchNormalization, Activation, Softmax, Conv1D, Conv2D, SeparableConv1D, SeparableConv2D, Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D, SimpleRNN, LSTM, GRU, Dot, Embedding, GarNet, GarNetStack
-from hls4ml.model.attributes import ChoiceAttribute, ConfigurableAttribute, TypeAttribute
-from hls4ml.model.types import NamedType, IntegerPrecisionType, FixedPrecisionType, XnorPrecisionType, ExponentPrecisionType, RoundingMode, SaturationMode
-from hls4ml.writer import get_writer
+from hls4ml.model.attributes import (ChoiceAttribute, ConfigurableAttribute,
+                                     TypeAttribute)
+from hls4ml.model.layers import (GRU, LSTM, Activation, BatchNormalization,
+                                 Conv1D, Conv2D, Dense, Dot, Embedding, GarNet,
+                                 GarNetStack, GlobalPooling1D, GlobalPooling2D,
+                                 Pooling1D, Pooling2D, SeparableConv1D,
+                                 SeparableConv2D, SimpleRNN, Softmax)
 from hls4ml.model.optimizer import model_optimizer
+from hls4ml.model.types import (ExponentPrecisionType, FixedPrecisionType,
+                                IntegerPrecisionType, RoundingMode,
+                                SaturationMode, XnorPrecisionType)
+from hls4ml.writer import get_writer
 
 
 class FPGABackend(Backend):
     def __init__(self, name):
-        super(FPGABackend, self).__init__(name)
+        super().__init__(name)
 
         self.writer = get_writer(self.name)
 
@@ -23,9 +31,17 @@ class FPGABackend(Backend):
 
         accum_layers = [
             Dense,
-            Conv1D, Conv2D, SeparableConv1D, SeparableConv2D,
-            Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D,
-            SimpleRNN, LSTM, GRU,
+            Conv1D,
+            Conv2D,
+            SeparableConv1D,
+            SeparableConv2D,
+            Pooling1D,
+            Pooling2D,
+            GlobalPooling1D,
+            GlobalPooling2D,
+            SimpleRNN,
+            LSTM,
+            GRU,
             Dot,
         ]
 
@@ -34,12 +50,7 @@ class FPGABackend(Backend):
             attrs.append(TypeAttribute('accum'))
             self.attribute_map[layer] = attrs
 
-        rf_layers = accum_layers + [
-            BatchNormalization,
-            Activation,
-            Embedding,
-            GarNet, GarNetStack
-        ]
+        rf_layers = accum_layers + [BatchNormalization, Activation, Embedding, GarNet, GarNetStack]
 
         for layer in rf_layers:
             attrs = self.attribute_map.get(layer, [])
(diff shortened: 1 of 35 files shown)


jmitrevs (changes requested):

hls4ml/utils/config.py
```diff
@@ -135,7 +127,7 @@ def config_from_keras_model(model, granularity='model', default_precision='ap_fi
 
     print('Topology:')
     for keras_layer in keras_layer_config:
-        if keras_layer['class_name'] not in supported_layers:
+        if keras_layer['class_name'] not in hls4ml.model.layers.layer_map:
```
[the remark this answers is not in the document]
vloncar: Yes, that's what I noticed as well. Anyway, I'm reworking this from the ground up, hopefully finished tomorrow, I'll let you know when it's really ready for review

vloncar:
@jmitrevs @jmduarte This is now redesigned and ready for review

jmduarte:
For the test that fail, i think it's related to this error from the keras model conversion step, which happens when we try to `import hls4ml`:
```
WARN: Unable to import optimizer(s) from qkeras.py: No module named 'pyparsing'
WARN: Unable to import optimizer(s) from quantization_templates.py: No module named 'pyparsing'
WARN: Unable to import optimizer(s) from quantization_templates.py: No module named 'pyparsing'
WARN: Unable to import optimizer(s) from quantization_templates.py: No module named 'pyparsing'
```
I think there's some weird conflict of libraries happening. I would do one of two things: update the docker image or just (re-?)install `pyparsing` in the Jenkinsfile

jmitrevs:
I confirmed that this works in our use of extension API (after removing the unused config). Let me look once more, but I think this is a good change.

hls4ml/backends/fpga/fpga_backend.py
```diff
@@ -2,14 +2,13 @@
...
 from bisect import bisect_left
-from queue import Queue
 from collections.abc import Iterable
 import re
 
 from hls4ml.backends.backend import Backend
-from hls4ml.model.layers import Layer
-from hls4ml.model.attributes import Attribute
-from hls4ml.model.types import IntegerPrecisionType, FixedPrecisionType, XnorPrecisionType, ExponentPrecisionType
+from hls4ml.model.layers import Layer, Dense, BatchNormalization, Activation, Softmax, Conv1D, Conv2D, SeparableConv1D, SeparableConv2D, Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D, SimpleRNN, LSTM, GRU, Dot, Embedding, GarNet, GarNetStack
```
jmitrevs, line 9: I would prefer doing something like
```python
from hls4ml.model.layers import (
    Layer, Dense, BatchNormalization, Activation, Softmax, Conv1D, Conv2D, SeparableConv1D,
    SeparableConv2D, Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D, SimpleRNN,
    LSTM, GRU, Dot, Embedding, GarNet, GarNetStack)
```
for the long lines. This is what I see in the matplotlib source code, for example:
https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/backends/backend_qt.py#L9-L12

(As an aside, VS code tells me that `Layer` is not used explicitly.)

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -8,7 +8,7 @@
 
 from hls4ml.model.types import FixedPrecisionType, NamedType, IntegerPrecisionType
 from hls4ml.model.layers import Layer, Dense, BatchNormalization, Embedding, Conv1D, Conv2D, Conv2DBatchnorm, SeparableConv1D, SeparableConv2D, DepthwiseConv2D, Activation, ParametrizedActivation, PReLU, Softmax, Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D, ZeroPadding1D, ZeroPadding2D, Merge, Concatenate, Dot, Resize, Transpose, SimpleRNN, LSTM, GRU, GarNet, GarNetStack
```
jmitrevs, line 10: Same comment here about avoiding super-long lines with parentheses, like https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/backends/backend_qt.py#L9-L12.

hls4ml/backends/fpga/fpga_backend.py
jmduarte: Hi @jmitrevs, I wouldn't worry about formatting things like this because we're going to run the `pre-commit` hooks on this PR after the review, which will resolve all of this

hls4ml/backends/vivado/vivado_backend.py
jmduarte: `pre-commit` will handle this

hls4ml/model/attributes.py
```diff
@@ -15,6 +16,14 @@ def validate_value(self, value):
         else:
             return True # Meaning we don't care
 
+    @property
+    def config_name(self):
+        return convert_to_pascal_case(self.name)
+
+class ConfigurableAttribute(Attribute):
```
jmitrevs, line 23: A doc-string would be useful to describe configurable attributes (and maybe the case conventions?)

jmitrevs (approved):

jmitrevs:
Let see the tests finish, but if people don't object, I think we can merge this PR. In general, though, I would prefer more documentation on the attributes, what are the common attributes for different categories, etc.

hls4ml/model/graph.py
```diff
@@ -712,7 +712,7 @@ class TraceData(ctypes.Structure):
         layer_sizes = {}
         n_traced = 0
         for layer in self.get_layers():
-            if layer.get_attr('function_cpp', None) and layer.get_attr('Trace', False):
+            if layer.get_attr('function_cpp', None) and layer.get_attr('trace', False):
```
jmduarte, line 715: Does this mean the key in the config is now called `'trace'` instead of `'Trace'`? Or is this just related to the `snakeCase -> pascal_case` internal translation? Basically, does the way a user manipulates the configuration keys stay the same?

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -23,12 +23,39 @@ def __init__(self):
...
+
+        for layer in rnn_layers:
+            attrs = self.attribute_map.get(layer, [])
+            attrs.append(ConfigurableAttribute('recurrent_reuse_factor', default=1))
+            attrs.append(ConfigurableAttribute('static', value_type=bool, default=True))
+            self.attribute_map[layer] = attrs
+        
+        # Add ParallelizationFactor to Conv1D/2D
+        pf_layers = [
+            Conv1D, Conv2D,
```
jmduarte, line 39: Should `PointwiseConv1D` and `PointwiseConv2D` from `hls4ml.backends.fpga.fpga_layers` be in this list (and ditto in other places)? Or how should they be handled? 

I know currently in the `main` branch they're not really doing anything, but this is a layer that I'm using in the Jet ID GNN, which in our custom branch *can* have a different behavior

hls4ml/utils/config.py
```diff
@@ -281,8 +188,9 @@ def make_layer_config(layer):
     model_config['Precision'] = default_precision
     model_config['ReuseFactor'] = default_reuse_factor
     model_config['Strategy'] = 'Latency'
+    model_config['BramFactor'] = 1_000_000_000
     #model_config['Compression'] = False
-    #model_config['Trace'] = False
+    model_config['TraceOutput'] = False
```
jmduarte, line 193: Does this key mean all layers are traced if this is set to `True`?

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -23,12 +23,39 @@ def __init__(self):
...
+        # Add ConvImplementation to Convolution+Pooling layers
+        cnn_layers = [
+            Conv1D, Conv2D,
+            SeparableConv1D, SeparableConv2D, DepthwiseConv2D,
+            Pooling1D, Pooling2D
+        ]
+
+        for layer in cnn_layers:
+            attrs = self.attribute_map.get(layer, [])
+            attrs.append(ConfigurableAttribute('conv_implementation', value_type=str, default='LineBuffer'))
```
jmduarte, line 56: Should `conv_implementation` be `ConfigurableAttribute `or `ChoiceAttribute`?

jmduarte (commented):
Overall looks great, and my initial testing works. I have some minor questions.

hls4ml/model/graph.py
vloncar: Yes, 'Trace' becomes `trace` because otherwise there would be both `Trace` and `trace` in the attribute list. This is just an interim solution, in the future I hope we can move to pascal_case everywhere.

hls4ml/backends/vivado/vivado_backend.py
vloncar: Lest there are bugs, those two layers will get a RF and PF because they inherit from Conv1D/2D. In general, the classes inherit the attributes of the superclass and only need to define new attributes.

hls4ml/utils/config.py
vloncar: No, this key simply turns on tracing. Then all layers that have `trace` set to true will be traced.

hls4ml/backends/vivado/vivado_backend.py
vloncar: You're right, fixed.

hls4ml/backends/vivado/vivado_backend.py
jmduarte: ok, i confirm this is the case

jmduarte (approved):

jmitrevs:
Assuming the CI looks good, I'll merge this later today, unless someone objects.

vloncar:
Wait, we need to do the pre-commit stuff. I'm working on it, but there are A LOT of issues.

jmitrevs:
My understanding from @jmduarte was that it would automatically format things, so you don't need to manually do it, though I am not 100% sure.

jmduarte:
some are fixed automatically (black formatting, isort sorting imports, etc.), some are not (flake8 issues). @vloncar is fixing (most) of the remaining issues now, and i will take a look at the rest, then we can merge

vloncar:
I fixed all of the issues from flake8, but I cheated a bit by ignoring the print statement warnings. `pre-commit` hook is changed to ignore this, but we should make a note of this and replace all print statements with proper logging (is on my TODO list, but way back). For some reason this still fails the black and isort stages. Javier understands this better.

PR #694, opened 2022-12-14
Fix function call in Alveo tcl script

vloncar:
# Description

After changes in #626 the file list passed to the `add_files` and `import_files` tcl functions is using a variable, but the `{...}` syntax doesn't do expansion so the name of the file is never properly made, causing the script to fail. This can be solved in several ways buy since the functions expect list of strings, the cleanest approach is to use `[list ...]`.

## Type of change

## Tests

Not really needed, this is a syntax fix.

## Checklist


jmitrevs (approved):

jmitrevs (changes requested):
Can you rebase and make sure the tests finish successfully?

jmitrevs (approved):

PR #700, opened 2022-12-22
Quartus bram factor

jmitrevs:
# Description

This enables the BramFactor functionality for the quartus backend. For now, it's very much a copy of how it is done for Vivado. We will test it's usefulness in the near future.

## Type of change

## Tests

The pytest, `test_bram_factor.py`, was broadened to cover this use case.

## Checklist


jmitrevs:
I am not sure this is the best approach in quartus, however. I believe all pointer variables are put into one memory. It may be better to have explicit Avalon MM interfaces for the weights. The question then becomes whether it's better for them to be Host or Agent. In Host, the memory is external, and has to be optimized externally. With Agent, it's internal, optimized by the HLS compiler. I am not sure what's better.

jmitrevs:
I think viewing this as code to put in a wrapper is useful. I think it's true that the performance wouldn't be good if this was used directly. I think making this perform better as is would make it more difficult to incorporate. One could argue in general if the `component` attribute is better left off the source code and put in the compilation flags only.

hls4ml/templates/quartus/firmware/myproject.cpp
```diff
@@ -42,25 +42,25 @@
...
+// hls-fpga-machine-learning initialize input/output
+
+// ****************************************
+// NETWORK INSTANTIATION
+// ****************************************
 
-   // ****************************************
-   // NETWORK INSTANTIATION
-   // ****************************************
+// hls-fpga-machine-learning insert layers
```
vloncar, line 64: Hmm, is there a way to exclude templates like this file from pre-commit? These comments are strategically placed to ensure the generated code remains readable. Writers expect this. If the spaces and indentations are removed, it will result in uglier code being emitted.

vloncar (approved):

PR #708, opened 2023-02-02
Fix for cloned stream that is subsequently flattened

### LINKED ISSUE #707: Cloned HLS stream is read from twice. Second duplicate stream is left full.
## Quick summary
During the writing of an HLS project for a model with a skip connection, only one of the two copies of the stream are used in the first "branch". The original stream is propigated to the second "branch" instead of the second copy. This results in reading from a stream that is already empty.
Commit hash: 107589fb1dac9aa898baea328acc6d519596e19d
 [TF Model architecture](https://github.com/abekek/Rapid-Fitting-of-BEPFM-and-Hysteresis-Loops-Using-Physics-Constrained-Unsupervised-Neural-Networks/blob/main/Assets/Figures/model_architecture_sho.png)

In addition tot this, the concatenate1d() function of nnet_merge_stream.h has a error in both loops where the out_data assignments don't have an additional offset to account for the progress of the outer loop. The fixed version is given below.

## Details
The appropriate output stream is cloned with nnet::clone_stream<>() (layer13_out stream becomes layer71_cpy1 & layer71_cpy2). layer71_cpy1 is then correctly passed to the first "branch" but layer13_out (which has already been read from for the cloning process) is incorrectly passed to the second branch instead of layer71_cpy2. These warnings are produced as a result along with incorrect predictions.
```
WARNING: Hls::stream 'layer13_out' is read while empty, which may result in RTL simulation hanging.
WARNING: Hls::stream 'layer71_cpy2' contains leftover data, which may result in RTL simulation hanging.
```

### Steps to Reproduce
Download & unzip files.zip
1. `conda env create -f environment.yml`
2. `conda activate bepfm_fpga_temp`
3. Install main branch of hls4ml
   a. `git clone https://github.com/fastmachinelearning/hls4ml.git`
   b. `cd hls4ml;pip install .;cd ..`
4. `python3 compile_prj.py`
5. Navigate to sho_fitter_stream_rf8_TEMP/firmware/myproject.cpp

### Expected behavior
layer13_out is cloned on line 173 of myproject.cpp to layer71_cpy1 and layer71_cpy2.
layer71_cpy1 should be passed into the pooling1d_cl() on line 177 which works currently.
layer71_cpy2 should be passed into the dense() on line 345.

### Actual behavior
layer13_out is cloned on line 173 to layer71_cpy1 and layer71_cpy2.
layer71_cpy1 should be passed into the pooling1d_cl() on line 177 which works currently.
layer71_cpy2 should be passed into the dense() on line 345 but instead layer13_out is passed in which is incorrect.

[files.zip](https://github.com/fastmachinelearning/hls4ml/files/10560665/files.zip)

### Possible fix
A possible fix for the concatenate1d function bug is below.
[Line 392](https://github.com/fastmachinelearning/hls4ml/blob/107589fb1dac9aa898baea328acc6d519596e19d/hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h#L392) can be changed to: `out_data[j + (i*input1_T::size)] = in_data1[j];`
[Line 400](https://github.com/fastmachinelearning/hls4ml/blob/107589fb1dac9aa898baea328acc6d519596e19d/hls4ml/templates/vivado/nnet_utils/nnet_merge_stream.h#L400) can be changed to: `out_data[j + (i*input2_T::size) + (CONFIG_T::n_elem1_0)] = in_data2[j];`

jmduarte:
# Description

Fixes #707. The issue is if a cloned stream is subsequently flattened, the original stream is used later instead of the clone. I think this is because the `InplaceVariable` is not correctly updated.

Somewhat related, there is also a bug in `concatenate1d` when you consider the case that a stream is not truly from a 1D array, but from a flattened 2D array. I tried to use @rfforelli's fix for this, but it seems to not be quite working. Maybe he has some input?

## Type of change

## Tests

- Added a test that should fail on `main` but succeed on this branch

**Test Configuration**:

## Checklist

- [x] I have installed and run `pre-commit` on the files I edited or added. (*to be done later*)


jmitrevs:
Flatten is one of the operations that often breaks in the current main branch. The qonnx branch has quite a few fixes there, because the inplace variable as is in main is generally broken when working with optimizers. It sets the cloning too early. Let me take a look more carefully at this fix.

jmitrevs:
I actually have the test fail in main, ingest-qonnx-master, and clone_flatten branches, but I think the form of the code in myproject.cpp produced by both the ingest-qonnx-master and clone_flatten branches seems correct. I would recommend using the ingest-qonnx-master branch for these cases, because as I mentioned, inplace variables are broken in main.

hls4ml/backends/fpga/passes/clone.py
```diff
@@ -64,13 +65,20 @@ def transform(self, model, node):
...
+                }
+                clone_layer = model.make_node(Clone, 'clone_' + node.name, attrs, [output], [output + '_cpy' + str(i + 1) for i in range(len(output_map[output]))])
                 for i, layer in enumerate(output_map[output], 1):
-                    attrs = {
-                        'size' : np.prod(out_var.shape)
-                    }
                     idx = layer.inputs.index(output)
                     layer.inputs[idx] = output + '_cpy' + str(i)
-                clone_layer = model.make_node(Clone, 'clone_' + node.name, attrs, [output], [output + '_cpy' + str(i + 1) for i in range(len(output_map[output]))])
+                    if isinstance(layer, Reshape):
```
jmitrevs, line 75: This is band-aid for the fundamental problem of InplaceVariables, and basically every optimizer would need to include it. The issue is not just here. This is what I attempted to solve in #525, but it subsequently got more work done in #591. I think that's the real solution.

jmduarte: Yes, I see your point.

jmduarte:
OK, the new test should now succeed in `clone_flatten` but fail in `main`.

The test fails in `main` for two reasons: (1) the `InplaceVariable` problem and (2) the incorrect `concatenate1d` HLS code for the case when the input stream depth is not the same as full size of the array (e.g. if they are from flattened 2D streams).

This PR has a band-aid solution for (1) and a good solution for (2). 

Perhaps, the best way to proceed is 
- Check that if we add the solution to (2) to the `ingest-qonnx-master`, the test passes.
- If so, then we update this PR to just solve (2) since (1) will be solved by the more general `ingest-qonnx-master` fixes.
- We can also add the new test to the `ingest-qonnx-master` branch (after it's rebased to include changes from this branch).

jmitrevs:
If I copy `nnet_merge_stream.h` from `clone_flatten` to `ingest-qonnx-master`, the test passes there. So (2) is definitely a bug to fix. I think I'll try to extract the inplace variable stuff from from `ingest-qonnx-master` and make it a separate PR, in order to make the merging of that PR easier.

One thing, can you check if the same fix is needed for quartus? On a first glance it looks to me like it's needed there, too. The code is at https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/templates/quartus/firmware/nnet_utils/nnet_merge_stream.h#L139-L165.

hls4ml/backends/fpga/passes/clone.py
jmduarte: OK removed from here now

jmduarte:
@jmitrevs OK, I confirm that the added test (including Quartus) passes if I merge this branch with #714. So should we plan to merge them both at the same time?

rfforelli (approved):

rfforelli (approved):

jmitrevs (approved):

jmitrevs (changes requested):
Please fix pre-commit, and I will approve.

vloncar:
What's the deal with the failed test?

jmitrevs:
I assume the failed tests are ones that succeed when #714 is also merged.

jmitrevs (approved):

PR #714, opened 2023-02-14
Fix inplace variables

jmitrevs:
# Description

Inplace variables are broken in the main branch because they are fixed too early. Subsequent optimizations can make this connection no longer valid. Issue #707 (and the attempted fix in #708) is an example of this issue. This was largely developed in the `ingest-qonnx-master` branch, but is extracted here, and an omission related to stream repacking for quartus was discovered and fixed.

## Type of change

Bug fix

## Tests

The pytest, `test_reshape.py`, is the main test, though `test_cnn_mnist.py` is also pertinent. The new reshape functionality for Intel needs testing, though. There is also the question of whether flatten can always be removed for io_stream. This should be checked further, and if exceptions are found, we may need to change when we run the optimizers.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works. (May want to add another for Flatten. in various cases.)

### DIFF
--- a/hls4ml/model/layers.py
+++ b/hls4ml/model/layers.py
@@ -383,7 +383,7 @@ class Reshape(Layer):
             target_shape = target_shape[1:]
 
         # take care of -1 shapes
-        shape = self.infer_shape(input_shape, target_shape)
+        shape = self._infer_output_shape(input_shape, target_shape)
 
         # update the target shape with chnges from above
         self.set_attr('target_shape', shape)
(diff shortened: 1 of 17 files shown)


hls4ml/model/layers.py
```diff
@@ -359,17 +362,44 @@ def initialize(self):
...
+            # for QONNX, remove batch dimension
+            if shape_node:
+                target_shape = shape_node.value[1:]
+            else:
+                raise RuntimeError("Reshape for ONNX requires the target shape to be a second input.")
 
-        out_name = self.outputs[0]
-        proxy = self.get_input_variable()
-        out = InplaceVariable(shape, dims, proxy)
+        # remove Nones -- is this ever triggered?
```
vloncar, line 381: The way I see this failing is if you specify the batch size during the creation of the model (like passing `batch_size` to the `Input` layer that some RNNs do). Then the first element won't be None in the json of the model that we parse. Not sure if you would even get this far, or will the conversion fail somewhere earlier.

jmitrevs: In ONNX the batch dimension is not None, and from what I gather can be any number, not necessarily just 1. We need to have a robust way to tell what is the batch dimension that needs to be removed. (Alternately, it can be kept and ignored in all cases.)

vloncar: The number of elements must match, if it doesn't and the first dimension is not none, that's a batch. Check that is a multiple just to be sure. The number of elements can match, the first one can be 1 and be a batch, but I think in that case we should still be fine.

vloncar (approved):

PR #715, opened 2023-02-15
Correctly expand dims of pointwise layer

vloncar:
# Description

When creating PointwiseConv layers we always expand dimensions of weights to `(1, 1, x, y)`, but in 1d we should only expand to `(1, x, y)`. This breaks some model Thea is using.

## Type of change

## Tests

From Thea:
```
import numpy as np
import tensorflow as tf
import hls4ml

np.random.seed(12)
tf.random.set_seed(12)

from tensorflow import keras
import tensorflow.keras.layers as KL

def get_deepsets_net(input_shape=(16, 16)):
    deepsets_input = keras.Input(shape=input_shape, name="input_layer")

    # Permutation Equivariant Layer
    x_max = KL.GlobalMaxPooling1D(keepdims=True)(deepsets_input)
    x_lambd = KL.Dense(32, use_bias=False)(x_max)
    x_gamma = KL.Dense(32)(deepsets_input)
    x = KL.Subtract()([x_gamma, x_lambd])
    x = KL.Activation('elu')(x)

    x_max = KL.GlobalMaxPooling1D()(x)
    x = KL.Dense(16)(x)
    x = KL.Activation('elu')(x)
    deepsets_output = KL.Dense(5)(x)
    deepsets_network = keras.Model(deepsets_input, deepsets_output, name="deepsets")
    return deepsets_network

input = np.random.rand(1,16,16)
input_shape = input.shape[1:]
model = get_deepsets_net(input_shape=input_shape)
model.summary()

config = hls4ml.utils.config_from_keras_model(model, granularity='model')
hls_model = hls4ml.converters.convert_from_keras_model(model, hls_config=config)
```
Note that there is a misparse due to a `keepdims` argument not being supported in GlobalPooling layer. I'll push a PR for that as well.

## Checklist

I can't run `pre-commit` due to a strange bug so this will probably fail the test, but I'll fix that afterwards.

### DIFF
--- a/hls4ml/backends/quartus/passes/pointwise.py
+++ b/hls4ml/backends/quartus/passes/pointwise.py
@@ -65,7 +65,8 @@ class OptimizePointwiseConv(OptimizerPass):
         dim = node.__class__.__name__[-2:] # '1D' or '2D'
         pw_node = model.make_node('PointwiseConv' + dim, node.name, copy(node.attributes), node.inputs.copy(), outputs=node.outputs.copy())
         if len(node.weights['weight'].data.shape) == 2: # This can happen if we assign weights of Dense layer to 1x1 Conv2D
-            pw_node.weights['weight'].data = np.expand_dims(node.weights['weight'].data, axis=(0,1))
+            expand_axis = tuple(range(int(dim[0])))
+            pw_node.weights['weight'].data = np.expand_dims(node.weights['weight'].data, axis=expand_axis)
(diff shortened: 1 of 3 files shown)


thaarres:
@jmduarte @vloncar Feel free to use the model above as test if it's useful, but it might have other issues down the line :)

vloncar:
I would add the deep sets-like model in the `example-models` repository, this is something we don't have and it is a useful demonstration. Do we have a reasonable model at hand?

PR #716, opened 2023-02-15
Support keepdims in GlobalPooling layers

vloncar:
# Description

Support `keepdims` argument of pooling layers. Helps correctly parse some models.

## Type of change

## Tests

Look in #715 for a test case.


jmduarte:
Tests fail because our 0.2 testing image had TensorFlow 2.4.1 before `keepdims` was added to the config of these pooling layers (I think).

I created a 0.4 testing image with TensorFlow 2.9.1, so I relaunched tests with that.

jmduarte (approved):
Looks good to me. I updated the test to include `keepdims=True/False` by parametrizing the fixture. That was the main change I was thinking of with the deep sets model, although I agree adding such a model to `example-models` would be a good follow-up PR.

Issue #409, opened 2021-10-04
PyTorch conversion issues


thesps:
## Improper handling of PyTorch `Linear` layer with `bias=False`

When a PyTorch model has a `Linear` layer with `bias=False`, conversion crashes out when trying to access the non-existent bias data from the model.
Minimal example to reproduce the issue:

```
import torch
import hls4ml
model = torch.nn.Sequential(torch.nn.Linear(1,1,bias=False))
config = hls4ml.utils.config_from_pytorch_model(model)
hls_model = hls4ml.converters.convert_from_pytorch_model(model, (1,1) hls_config=config)
hls_model = hls4ml.converters.convert_from_pytorch_model(model, hls_config=config)
```

End of error message:
```
~/Work/hls4ml/hls4ml-master/hls4ml/model/hls_layers.py in add_bias(self, quantizer)
    460
    461     def add_bias(self, quantizer=None):
--> 462         data = self.model.get_weights_data(self.name, 'bias')
    463         precision = None
    464         type_name = None

~/Work/hls4ml/hls4ml-master/hls4ml/model/hls_model.py in get_weights_data(self, layer_name, var_name)
    477
    478     def get_weights_data(self, layer_name, var_name):
--> 479         return self.reader.get_weights_data(layer_name, var_name)
    480
    481     def next_layer(self):

~/Work/hls4ml/hls4ml-master/hls4ml/converters/pytorch_to_hls.py in get_weights_data(self, layer_name, var_name)
     58             var_name = torch_paramap[var_name]
     59
---> 60         data = self.state_dict[layer_name + '.' + var_name].numpy().transpose() #Look at transpose when systhesis produce lousy results. Might need to remove it.
     61
     62         return data

KeyError: '0.bias'
```

## Non-handling of `ParametrizedActivation` parameter

The parameter(s) of `ParametrizedActivation` layers, e.g. `ELU`, `PReLU`, `LeakyReLU` don't get propagated through the conversion. Simple demonstration case:
```
import torch
import hls4ml
model = torch.nn.Sequential(torch.nn.ELU(alpha=0.5))
hls4ml.utils.config_from_pytorch_model(model)
hls_model = hls4ml.converters.convert_from_pytorch_model(model, input_shape=[1,1])
print(list(hls_model.get_layers())[1].attributes)
```
Gives:
```
[...]

PR #723, opened 2023-02-24
[WIP] Improved parsing of pytorch models using torch.FX

JanFSchulte:
Current parsing of pytorch models uses a loop of the `named_modules` of the model (https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/converters/pytorch_to_hls.py#L163). This has several disadvantages: 

- Captures only layers that are defined as members of the model class
- Can't infer the correct order of models
- Ignores other operations that are part of the `forward()` method of the model

In this PR, we propose to fix this by first created a graph representation of the model's `forward()` function using the symbolic tracing functionality of [https://pytorch.org/docs/stable/fx.html](torch.FX). Each operation in the `forward()` is represented by a node in the graph. Nodes can be of these types:
image

For example, for this model

```
class MyModuleConvRelu(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = torch.nn.Conv2d(3,3,3)
        
    def forward(self, x):
        y1 = self.conv(x)
        y = torch.relu(y1)
        y = y + y1
        y = torch.relu(y)
        return y
```
the resulting graph representation is 

```
graph():
    %x : [#users=1] = placeholder[target=x]
    %conv : [#users=2] = call_module[target=conv](args = (%x,), kwargs = {})
    %relu : [#users=1] = call_function[target=torch.relu](args = (%conv,), kwargs = {})
    %add : [#users=1] = call_function[target=operator.add](args = (%relu, %conv), kwargs = {})
    %relu_1 : [#users=1] = call_function[target=torch.relu](args = (%add,), kwargs = {})
    return relu_1
```

As the nodes in the graph follow the order of operations of the `forward()` function, we can then simply loop over them and parse each node into one node in the hls4ml model representation. For the parsing of the individual layers, existing code is used where available without significant changes. Functionality for more types of layers is also added by this PR. 

The types of layers currently understood by the parser are

- Linear
- Softmax
- Relu
- LeakyReLU
- ThresholdedReLU
- ELU
- PReLU
- Sigmoid
- BatchNorm2d
- BatchNorm1d'
- Batch_norm
- MaxPool1d
- MaxPool2d
- AvgPool1d
- AvgPool2d
- Add
- Subtract
- Multiply
- Average
- Maximum
- Minimum
- Concatenate
- Dot
- Conv1d
- Conv2d
- View
- Dropout
- Flatten
- Sequential

This PR also fixes https://github.com/fastmachinelearning/hls4ml/issues/409 

Changes are mostly confined to the frontend, but small changes are made to the backend in 

hls4ml/backends/fpga/passes/codegen.py
hls4ml/backends/vivado/passes/conv_stream.py
hls4ml/backends/vivado/passes/repack_stream.py

to handle the different order of inputs in pytorch compared to keras.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Tests

The new parsing was tested using 5-6 different pytorch model examples from around the web. In addition, I verified that the two example models for pytroch included with hls4ml get parsed successfully. 
 
## Checklist


JanFSchulte:
The baseline version of this new pytorch parser has reached a stable state now and the code is ready for review. This implementation of baseline support will then be used to built functionality for graph models down the road.

JanFSchulte:
Closed because git history of this branch is irrecoverable messed up.

PR #730, opened 2023-03-12
print_vivado_report function for nicer reports

vloncar:
# Description

As a cherry on top for a new release, I propose we make the printing of reports nicer for the user. Yes, this is a result of boredom because I didn't go skiing this weekend, but also because I don't like the ugly screenshots of text reports that many people place on their slides. I guess we can expand this in the future to be a bit more interactive and include the per-layer reports.

Note that the hidden among the style changes is a fix for parsing cosim reports. This previously failed on the accelerator backend.

To try it out, run something like
```python
report = hls4ml.report.parse_vivado_report('/some/path/') # or report = hls_model.build(...)
hls4ml.report.print_vivado_report(report)
```

Hint: Run it in jupyter and as a script to see the difference, I won't spoil the fun here :wink: 

## Type of change

## Tests

The notebook part is not really testable with the current setup, and the text-based reports can only be tested for sanity (if the function fails or not), so I didn't add any tests.

## Checklist


hls4ml/report/vivado_report.py
```diff
@@ -309,3 +359,304 @@ def parse_vivado_report(hls_dir):
...
+        params['interval_min'] = csynth_report['IntervalMin']
+        params['interval_max'] = csynth_report['IntervalMax']
+        params['estimated_clock'] = csynth_report['EstimatedClockPeriod']
+
+        params['bram'] = str(bram) + ' / ' + str(avail_bram) + ' (' + str(round(bram / avail_bram * 100, 1)) + '%)'
+        params['dsp'] = str(dsp) + ' / ' + str(avail_dsp) + ' (' + str(round(dsp / avail_dsp * 100, 1)) + '%)'
+        params['ff'] = str(ff) + ' / ' + str(avail_ff) + ' (' + str(round(ff / avail_ff * 100, 1)) + '%)'
+        params['lut'] = str(lut) + ' / ' + str(avail_lut) + ' (' + str(round(lut / avail_lut * 100, 1)) + '%)'
+        if 'URAM' in csynth_report:
+            params['uram'] = str(uram) + ' / ' + str(avail_uram) + ' (' + str(round(uram / avail_uram * 100, 1)) + '%)'
```
jmduarte, line 555: from my tests, this can still cause a divide by zero error (because both `uram` and `avail_uram` are 0). We could just explicitly check if `avail_uram > 0` before dividing.

jmduarte (commented):
Looks great! I have some minor feedback. 

1. It looks like `avail_uram` can be 0 and we try to divide by it
2. I think it would be nice to add a test. Here's one approach: we can run the parser/printer on a pre-generated report. Then we can check that the stdout matches what we expect (because pytest captures all stdout/stderr). 

I submitted a PR to your branch to address these: https://github.com/vloncar/hls4ml/pull/51
Let me know what you think about this testing approach.

vloncar:
I made this PR as a little bonus that would make the notebook-based workflows (and tutorials) more appealing, I didn't think of it being a PR that introduces testing for reporting feature. You added a basic test that IMO suffices, but if you want to do more, I'd be happy to review and merge it.

jmduarte (approved):

Issue #740, opened 2023-03-28
CNNs with binary quantizer fail


sandeep1404:
Hi, 
I am using hls4ml version 0.6.0 when I used to compile the hls_model using hls_model.compile() it throws me the below error
```
Writing HLS project
Done
firmware/myproject.cpp: In function ‘void myproject(input_t*, result_t*, short unsigned int&, short unsigned int&)’:
firmware/myproject.cpp:101:17: error: ‘layer15_t’ was not declared in this scope; did you mean ‘layer17_t’?
  101 |     nnet::dense<layer15_t, layer17_t, config17>(layer15_out, layer17_out, w17, b17); // qdense_1
      |                 ^~~~~~~~~
      |                 layer17_t
firmware/myproject.cpp:101:49: error: ‘layer15_out’ was not declared in this scope; did you mean ‘layer17_out’?
  101 |     nnet::dense<layer15_t, layer17_t, config17>(layer15_out, layer17_out, w17, b17); // qdense_1
      |                                                 ^~~~~~~~~~~
      |                                                 layer17_out
g++: error: myproject.o: No such file or directory
---------------------------------------------------------------------------


`OSError                                   Traceback (most recent call last)
Cell In[31], line 1
----> 1 hls_model1.compile()

File ~/.local/lib/python3.8/site-packages/hls4ml/model/hls_model.py:546, in HLSModel.compile(self)
    544         dlclose_func.restype = ctypes.c_int
    545         dlclose_func(self._top_function_lib._handle)
--> 546     self._top_function_lib = ctypes.cdll.LoadLibrary(lib_name)
    547 finally:
    548     os.chdir(curr_dir)

File /usr/lib/python3.8/ctypes/__init__.py:451, in LibraryLoader.LoadLibrary(self, name)
    450 def LoadLibrary(self, name):
--> 451     return self._dlltype(name)

File /usr/lib/python3.8/ctypes/__init__.py:373, in CDLL.__init__(self, name, mode, handle, use_errno, use_last_error, winmode)
    370 self._FuncPtr = _FuncPtr
    372 if handle is None:
--> 373     self._handle = _dlopen(self._name, mode)
    374 else:
    375     self._handle = handle

OSError: firmware/myproject-0FfB97A6.so: cannot open shared object file: No such file or directory`
```
I have no idea how to resolve this can anyone help me in resolving this issue` 
#333 #276 #264  I have checked the earlier issues who faced this problem but not able to resolve the issue from the discussion

[...]

PR #749, opened 2023-04-06
CNNs with binary inputs and weights need fixes

jmitrevs:
# Description

As shown by #740, CNNs with binary quantizers don't currently work properly. This PR attempts to fix it.

## Type of change

## Tests

A pytest is added. However, it still shows errors in streaming.

## Checklist


jmitrevs:
Before I pushed 7186346 the test results were:
```
test_binary_cnn.py::test_model2[io_parallel-Vivado] PASSED                                                    [ 16%]
test_binary_cnn.py::test_model2[io_stream-Quartus] FAILED                                                     [ 33%]
test_binary_cnn.py::test_model2[io_stream-Vivado] FAILED                                                      [ 50%]
test_binary_cnn.py::test_model2[io_stream-Vitis] FAILED                                                       [ 66%]
test_binary_cnn.py::test_model2[io_parallel-Quartus] FAILED                                                   [ 83%]
test_binary_cnn.py::test_model2[io_parallel-Vitis] PASSED                                                     [100%]
```
After, they all fail. So it was not a successful fix.

jmitrevs:
This is not complete, and the final test has been disabled, but I think the fixes that we have here should be added before we make 0.7.0 RC. The other stuff we can add after the RC is built or later.

The remaining to-dos:

1. Max Pooling does not propagate XnorPrecisionType from the input to the output, so the logic is confused. The output zeros are no longer considered to be -1 but just 0.
2. Conv2d, io_parallel, Resource fails badly with binary inputs--the problem has not been debugged carefully.

If this is accepted, those to-dos should be moved to an issue.

jmitrevs:
This fixes a few things:

1. `cast` is called with the correct mult_config in CNN cases in all cases. Previously it was correct in many cases but not all.
2. `cast` was fixed in the quartus case since you cannot cast `ac_fixed` to `ac_int` without calling `.to_ac_int()`. Maybe one needs to add more templates here depending on whether the input types are ac_fixed or ac_int to cover all the options. (I understand `ac_int` to `ac_fixed` casting is allowed.) It's worth looking in more detail since we probably do not cover all cases.
3. The binary versions of batchnorm did not make use of `n_scale_bias`, which is not the same as `n_in` when `n_filt != -1`. They have been updated to match the regular batchnorm. (This was also coped to quartus, where it was missing everywhere.)

vloncar (approved):

PR #754, opened 2023-04-09
Fix for BatchNormalization layers with `center=False` or `scale=False`

jmduarte:
- Fix for batch normalization layers with `center=False` or `scale=False` (i.e. `beta` or `gamma` are `None`)


jmitrevs:
What milestone is this for?

jicampos:
@jmduarte I agree with @vloncar for the PyTorch parser. Would you like to add the fix for Keras now and follow up after #723?

JanFSchulte:
I have updated #723 to support parsing of the `affine` argument so it will work with these changes.

PR #772, opened 2023-04-19
add a default weight_size

jmitrevs:
# Description

It's possible to read the variable before writing. This sets a default value

## Type of change

## Tests

Standard tests should be enough

## Checklist


jmitrevs:
I believe an error was triggered with batch norms in Quartus, with weight_size being undefined. This fix treats BatchNorm weights as being small, which might not be the best fix in all cases. Nevertheless, we should not read a variable without making sure that each possible branch has it defined.

vloncar:
0.7.0?

jmitrevs:
I think so. It's a 1-line change, which prevents a crash.

vloncar (approved):

Issue #759, opened 2023-04-13
Conv layer changes other layers to Resource unless specifically set


jmitrevs:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x] If there are steps to reproduce the problem, make sure to write them down below.
- [x] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary

If you have a conv layer in your network, setting the Strategy at the top level is overwritten after the first Conv layer. If you want to have all the nodes be implemented with the Latency strategy, 

```python
hls_config["Model"]["Strategy"] = "Latency"
```
is not enough. You have to have:

```python
for layer in hls_config['LayerName'].keys():
    hls_config['LayerName'][layer]['Strategy'] = "Latency"
```

## Details
I believe this is because Conv requires the dataflow style at the top level. The Conv layer changes the setup to dataflow but inadvertently changes all the layers not explicitly configured to "Resource".  The cleanest solution may be to split the dataflow style from layer Resource/Latency strategies.

### Steps to Reproduce

A simple setup is to look at `test_binary_cnn.py` (from #749) and change the strategy to Latency. You can see that only the first CNN is Latency, the others are Resource, in the parameters.h.

### Possible fix
The cleanest solution may be to split the dataflow style from layer Resource/Latency strategies.

vandenBergArthur:
Hi, 
I have a quick question related to this.
When I create my HLS model with `hls4ml.converters.keras_to_hls`, I get the following warning: `WARNING: Cannot use "Latency" model strategy for conv2d layer. Switching to "Resource" strategy.`

Is this a bug? If not: why can't I use `Latency` strategy?

jmitrevs:
You can get rid of that warning by having
```python
hls_config["Model"]["Strategy"] = "Resource"
for layer in hls_config['LayerName'].keys():
    hls_config['LayerName'][layer]['Strategy'] = "Latency"
```
Then you can use Latency for the layers but there needs to be a dataflow construct at the top level. Dataflow has traditionally been associated with Resource. We will make this less confusing in the future. It is a bug that you cannot just say `hls_config["Model"]["Strategy"] = "Latency"`

vandenBergArthur:
Thanks for the reply @jmitrevs! 
[...]

PR #781, opened 2023-05-03
Decouple pipeline style from strategy

### LINKED ISSUE #699: Separate top level dataflow concept from layer strategy
While updating the configuration, I think we should separate out the concept of top-level dataflow vs pipeline from the layer strategy. Currently, the `_validate_conv_strategy` could inadvertently change a Dense layer in the model from Latency to Resource. I think the concepts are different enough that they should have different configuration parameters.

vloncar:
# Description

We control the top function pipeline style with the strategy, and while this works, sometimes the change to the top pipelining is forced by a layer (the conv layer) and as a consequence this causes all other layers to use `resource` strategy which is unintended. Decoupling the strategy from the pipeline style fixes this. The explicitly set resource strategy still forces the `dataflow` style to be used. The configuration can be explicitly set with `PipelineStyle` (values can be `pipeline` and `dataflow`), but normally the users would still continue using the `Strategy` config parameter. Fixes #699 and #759.

## Type of change

## Tests

I didn't add any extra tests. The best way to test this is to verify the generated HLS has the correct pragmas and `parameters.h`. I manually tested a whole lot of combinations of this. Checking the IR in python may also be possible, but I'm not sure if this will capture all combinations.

## Checklist

### DIFF
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -279,7 +279,7 @@ class HLSConfig:
     def _validate_hls_config(self):
         use_dataflow = False
         if self.pipeline_style.lower() == 'pipeline' and self.model_compression:
-            print('WARNING: Compression enabled while model strategy set to "Latency".')
+            print('WARNING: Compression enabled while pipeline style set to "pipeline".')
             use_dataflow = True
         for layer_type, strategy in self.layer_type_strategy.items():
             if strategy.lower() == 'resource' and self.pipeline_style.lower() == 'pipeline':
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -49,6 +49,8 @@ class HLSConfig:
 
         self.trace_output = self.get_config_value('TraceOutput', False)
 
+        self.pipeline_style = 'pipeline'
+
         self._parse_hls_config()
         self._validate_hls_config()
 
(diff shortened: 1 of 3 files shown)


hls4ml/model/graph.py
```diff
@@ -274,45 +277,48 @@ def _parse_hls_config(self):
                     self.layer_name_compression[layer_name.lower()] = bool(compression)
 
     def _validate_hls_config(self):
-        use_resource = False
-        if self.model_strategy.lower() == 'latency' and self.model_compression:
+        use_dataflow = False
+        if self.pipeline_style.lower() == 'pipeline' and self.model_compression:
             print('WARNING: Compression enabled while model strategy set to "Latency".')
```
jmitrevs, line 282: The warning needs updating

jmitrevs:
Do we understand why the tests failed? Otherwise, the requested change is quite minor.

vloncar:
Not sure why the tests failed, but certainly it is not due to the changes in the code. Since this started happening recently and we didn't change the CI configuration, it may be due to the increasing number of parallel tests each creating a new environment. Not sure why it doesn't deterministically fail. We need to investigate further. You can rerun the tests or run them locally to be sure that indeed nothing got broken. I ran locally, it all passed.

jmitrevs (approved):

jmitrevs:
Unless people object, I will merge this tomorrow. But is this for 0.8 or 0.7.1?

vloncar:
How about we merge #780, fix the capitalization issue observed in #782, release that as 0.7.1 and then start merging for 0.8.0 with this PR?

PR #784, opened 2023-05-08
Configuration editor

vloncar:
# Description

Changing configuration in hls4ml requires deep understanding of the conversion process in order to be successful. New layers can be introduced, which pick up default precision definition causing problems for subsequent layers, unexpected allocation of DSPs etc. To solve this, users have to anticipate new layer names and with a meticulous trial-and-error procedure nail the initial configuration that will result in perfect conversion. This is way too cumbersome but until we have a better configuration system it is the only way. This PR proposes an alternative, to be able update the configuration of an existing hls4ml model _interactively_, via a GUI. It came out of a crazy idea I had last night after a few beers. See it in action:

Screenshot from 2023-05-08 21-26-23

It is very simple to use, make your changes and click "Update", the model will be updated and the changes should be reflected on the architecture plot on the left side.

This builds upon the recent automatic configuration support, by exposing only the configurable attributes (precision, strategy, RF etc). Some parts of it are a bit hacky for now, until we improve the configuration itself (ideally, we should have an `model.get_config()/model.set_config()` API).

The GUI is built using `PySimpleGUI` library, you can install it with `pip install pysimplegui`. Since the feature is optional, I also added an installation extra, so installation of `pip install hls4ml[gui]` will ensure the dependency is there. I considered a pure Tk implementation using `tkinter` from the standard library to avoid the new dependency, but the equivalent GUI had 400 lines of code and no support for mouse scroll events so I abandoned it.

The GUI is part of the utilities, to use it, run:
```python
model = hls4ml.converters.convert_from_keras_model(...)
hls4ml.utils.edit_model_configuration(model)
```
A GUI should open up. If you use this through SSH or VS code server it may be slow to open. I don't know what is causing this, possibly an issue with the GUI library itself. Pure `tkinter` implementation didn't have this issue.

In the future, we can expand it to support creating the _initial_ configuration, based on Keras/PyTorch/ONNX model.

## Type of change

Also fixes a minor bug in definition of the `trace` attribute (was string, should be bool) and adds `strategy` attribute to supported layers.

## Tests

We currently don't have any way of testing GUIs and this is a can of worms, so I suggest we skip it for now.

## Checklist

### DIFF
--- a/hls4ml/utils/__init__.py
+++ b/hls4ml/utils/__init__.py
@@ -1,3 +1,8 @@
 from hls4ml.utils.config import config_from_keras_model, config_from_onnx_model, config_from_pytorch_model  # noqa: F401
 from hls4ml.utils.example_models import fetch_example_list, fetch_example_model  # noqa: F401
 from hls4ml.utils.plot import plot_model  # noqa: F401
+
+try:
+    from hls4ml.utils.editor import edit_model_configuration  # noqa: F401
+except Exception:
+    pass
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -75,6 +75,16 @@ class VivadoBackend(FPGABackend):
             attrs.append(ChoiceAttribute('conv_implementation', choices=['LineBuffer', 'Encoded'], default='LineBuffer'))
             self.attribute_map[layer] = attrs
 
+        # Add strategy attribute to supported layers
+        strategy_layers = [Dense] + rnn_layers + cnn_layers
+        strategy_layers.remove(Pooling1D)
+        strategy_layers.remove(Pooling2D)
+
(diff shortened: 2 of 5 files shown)


hls4ml/utils/__init__.py
```diff
@@ -1,3 +1,8 @@
 from hls4ml.utils.config import config_from_keras_model, config_from_onnx_model, config_from_pytorch_model  # noqa: F401
 from hls4ml.utils.example_models import fetch_example_list, fetch_example_model  # noqa: F401
 from hls4ml.utils.plot import plot_model  # noqa: F401
+
+try:
```
jmitrevs, line 5: I wonder if there's a better way to do this because now if the import fails, you just gen an error while using it of:
```
AttributeError: module 'hls4ml.utils' has no attribute 'edit_model_configuration'
```
ideally the error message would be better.

PR #786, opened 2023-05-10
Update Jenkins test environment to avoid dependency hell

vloncar:
# Description

Jenkins tests are based on python 3.7 environment, in which we can install TF 2.11 as the latest version. When hls4ml is installed on top, `onnx` dependency causes the update of `protobuf` package (to 4.21) causing the environment to break and hls4ml (and tensorflow and qkeras) are unusable. We should consider updating the Python requirement to 3.8 for the next version of hls4ml to avoid this, but as an immediate band-aid solution to ensure tests run, ~~we can install `onnx==1.12` prior to hls4ml, that won't cause the update of `protobuf` and the environment works.~~ we can update just the Jenkins environment to Python 3.8.

## Type of change

## Tests

Jenkins tests should run now, they fail on every branch currently.

## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
   agent {
     docker {
-      image 'vivado-el7:2'
+      image 'vivado-el7:3'
       args  '-v /data/Xilinx:/data/Xilinx'
     }
   }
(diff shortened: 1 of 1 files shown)


vloncar:
After a discussion, updated the environment to python 3.8

jmitrevs (approved):

PR #795, opened 2023-05-15
Remove tf_to_hls

vloncar:
# Description

This PR removes `tf_to_hls`. The `tf_to_hls` converter is outdated, unused, not exposed through the main API and based on old approach that likely won't work anymore. If we decide to support lower-level tensorflow API alongside Keras, this approach would be different anyway, so there's no reason to keep this around.

A question to be addressed: this converter introduced `BiasAdd` as a temporary layer and the associated optimizer. If ONNX and PyTorch converters don't plan to use it, it should be removed in this PR as well.

## Type of change

- [x] Other (Specify) - Cleanup

## Tests

Nothing to test.

## Checklist


jmitrevs:
I don't think the failure is transient. The tests succeed on my computer.

jmitrevs (approved):

PR #800, opened 2023-06-01
Discussion - Inlined Conv slows down latency significantly (up to x15 - x20)

bo3z:
# Description
> * While testing some code unrolling for the hls4ml Optimisation API, I noticed that inlining in Conv2D can allocate unnecessary RAM.
> * When tested on the current version of Conv2D (line buffer, streaming, Resource strategy, RF > 1), there is a significant difference in latency (between 3x and nearly 20x)
> * Still unsure what cause this bug and if it is present for (i) Latency strategy, (ii) RF = 1 and (iii) encoded convolution. But is certainly seems that for RF > 1 in Resource to seem a bug. Opening this as a discussion until further synthesis results are obtained.

## Type of change
- [x] Bug fix
- [x] Breaking change (potentially)
- [x] Discussion

## Tests

>* Below are report files following a full Vivado synthesis and CoSim analysis, for the SVHN paper model, with RF = 9
>* Underscored _master, corresponds to implementations of the current, line-buffer, resource, streaming Conv2D
>* Underscored _no_pragma, corresponds to implementations with the inline keyword removed, as per the PR
>* Inspecting the report files, the models are clearly equivalent (in terms of HLS config and architecture) as they use the same number of DSPs and BRAM and similar utilisation of LUT & FF. However, latencies differ up to 20x.
>* Source of report files: https://cernbox.cern.ch/s/DK4v2KUTiBmFvYN 

## Checklist

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_conv_stream.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_conv_stream.h
@@ -260,7 +260,7 @@ void compute_output_buffer_2d(
     hls::stream<res_T> &res_stream,
     typename CONFIG_T::weight_t weights[CONFIG_T::kernel_size * CONFIG_T::n_chan * CONFIG_T::n_filt],
     typename CONFIG_T::bias_t biases[CONFIG_T::n_filt]) {
-    #pragma HLS INLINE
+    #pragma HLS INLINE OFF
 
     // Thresholds
     const static int lShiftX = CONFIG_T::filt_width - 1;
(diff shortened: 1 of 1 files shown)


jmitrevs:
The changes are pretty minimal and should do no harm. If they fix an issue, I am not opposed to merging them, even if we don't fully understand why.

jmitrevs (approved):

jmitrevs:
I think I'll go ahead and merge this since it causes no problems and seems to help.

PR #804, opened 2023-06-08
Make binary CNN match between Keras and hls4ml

jmitrevs:
# Description

The check on the pytest for binary CNNs had to be disabled because it failed. This enables it, and attempts to fix issues that it discovers

## Type of change

## Tests

The test checking has been enabled. It still fails in most cases.

## Checklist


jmitrevs:
This now matches for io_parallel implementations with the Latency strategy. Resource still fails (potentially for unrelated reasons also on Quartus) and for io_stream. @vloncar , I am curious if you think there's a better way to preserve the XnorPrecision designation when doing MaxPooling than I am doing here. It is a bit related to precision propagation, but I think in this case you really need to preserve the type for the network to make sense.

jmitrevs:
With the latest push, this works for Vitis and Vivado. Note that #817 was triggered with the test previously, which has been modified to not trigger the problem. However, a better fix for #817 would be preferred. It is somewhat beyond the scope of this PR, though.

jmitrevs:
The way Winograd is written it doesn't support the xnor types, so I have forced the Quartus io_parallel tests to use im2col. If we want to update Winograd to do binary, I'll leave that as a separate development.

jmitrevs:
Don't fully understand why the pytests fail here but succeeded on my laptop. I thought I was using the same random number, but maybe not.

jmitrevs:
I think the issue was just accumulator precision in the test (big number - almost equal big number) and then quantization based on > or < 0. Let's see if increasing the bitwidth makes the tests succeed.

test/pytest/test_binary_cnn.py
```diff
@@ -12,9 +12,22 @@
...
+        ('Vivado', 'io_parallel', 'latency'),
+        ('Vivado', 'io_stream', 'latency'),
+        ('Vivado', 'io_stream', 'resource'),
+        ('Vitis', 'io_parallel', 'resource'),
+        ('Vitis', 'io_parallel', 'latency'),
+        ('Vitis', 'io_stream', 'latency'),
+        ('Vitis', 'io_stream', 'resource'),
+    ],
+)
+def test_model2(backend, io_type, strategy):
```
bo3z, line 30: Very minor comment - I assume the name `test_model2` is left-over from way before - we should maybe change it, the name could be a bit more informative (even though it is already in the correct file)

test/pytest/test_binary_cnn.py
```diff
@@ -12,9 +12,22 @@
...
     x_in = Input(shape=(28, 28, 1))
 
     x = QConv2D(4, (3, 3), kernel_quantizer="binary", name="conv2d_1", kernel_regularizer=l2(0.0001), use_bias=False)(x_in)
```
bo3z, line 33: Can bias be enabled? The emphasis here I guess is on the multiplications and pooling precision, but just to confirm it works.

test/pytest/test_binary_cnn.py
```diff
@@ -73,5 +90,4 @@ def test_model2(backend, io_type):
     y = model2.predict(X)  # noqa: F841
     y_hls = hls_model.predict(X)  # noqa: F841
 
-    # # TODO:  enable the comparions after fixing the remaing issues
-    # np.testing.assert_allclose(np.squeeze(y_hls), np.squeeze(y), rtol=1e-2, atol=0.01)
+    np.testing.assert_allclose(np.squeeze(y_hls), np.squeeze(y), rtol=1e-2, atol=0.01)
```
bo3z, line 93: The old version of this test seems to have had one data point to verify functionality between QKeras and hls4ml - can that be increased, to maybe a 100? Increasing also removes the need for `np.squeeze`.

test/pytest/test_binary_cnn.py
```diff
@@ -37,28 +50,32 @@ def test_model2(backend, io_type):
     x = QBatchNormalization()(x)
     x = QActivation("binary_tanh", name="act4")(x)
 
-    x = QDense(10, kernel_quantizer="binary", activation="softmax", name="q_dense_7", use_bias=False)(x)
+    x = QDense(10, kernel_quantizer="binary", activation="linear", name="q_dense_7", use_bias=False)(x)
```
bo3z, line 53: Does Softmax break binary CNN?

bo3z (changes requested):
Changes look good to me - main comments on the test setup, that was left-over from the previous changes.

bo3z:
> The way Winograd is written it doesn't support the xnor types, so I have forced the Quartus io_parallel tests to use im2col. If we want to update Winograd to do binary, I'll leave that as a separate development.

Enabling binary support for Winograd needs to change (i) the explicit multiplication for binary weights, to use the product function from `nnet_mult` and (ii) for binary inputs, see how the addition of input elements behaves. But Winograd's algorithm is really meant to reduce the number of multiplications/DSP utilisation and with binary CNNs, that's not really a concern, so I don't think it is necessarily important in this case.

test/pytest/test_binary_cnn.py
jmitrevs: I agree. I will change it. It was the name of the model as discussed in a issue.

test/pytest/test_binary_cnn.py
jmitrevs: It doesn't break it, but the implementations differ slightly, so it's hard to get exact match. The legacy implementation got close--maybe exact. The others were off. But that didn't seem to be the focus of this PR, so I went for the simpler linear activation.

test/pytest/test_binary_cnn.py
jmitrevs: Agree.

test/pytest/test_binary_cnn.py
jmitrevs: I am not sure if bias is supported in binary models, especially if both the weight and input are xnor. I wasn't sure how the bias would work. Let me look at it a bit more. Xnor bias (add or subtract one) didn't seem to make much sense to me, but I am no expert in binary networks. But maybe in the case when only the weights are binary but the inputs aren't it makes sense to have a quantized bias?

jmitrevs: So 'binary' bias implemented as `XnorPrecisonType` does not work, but I don't think that's really useful. Now `quantized_bits` or standard unquantized bias I think works with fixed or integer inputs and binary (xnor) weights, but I am not sure with xnor inputs and outputs. The complication, if I understand the code, is that the bias addition in the code happens before the cast:
```c++
inline typename std::enable_if<std::is_same<data_T, ap_uint<1>>::value &&
                                   std::is_same<typename CONFIG_T::weight_t, ap_uint<1>>::value,
                               ap_int<nnet::ceillog2(CONFIG_T::n_in) + 2>>::type
cast(typename CONFIG_T::accum_t x) {
    return (ap_int<nnet::ceillog2(CONFIG_T::n_in) + 2>)(x - CONFIG_T::n_in / 2) * 2;
}
```
while logically it should happen after. I think the bias is doubled, and the type calculation doesn't include the bias. But let me know if I am wrong. It's easy to lose track when following the code.

So I think I will add bias in all cases except when both the inputs and weights are xnor type.

jmitrevs:
I see there are some errors. Will make it draft till I fix them.

jmitrevs:
The failures seem to be due to precision of the quantization, so I am removing the draft designation

bo3z:
Looks good to me.

jmitrevs:
The test passed, so I think we can maybe merge this.

vloncar (approved):

Issue #798, opened 2023-05-22
Vivado synthesis report - zero BRAM utilisation (OOC)


bo3z:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [ ] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [ ] Check that the issue hasn't already been reported, by checking the currently open issues.
- [ ] If there are steps to reproduce the problem, make sure to write them down below.
- [ ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
Please give a brief and concise description of the bug.

## Details
Please add to the following sections to describe the bug as accurately as possible.

### Steps to Reproduce
Add what needs to be done to reproduce the bug. Add *commented* code examples and make sure to include the original model files / code, and the commit hash you are working on.

1. Clone the hls4ml repository
2. Checkout the master branch, with commit hash: [...]
3. Run conversion [...] on model file with code [...]
4. [Further steps ...]

### Expected behavior
Please add a brief description of what you expected to happen.

### Actual behavior
Describe what actually happens instead.

### Possible fix
If you already know where the issue stems from, or you have a hint please let us know.

PR #768, opened 2023-04-16
hls4ml Optimization API [Part 1]

bo3z:
This pull request introduces the first part of the hls4ml Optimization API - an automated workflow for hardware-aware model compression. By formulating pruning and weight sharing as a linear optimisation problem, the workflow iteratively selects redundant weights, considering the overall impact on hardware. The tool currently supports Keras and QKeras models, as well as various hardware objectives on GPUs (FLOPs) and FPGAs, with a Vivado hls4ml backend (DSP, BRAM, FF). However, the tool is both hardware- and framework-agnostic - most of the concepts readily generalise to other frameworks (e.g. PyTorch) and other hardware (e.g. Quartus backend). This allows end users to write custom objectives (e.g. Quartus latency optimisation), following a similar template.

Furthermore, this tool aims to bridge the gap between other libraries for model compression, such as TensorFlow Model Optimization and QKeras. The tool is directly integrated with QKeras and an updated version of Keras Surgeon, to aid model compression. Finally, this tool provides out-of-the-box support for structured pruning (filters, neurons), as well as gradient based ranking methods.

The exact implementation and motivations are further explained in the attached presentation. Initial results are shown on both classification and regression with various objectives including,  sparsity, GPU FLOP reduction, Vivado DSPs and FFs. Since this is a large PR, it is recommended to review the commits one by one, as each commit is self-contained and can be checked out by itself. They are briefly explained below.

## Supporting document and presentation
Available at: https://indico.cern.ch/event/1278049/ 

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] A new research paper code implementation

## Description
Contributions:
* A new pattern pruning approach, inspired by the parallelism of FPGAs. For more information on the inspiration behind this approach, see the attached presentation. 
* Easy, out-of-the-box support for structured pruning and weight sharing.
* Formulation of pruning as a Knapsack optimisation problem, relative to a hardware objective - maximise network performance while minimising some hardware resource(s)
 * Integration with Keras Surgeon and extended support for QKeras, to further reduce model footprint.
 * End-to-end flow for model optimisation, including an extensible & open-source library for hardware-aware and agnostic pruning of Keras & TensorFlow models.

## Tests
 * Eight new unit tests were written in the `PyTest` framework. These tests are stored under `hls4ml/test/pytest/optimization`. Each test follows a single addition to the framework and are better explained by the individual commits.
[...]

PR #809, opened 2023-06-13
hls4ml Optimization API [Part 2]

bo3z:
# Description
>
> * Second part of hls4ml Optimization API #768 
> * Introduces Dense Unrolled layers, optimising multiplications with zero in Resource strategy with RF > 1
> * Introduces additional TCL scripts, to optimise zero BRAM blocks.

## Type of change
- [x] Fix issue #798 

## Checklist


bo3z:
I will add pre-commit additionally, last time I ran it, some tests were broken, so will add it a subsequent commit.

jmitrevs:
We merged part 1. Should we merge part 2?

vloncar:
I'm reviewing it. Slowly :smiley: . But it's next in line, then HGQ.

jmitrevs:
The pytest error is unrelated to the PR so from my side this can be merged. I'll let Vladimir give the last OK.

hls4ml/templates/vivado/vivado_synth.tcl
calad0i, line 1: Why switching the default impl to verilog here? What's the impact of the added optimization on firmware performance and synth time?

hls4ml/writer/vivado_writer.py
```diff
@@ -171,10 +179,18 @@ def write_project_cpp(self, model):
                     newline += indent + '#pragma HLS INTERFACE ap_vld port={},{} \n'.format(
                         ','.join(all_inputs), ','.join(all_outputs)
                     )
-                    if model.config.pipeline_style.lower() == 'dataflow':
-                        newline += indent + '#pragma HLS DATAFLOW \n'
+
+                    model_cfg = model.config.get_config_value('HLSConfig')['Model']
+                    if (
```
calad0i, line 184: It seems that if unrolled resource conv layers are used, dataflow will not be used. Also, if `n_partition` is not 1, the global II won't the be larger than the internal dense II (`reuse_factor`)?

vloncar:
This PR was refactored to introduce the new unrolled implementation as a "strategy", to be an alternative to existing latency and resource strategies. This allowed the the matrix-vector multiplication kernel to be used as a function, simplifying the integration with the rest of the code. The PR also has changes to the top pipeline style pragma, so the config now includes a new "auto" option (the default) which allows the optimizer to choose the best one. All of pipeline style decisions are now made in the new optimizer, instead of being scattered around the HLSConfig class and the backend.

One more minor change may come. Since we will have multiple new strategies and optimization options, it was suggested to give this optimization technique a name and move it to a submodule of that name. Discussion on this is welcome.

hls4ml/templates/vivado/vivado_synth.tcl
bo3z: There shouldn't be any noticeable impact. There are 2 reasons this change was made:

1. When you use VHDL with Resource strategy the synthesis report gives 0 BRAM (which cannot be the case at all, especially since Verilog shows the correct BRAM). I suspect this is something to do out-of-context synthesis in Vivado but I haven't figured out how to enable / disable it. This is explained more in:  https://github.com/fastmachinelearning/hls4ml/issues/798

2. It's "functionally" not correct to do co-simulation with Verilog and synthesis with VHDL (since HLS always generates both formats anyway). Yes, if one is correct and the other is wrong then it's an issue with the HLS compiler and we can't do much about it. But my reasoning is that using we should test and "deploy" the same set of end-files.

calad0i: Then this should probably go into a separate PR I guess, as it changs the behavior of the library while not using dsp aware pruning. How about let's isolate https://github.com/bo3z/hls4ml/commit/e53ec9c1c2673092b126101d544f18d96b7a692c into another PR?

vloncar:
All the comments from the last discussion were addressed (renamed the strategy and moved the files to a new submodule). It's now ready to be merged

JanFSchulte (approved):

PR #815, opened 2023-06-16
Fix Vitis Conv1D/2D latency strategy

### LINKED ISSUE #805: [Vitis hls] Cannot apply array transformation pragma/directive because of full array load/store.
## Quick summary

I was trying to convert a quantized Conv1D model using Vitis HLS, and was seeing this error:

```
ERROR: [HLS 214-384] in function 'void nnet::conv_1d_latency_cl<ap_fixed<9, 3, (ap_q_mode)5, (ap_o_mode)3, 0>, ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>, config2>(ap_fixed<9, 3, (ap_q_mode)5, (ap_o_mode)3, 0>*, ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>*, config2::weight_t*, config2::bias_t*) (.304)': Cannot apply array transformation pragma/directive because of full array load/store. (firmware/nnet_utils/nnet_conv1d_latency.h:81:11)
```

## Details

I'm using the current master hls4ml branch where the lastest commit is:

```
commit 2e71ff451cc36ce8e7319a92b65a5d9be8bff427 (HEAD -> main, origin/main, origin/HEAD)
Merge: eca1ea37 9d2e6640
Author: Jovan Mitrevski <jmitrevs@fnal.gov>
Date:   Tue May 16 13:55:19 2023 -0500

    Merge pull request #796 from fastmachinelearning/pre-commit-ci-update-config
```

### Steps to Reproduce
I included the `.h5` file of the model [here.](https://www.dropbox.com/s/r6pxc8uttonqgny/QModel_9bit_150.h5?dl=0) 

And a script to convert the model [here. ](https://www.dropbox.com/s/v0gb8s70dkq9sk8/hls_9bit_model.py?dl=0)

You could run the conversion using:
 ```
python hls_9bit_model.py
```

I'm not really sure what's happening here ...

vloncar:
# Description

This fixes #805. Vitis HLS compiler has problems with synthesizing pointer expressions like `*(res++)`. Using array indexing resolves the compilation issue, but then the design fails to meet timing, due to the inlining of the `conv_1d_latency_cl` (ditto 2d) function. This seems to be on a per-layer (or model) basis, since it works sometimes. It's best to disable it to meet timing in the general case, but then people can explore enabling it to save a few cycles. I left a comment along those lines. I implemented it only for Vitis, but if people encounter the issue in Vivado we can use the same fix there, again bringing the backends to single-source HLS implementation.

Also, I used the opportunity to sneak a commit that fixes the Vitis pragma warnings regarding `DATA_PACK` pragma. This seems to be a leftover from a bad merge. The workaround used brings it in line with the rest of the codebase where we know it works.

## Type of change

## Tests

Only synthesis will confirm this is the fix, and we don't (yet) have synthesis tests for Vitis, so this one is _trust me bro_ type of PR. But @Duchstf will properly test it and report back, right Duc?

## Checklist

I did all the things in the checklist.

### DIFF
--- a/hls4ml/templates/vitis/nnet_utils/nnet_conv1d.h
+++ b/hls4ml/templates/vitis/nnet_utils/nnet_conv1d.h
@@ -1,0 +1,68 @@
+#ifndef NNET_CONV1D_H_
+#define NNET_CONV1D_H_
+
+#include "nnet_common.h"
+#include "nnet_conv1d_latency.h"
+#include "nnet_conv1d_resource.h"
+#include <cstdlib>
+
+namespace nnet {
+
+struct conv1d_config {
+    // Internal data type definitions
+    typedef float bias_t;
+    typedef float weight_t;
+    typedef float accum_t;
+
+    // Convolutional parameters
+    static const unsigned pad_left = 0;
+    static const unsigned pad_right = 0;
+    static const unsigned in_width = 10;
+    static const unsigned n_chan = 0;
+    static const unsigned filt_width = 1;
+    static const unsigned kernel_size = filt_width;
+    static const unsigned n_filt = 1;
+    static const unsigned stride_width = 1;
+    static const unsigned dilation = 1;
+    static const unsigned out_width = 10; //(N_IN + PAD_LEFT * PAD_RIGHT - (DILATION * (FILT_WIDTH - 1) + 1)) / STRIDE + 1
+
+    static const unsigned reuse_factor = 1;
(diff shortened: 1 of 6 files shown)


Duchstf:
I trust @vloncar on this PR, see #805.

jmitrevs (approved):

PR #827, opened 2023-06-30
Fix loading weights in GarNetStacked and GarNet internal array precisions

joshlerner:
## Description

In the change from using the reader to storing weights as attributes, GarNetStack input feature weights + biases and output feature weights were missed. Fixed by storing all GarNetStack weights/biases as attributes.

All non default precisions specified for internal GarNet arrays (edge weight, norm, etc.) were not converted to CPP definitions, and produced typedef errors in `firmware/parameters.h`. Fixed by applying an APTypeConverter to all internal array precisions, not just those with default values.

Modified `contrib/garnet.py` to include an output activation for GarNetStack models, which was necessary to test above changes. This had previously been commented out due to being unused.

## Type of change

## Tests

Added test similar to pre-existing in `test_garnet.py` for GarNetStack models
> * GarNet internal arrays are included in the generated config automatically for `name` and `type` granularity
> * Previous tests overwrote these specifications, but the new test includes non-default internal arrays in the config

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ docs/_build
 docs/autodoc/*
 hls4mlprj_*
 *~
+*.ipynb_checkpoints/
--- a/contrib/garnet.py
+++ b/contrib/garnet.py
@@ -314,8 +314,7 @@ class GarNetStack(GarNet):
                 )
 
                 if self._output_activation is None or self._output_activation == "linear":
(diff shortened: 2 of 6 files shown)


joshlerner:
Hello! I have been working on a GarNet calibration model to be run on an FPGA using hls4ml. This is my first pull request, so some of the formalities are new to me. Thank you for your patience!

PR #834, opened 2023-07-17
Add QDepthwiseConv2D, DepthwiseConv2D, DepthwiseConv1D support

jmitrevs:
# Description

There was a request to add support for `QDepthwiseConv2D` for QKeras. Previously we had Keras support for `DepthwiseConv1D` and `DepthwiseConv2D` but only as part of separable convolutions. This PR adds support for standalone `DepthwiseConv1D` and `DepthwiseConv2D`, and also `QDepthwiseConv2D`. Currently only `depth_multiplier=1` is supported.

Note that QKeras doesn't support `QDepthwiseConv1D`. 

Adding this found an issue in that arguments to activations passed as parameters to, e.g. `QDense`, such as `activation='quantized_relu(3, 0)'`, were ignored. This had the same effect as `activation='quantized_relu()'`.  

## Type of change

## Tests

Tests added to both `test_keras_api.py` and `test_qkeras.py`

## Checklist


jmitrevs:
I should also mention that this is only for 'io_stream` and only for Vivado and Vitis. There is not Quartus support.

bo3z (commented):
Looks good to me - for some reason the PyTests weren't triggered, so probably need to be triggered manually.

As a separate PR it could be worth including a Resource implementation of depth-wise multiplications and parameterise the new tests for both strategies.

hls4ml/templates/vivado/nnet_utils/nnet_sepconv1d_stream.h
```diff
@@ -57,6 +57,21 @@ void depthwise_conv_1d_buffer_cl(hls::stream<data_T> &data, hls::stream<res_T> &
     }
 }
 
+template <class data_T, class res_T, typename CONFIG_T>
```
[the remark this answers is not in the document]
jmitrevs: You are right. The reason of course historic, since the function on 112 -122 existed before. But I think I'll make the suggested change.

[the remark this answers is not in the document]
vloncar: We should check if using it this way it get inlined or not, otherwise it costs a cycle.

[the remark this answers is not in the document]
jmitrevs: Anyway, I pushed the change. I expect that it would cause no difference in the produced code.

bo3z:
> @bo3z, are there any other things needed for this PR?

Looks good to me. Maybe worth opening an issue to keep track of the TODOs from this PR (resource strategy and multiplier != 1).

bo3z (approved):

test/pytest/test_keras_api.py
```diff
@@ -297,6 +299,58 @@ def test_conv2d(chans, padds, backend, io_type):
...
+    '''
+    Test proper handling of DepthwiseConv2D
+    '''
+    X = np.random.rand(10, 32, 32, 3)
+    X = np.round(X * 2**10) * 2**-10  # make it an exact ap_fixed<16,6>
+    model = tf.keras.models.Sequential()
+    model.add(DepthwiseConv2D(kernel_size=(3, 3), input_shape=(32, 32, 3)))
+    model.compile()
+
+    config = hls4ml.utils.config_from_keras_model(model, granularity='name', default_precision='fixed<32,12>')
```
vloncar, line 315: If the input is clipped to exactly match <16,6>, why do we need to increase the precision this way? Should only `accum` be increased?

test/pytest/test_keras_api.py
```diff
@@ -297,6 +299,58 @@ def test_conv2d(chans, padds, backend, io_type):
...
+
+    np.testing.assert_allclose(y_qkeras, y_hls4ml.reshape(y_qkeras.shape), rtol=1e-2, atol=0.01)
+
+
+# Currently only Vivado and Vitis is supported for io_stream.
+@pytest.mark.parametrize('backend', ['Vivado', 'Vitis'])
+@pytest.mark.parametrize('io_type', ['io_stream'])
+def test_depthwise1d(backend, io_type):
+    '''
+    Test proper handling of QConv2DBatchnorm.
```
vloncar, line 333: typo :wink:

test/pytest/test_qkeras.py
```diff
@@ -400,6 +401,48 @@ def test_qconv2dbn(randX_100_8_8_1, backend, io_type):
...
+    '''
+    X = randX_10_32_32_3
+    X = np.round(X * 2**10) * 2**-10  # make it an exact ap_fixed<16,6>
+    model = Sequential()
+    model.add(
+        QDepthwiseConv2D(
+            kernel_size=(3, 3),
+            input_shape=(32, 32, 3),
+            depthwise_quantizer='quantized_bits(6, 0, alpha=1)',
+            depthwise_initializer='ones',
```
vloncar, line 425: Isn't this test cheating a bit? All the weights will be 1s and the biases 0s, no point in using quantizers. It doesn't really test the proper handling of quantizers and the accuracy of the implementation.

test/pytest/test_keras_api.py
jmitrevs: Will fix. (same is test_qkeras.)

test/pytest/test_keras_api.py
jmitrevs: It's not clipped for the keras input, just the hls4ml input. The input is actually all in the [0,1) range. I didn't try to optimize the precision in the model and just sought to make it irrelevant.

test/pytest/test_qkeras.py
jmitrevs: That's a leftover from me playing around with the initializer. Let me find better ones. For depthwise, the default is probably good ("he_normal"), so I'll move to that. The default for bias is 0. Let me see if I should change it.

bo3z: Probably worth changing the bias initialiser - just to be sure the bias is not ignored in the C++ implementation.

georgenardeswebphy: Hi @jmitrevs  . 
Did you tested with large kernel size?
For example: model.add(DepthwiseConv2D(kernel_size=(16,16), input_shape=(32, 32, 24))).

I'm facing some errors regarding parameters that are bigger than the int256 representation in the parameters.h file.

PR #845, opened 2023-08-04
No longer make ExponentPrecisionType and XnorPrecisionType inherit from IntegerPrecisionType

jmitrevs:
# Description

When trying to propagate precisions, `isinstance(var, (IntegerPrecisionType, FixedPrecisionType))` produces incorrect results if the variable is `ExponentPrecisionType` or `XnorPrecisionType`. The "is a" requirement for inheritance is violated. Therefore it's better to remove the inheritance, which this PR does. It also cleans up a bit of the remaining string parsing there was and switches from the old `%` string formatting to f-strings or `.format()`. 

## Type of change

It is not really a "bug" but the previous inheritance structure caused issues with yet to be added precision propagation.

## Tests

This should not break anything. The qkeras tests in particular are the main tests for this.

## Checklist

### DIFF
--- a/hls4ml/model/types.py
+++ b/hls4ml/model/types.py
@@ -5,7 +5,6 @@ The Precision types are given names for convenience (``NamedType``). Named types
 higher-dimensional tensors, which are defined as arrays or FIFO streams in the generated code.
 """
 
-import re
 from enum import Enum
 
 import numpy as np
@@ -324,7 +323,7 @@ class XnorPrecisionType(PrecisionType):
         super().__init__(width=1, signed=False)
 
     def __str__(self):
-        typestring = 'xnor'
+        typestring = 'uint<1>'
         return typestring
 
 
@@ -338,7 +337,7 @@ class ExponentPrecisionType(PrecisionType):
         super().__init__(width=width, signed=signed)
 
     def __str__(self):
-        typestring = '{signed}exp<{width}>'.format(signed='u' if not self.signed else '', width=self.width)
+        typestring = '{signed}int<{width}>'.format(signed='u' if not self.signed else '', width=self.width)
         return typestring
 
 
@@ -548,7 +547,7 @@ class WeightVariable(Variable):
         if not self._iterator.finished:
             value = self._iterator[0]
             self._iterator.iternext()
-            return self.precision_fmt % value
+            return self.precision_fmt.format(value)
         else:
             raise StopIteration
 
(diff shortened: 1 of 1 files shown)


jmitrevs:
About the pytest failure, Quartus Winograd is broken for Xnor inputs. #804 forces im2col instead of Winograd for this reason. I can hide this failure if desired.

hls4ml/model/types.py
```diff
@@ -311,16 +314,20 @@ def __eq__(self, other):
...
+class XnorPrecisionType(PrecisionType):
     """
     Convenience class to differentiate 'regular' integers from BNN Xnor ones
     """
 
     def __init__(self):
         super().__init__(width=1, signed=False)
 
+    def __str__(self):
+        typestring = 'uint<1>'
```
jmitrevs, line 326: I am truthfully not much of a fan of the typestring. What is it's function? There is not a 1:1 between a typestring and and the type. Part of me thinks this should be 'xnor' and not 'uint<1>', though this requires a few more changes downstream.

hls4ml/model/types.py
```diff
@@ -311,16 +314,20 @@ def __eq__(self, other):
...
 
 
-class XnorPrecisionType(IntegerPrecisionType):
+class XnorPrecisionType(PrecisionType):
     """
     Convenience class to differentiate 'regular' integers from BNN Xnor ones
     """
 
     def __init__(self):
         super().__init__(width=1, signed=False)
```
jmitrevs, line 323: Logically xnor is signed, even if we represent it as unit<1>. How it gets implemented should be a backend issue, not a definition issue.

jmitrevs:
I think after #804 is merged the pytest failure will go away. But I am curious what people think about the "review" comments I made.

hls4ml/model/types.py
vloncar: Not 100% sure, but this may be the last leftover from the previous way of doing things. I think that it is only used in `WeightVariable.update_precision()` to set a new string format that is then not used anywhere (it was used only in `print_array_to_cpp` in the writer, but this was changed). If it doesn't hurt your eyes too much I would leave it as-is here, and have a follow up PR that removes it, `size_cpp()` and perhaps the whole `dim_names` concept as one of our usual "cleanup" PRs that we do prior to a release.

hls4ml/model/types.py
vloncar: I think the XNOR idea was that it is a single bit implementation, the more general definition of BNN is signed (-1 and 1), but that's a different case.

vloncar (approved):

PR #848, opened 2023-08-11
Bug fix for named nn.Sequential in pytorch parser

JanFSchulte:
Parsing of nn.Sequentials that are named members of a model class results in a naming convention for the tensors in the `state_dict` of the model different from what the parser expects, since it was so far tested only on unnamed nn.Sequentials. This PR catches this and adjusts the name of the tensors we are importing from the `state_dict` accordingly. A test is added to ensure that we keep parsing both cases successfully. 

## Type of change

## Tests

To reproduce, this will fail with this PR:

```

import torch.nn as nn

from hls4ml.converters import convert_from_pytorch_model
from hls4ml.utils.config import config_from_pytorch_model



#simple model with namend sequential
class SeqModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.layer = nn.Sequential(
          nn.Conv2d(1,20,5),
          nn.ReLU(),
          nn.Conv2d(20,64,5),
          nn.ReLU()
        )   

    def forward(self, x):
        output = self.layer(x)
        return output

model = SeqModel()

config = config_from_pytorch_model(model)
output_dir = 'test_pytorch'

convert_from_pytorch_model(
        model, (None, 1, 5, 5), hls_config=config, output_dir=output_dir)
```

pytests have been added to verify that this keeps working. 

## Checklist

### DIFF
--- a/hls4ml/converters/pytorch_to_hls.py
+++ b/hls4ml/converters/pytorch_to_hls.py
@@ -33,7 +33,7 @@ class PyTorchModelReader:
         tensorName = layer_name + '.' + var_name
         if "layer." in tensorName and not layerInKeyName:
             tensorName = tensorName.lstrip("layer.")
-        print(tensorName)
+
         if tensorName in self.state_dict:
             data = self.state_dict[tensorName].numpy()
             return data
--- a/hls4ml/converters/pytorch_to_hls.py
+++ b/hls4ml/converters/pytorch_to_hls.py
@@ -23,6 +23,7 @@ class PyTorchModelReader:
         # were an named nn.Sequential with "layer" in it's name is used.
         # We make sure that we pick the correct name for the tensor
         layerInKeyName = False
+
         for key in self.state_dict.keys():
             if "layer." in key:
                 layerInKeyName = True
@@ -37,7 +38,11 @@ class PyTorchModelReader:
         if tensorName in self.state_dict:
             data = self.state_dict[tensorName].numpy()
             return data
-
+        # if a layer is reused in the model, torch.FX will append a "_n" for the n-th use
+        # have to remove that integer to find the tensors
+        elif '.'.join(layer_name.split('.')[:-1]) + '.' + var_name in self.state_dict:
+            data = self.state_dict['.'.join(layer_name.split('.')[:-1]) + '.' + var_name].numpy()
+            return data
         else:
             return None
 
--- a/hls4ml/converters/pytorch_to_hls.py
+++ b/hls4ml/converters/pytorch_to_hls.py
@@ -18,19 +18,24 @@ class PyTorchModelReader:
         data = None
 
         # Workaround for naming schme in nn.Sequential,
-        # have to remove the prefix we previously had to add to make sure the tensors are found
-        if 'layer_' in layer_name:
-            layer_name = layer_name.split('layer_')[-1]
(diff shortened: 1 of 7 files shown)


hls4ml/converters/pytorch_to_hls.py
```diff
@@ -18,17 +18,24 @@ def get_weights_data(self, layer_name, var_name):
...
+        # We have to remove the prefix we added earlier to not have layer
+        # names start with an underscore. However, this clashes with cases
+        # were an named nn.Sequential with "layer" in it's name is used.
+        # We make sure that we pick the correct name for the tensor
+        layerInKeyName = False
+        for key in self.state_dict.keys():
+            if "layer." in key:
+                layerInKeyName = True
+
+        if '_' in layer_name:
```
jmitrevs, line 30: I am not sure I follow the logic here, but I am not a pytorch expert. Can you double-check?

JanFSchulte: I had another look and this is a bit convoluted, but it works. The issue is that the layers inside a torch.nn.Sequential get named in the pattern `nameOfSquential_n` where `n` just numbers the layers inside the sequential. If the sequential does not have a name, which happens if you just go `model = nn.Sequential(..)`, the layers will have names that just start with an underscore, which hls4ml doesn't like. So we add the prefix "layer" to those in the loop over the layers, which we have to remove when we go and load the tensors. The changes in this PR account for the fact that someone could create a named nn.Sequential just named `layer`, which then clashes with our previous assumption that if a layer name starts with `layer_` it is because we added it by hand to get around the issue with layer names starting with just an `_`.

A further complication is that while torch.FX reports these structured layer names with underscores while in the `state_dict` a `.` is used, so we have to replace the `_` with `.` here. The last complication is the case of layers that used multiple times in the same model. torch.FX reports them as different layers, adding an `_n` for the n-th time it is used to the layer name. But the tensors in the state dict will not have those modifications, so we also have to account for this. 

Bit of a mess, but I have tested all these cases and this implementations catches all edge cases that I'm aware off.

JanFSchulte:
pre-commit.ci autofix

hls4ml/converters/pytorch_to_hls.py
JanFSchulte: This has now been solved significantly nicer by Vladimir :)

vloncar:
This now includes the changes from #840. There'll be a follow-up PR with the remaining bits we need to parse sPHENIX tracking GNN.

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -98,7 +88,7 @@ def decorator(function):
     'elu': 'ELU',
     'prelu': 'PReLU',
     'sigmoid': 'Sigmoid',
-    'layer_threshold': 'Threshold',
+    '_threshold': 'Threshold',
```
jmitrevs, line 91: I just wanted confirmation that this is supposed to have the leading underscore.

vloncar: Yes. Once we get rid of prefixing with `layer` this one indeed remains as `_threshold`. It corresponds to the [torch.functional call](https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py#L1431). I don't understand the reasoning behind the naming :smiley:

jmitrevs (approved):

PR #851, opened 2023-08-18
Add option to configure IP version

AdrianAlan:
A small change of adding an option to specify the IP version, defaulting to `1.0.0`.

Motivation: although it is a long time since the [Y2K22](https://support.xilinx.com/s/article/76960?language=en_US ) issue I hear people still run into it when running on systems where admins didn't install the patch. The same workaround was suggested in [here](https://support.xilinx.com/s/question/0D52E00006vDlvJSAS/export-ip-invalid-argument-revision-number-overflow-issue?language=en_US). Issue mentioned in #633.

## Type of change

## Tests

Please suggest tests for this change.

## Checklist

### DIFF
--- a/hls4ml/templates/vivado/build_prj.tcl
+++ b/hls4ml/templates/vivado/build_prj.tcl
@@ -229,7 +229,7 @@ if {$opt(validation)} {
 if {$opt(export)} {
     puts "***** EXPORT IP *****"
     set time_start [clock clicks -milliseconds]
-    export_design -format ip_catalog
+    export_design -format ip_catalog -version $version
     set time_end [clock clicks -milliseconds]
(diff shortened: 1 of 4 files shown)


vloncar:
Nice!

But how would one set a version using the public API? The only way to use it is to create the config manually and call the `keras_to_hls()` or set it after the `ModelGraph` was created but before the `write()` is called.

AdrianAlan:
Now it should be possible to specify it through `hls4ml.converters.create_config` as well. Is there another way you thought about it?

PS Wouldn't you have the same concern about `ClockUncertainty`?

vloncar:
Looks good. It failed for the accelerator backend, so I added support for it as well.

PR #855, opened 2023-08-20
Automatic precision inference

vloncar:
# Description

This introduces the ability to specify `auto` as a precision string, that implies hls4ml should infer the precision in some way. This is not exposed by default via the `config_from...` functions for now. The goal is to have the framework for inferring types in some ways within hls4ml (e.g., QONNX parser) before fully exposing it to users. An initial inference of precision has been added via the `infer_precision_types` optimizer, based on previous attempts by various people. It's not advanced in any way. During testing, I encountered some issues with `SeparableConv1D` templates which I fixed.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) - Only related to the `SeparableConv1D` issue

## Tests

There are new tests in `test_auto_precision.py` that cover the few use cases.

## Checklist


jmitrevs:
Clang doesn't seem to like this
```
(fastml39) Jovans-Mac:hls4mlprj_auto_conv2d_Quartus_io_stream jmitrevs$ bash build_lib.sh 
In file included from firmware/myproject.cpp:2:
In file included from firmware/parameters.h:11:
firmware/nnet_utils/nnet_conv2d_stream.h:135:5: error: no matching function for call to 'shift_line_buffer_2d'
    nnet::shift_line_buffer_2d<data_T, CONFIG_T>(in_elem, line_buffer, shift_buffer);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
firmware/nnet_utils/nnet_conv2d_stream.h:199:13: note: in instantiation of function template specialization 'nnet::compute_output_buffer_2d<nnet::array<ac_fixed<16, 6>, 4>, nnet::array<ac_fixed<35, 15>, 4>, config8>' requested here
            compute_output_buffer_2d<data_T, res_T, CONFIG_T>(padds, res, line_buffer, kernel_window, weights, biases);
            ^
firmware/myproject.cpp:93:11: note: in instantiation of function template specialization 'nnet::conv_2d_cl<nnet::array<ac_fixed<16, 6>, 4>, nnet::array<ac_fixed<35, 15>, 4>, config8>' requested here
    nnet::conv_2d_cl<layer7_t, last_layer_result_t, config8>(layer7_out, layer8_out, w8, b8);
          ^
firmware/nnet_utils/nnet_conv2d_stream.h:69:6: note: candidate template ignored: substitution failure [with data_T = nnet::array<ac_fixed<16, 6>, 4>, CONFIG_T = config8]: zero-length arrays are not permitted in C++
void shift_line_buffer_2d(
     ^
1 error generated.
clang: error: no such file or directory: 'myproject.o'
```
while g++ compiles it just fine.

jmitrevs:
The issue seems to be that this:
```python
    model.add(Conv2D(4, kernel_size=(1, 1), activation='relu', name='last_layer'))  # Will become PointwiseConv2D
```
doesn't actually become PointwiseConv2D for Quartus. The bug it uncovered seems tangential to this PR; nevertheless, we need to fix it, either as part of this PR or separately.

vloncar:
is this the same issue that was observed in #878?

jmitrevs:
> is this the same issue that was observed in #878?

I believe so. If you have a filter of size 1, then things like `line_buffer[CONFIG_T::filt_height - 1][CONFIG_T::n_chan]` wind up with zero-size arrays.

hls4ml/backends/quartus/quartus_backend.py
```diff
@@ -72,6 +72,7 @@ def _register_flows(self):
             'quartus:inplace_parallel_reshape',
             'quartus:inplace_stream_flatten',
             'quartus:skip_softmax',
+            'infer_precision_types',
```
jmitrevs, line 75: Shouldn't this have quartus: prepended?

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -108,6 +108,7 @@ def _register_flows(self):
             'vivado:inplace_parallel_reshape',
             'vivado:inplace_stream_flatten',
             'vivado:skip_softmax',
+            'infer_precision_types',
```
jmitrevs, line 111: As in the quartus case, shouldn't this have vivado: prepended?

vloncar: No, as a general optimizer that doesn't depend on any backend, it needs no prefix. It is used here since backend-specific optimizers can introduce new layers with types that are "auto", so it needs to run again. Similar to how you have to repeat specific_types optimizer in different flows

test/pytest/test_auto_precision.py
```diff
@@ -0,0 +1,255 @@
...
+                'weight': 'auto',
+            },
+        },
+        'last_layer': {
+            'Precision': {
+                'result': 'auto',
+            },
+        },
+    }
+    odir = str(test_root_path / f'hls4mlprj_auto_{model_type}_{backend}_{io_type}')
```
jmitrevs, line 153: Might want to add something like this to avoid Winograd, which is lower precision:
```python
    if backend == 'Quartus' and io_type == 'io_parallel':
        # Winegrad imp[lementation does not support binary
        hls_config["LayerName"]["conv2d_1"]["Implementation"] = "im2col"
        hls_config["LayerName"]["conv2d_2"]["Implementation"] = "im2col"
        hls_config["LayerName"]["conv2d_3"]["Implementation"] = "im2col"
```

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+        inferred_types = self._infer_precision(node, types_to_infer)
+        for type_name in types_to_infer:
+            if type_name not in inferred_types:
+                self._infer_default_type(node, type_name)
+
+        return False  # No model graph changes made
+
+    def _infer_precision(self, node, types_to_infer):
+        node_class = node.class_name
+        if node_class in ['Dense']:
```
jmitrevs, line 31: I winder if it's better to use something like `isinstance(node, Dense)` instead of matching to a class name. Matching to a class name doesn't deal with inheritance. For example, I can see the BatchNormalization matching failing for ApplyAlpha matching.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+            return self._infer_dense_precision(node, types_to_infer)
+
+        if node_class in ['BatchNormalization']:
+            return self._infer_bn_precision(node, types_to_infer)
+
+        if node_class in ['Conv1D', 'Conv2D', 'PointwiseConv1D', 'PointwiseConv2D', 'Conv2DBatchnorm']:
```
jmitrevs, line 37: Why are all the ifs independent, not `elif`? Do we expect to match multiple ones?

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: LOL, I did it specifically to avoid this since it is usually not what we want. The ApplyAlpha is an example of this.

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: As a coding style, `if`s with a return statement don't really need `elif`, it's just clutter

hls4ml/model/optimizer/passes/infer_precision.py
jmitrevs: With QONNX ApplyAlpha does need the precision propagated. It worries me if derived classes by default have different behavior. I think if you want different behavior you explicitly should code it. What happens in ApplyAlpha if you don't forbid it? Should QONNX not use ApplyAlpha in this case?

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+        if 'result_t' in types_to_infer:
+            node.types['result_t'].name = node.name + '_result_t'
+            node.types['result_t'].precision = new_type
+
+            inferred_types.append('result_t')
+
+        return inferred_types
+
+    def _infer_dense_precision(self, node, types_to_infer):
+        n_ops = node.get_attr('n_in') * node.get_attr('n_out')
```
jmitrevs, line 136: I don't think the total n_ops is the important value for the accumulator precision. Ignoring bias for now, each output value is a certain number of multiplies, resulting in the `input_width + weight_width` part of the equation, plus the accumulation, `math.ceil(np.log2(num_acc))`. But num_acc != num_ops. In particular, it's `node.get_attr('n_in')` (-1?), at least for the standard 1D Dense layer. Bias modifies things a bit, but the general trends are the same. This is the result we had from the CMS hackathon with Sioni:  https://github.com/jmitrevs/hls4ml/blob/bit-correct/hls4ml/model/optimizer/passes/propagate_dense_precision.py

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+                bias_width = node.types['bias_t'].precision.width
+                bias_integers = node.types['bias_t'].precision.integer
+            node.weights['bias'].update_precision(node.types['bias_t'].precision)
+
+            inferred_types.append('bias_t')
+        else:
+            bias_width = node.types['bias_t'].precision.width
+            bias_integers = node.types['bias_t'].precision.integer
+
+        new_type = FixedPrecisionType(
```
jmitrevs, line 116: Should we worry about unsigned inputs (or less likely, weights or biases)? The old version explicitly handled signedness.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+        assert 'result_t' in types_to_infer and len(types_to_infer) == 1
+
+        in_var = node.get_input_variable()
+        out_var = node.get_output_variable()
+        in_out_type = in_var.type.precision
+        out_var.type.precision = in_out_type
+
+        return ['result_t']
+
+    def _infer_common_precision(self, node, types_to_infer, n_ops):
```
jmitrevs, line 73: I think this assumes integer or fixed precision types. Do we need to handle, e.g., xnor precision types?

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: Ah, true, I'll fix this.

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: If I say we shouldn't, someone will come tomorrow with unsigned inputs, so perhaps we should consider doing it. I'll update based on your code.

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: I should test this PR with the binary model that has Xnor types. I wouldn't expect that you need to infer anything in that case, but perhaps it breaks the optimizer.

hls4ml/model/optimizer/passes/infer_precision.py
jmitrevs: Just for concreteness, here is the precision inference we used when we tried to match QKeras at the CMS Trigger workshop:
```python
def _propagate_type_dense(input_precision, weight_precision, bias_precision, num_acc):
    '''
    Propagate the precion type across a multiply. Rounding modes are propagated from input_precision
    '''

    # check to make sure none are None
    bitwidth = weight_precision.width + input_precision.width + math.ceil(np.log2(num_acc))
    integer = weight_precision.integer + input_precision.integer + math.ceil(np.log2(num_acc))
    signed = weight_precision.signed or input_precision.signed

    # Because calculating precision, no need to round or satration
    rounding_mode = None
    saturation_mode = None

    frac = bitwidth - integer

    # correct for bias
    if bias_precision:
        integer = max(integer + (bias_precision.signed and not signed),
                      bias_precision.integer + (signed and not bias_precision.signed)) + 1
        bitwidth = integer + max(frac, bias_precision.width - bias_precision.integer)
        signed = signed or bias_precision.signed

    return FixedPrecisionType(bitwidth, integer, signed, rounding_mode, saturation_mode)
```

jmitrevs:
I made [vloncar #53](https://github.com/vloncar/hls4ml/pull/53) into your branch with changes for dense and standard convolution. Let me know what you think. If you like the way I made this, I can also add signed/unsigned support to the other precision propagations (like merge, bn, sepconv) either in that PR or a different one.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+            node.types['result_t'].precision = new_type
+
+            inferred_types.append('result_t')
+
+        return inferred_types
+
+    def _infer_bn_precision(self, node, types_to_infer):
+        inferred_types = []
+
+        if 'scale_t' in types_to_infer:
```
jmitrevs, line 222: Wouldn't the input quantization in this case be the input mean and variance (+ other things), not the scale and bias?

jmitrevs: (actually I have to see how this is handled by the qkeras parser)

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -0,0 +1,290 @@
...
+            node.weights['bias'].update_precision(node.types['bias_t'].precision)
+            inferred_types.append('bias_t')
+
+        if 'result_t' in types_to_infer:
+            scale_precision = node.types['scale_t'].precision
+            bias_precision = node.types['bias_t'].precision
+
+            out_precision = deepcopy(node.get_input_node().get_output_variable().type.precision)
+            out_precision.integer += scale_precision.integer
+            out_precision.fractional = max(out_precision.fractional, scale_precision.fractional)
```
jmitrevs, line 238: Explicitly keeping integer, fractional, and consistent worries me. I would propose making fractional a property of the `FixedPrecisionType` like:
```python
    @property
    def fractional(self):
        self.width - self.integer
```

Also, another concern with this logic is if the input node type is an `IntegerPrecionsType`. Setting fractional and integer explicitly could result in a misformed type. I would propose explicitly creating a new `FixedPrecisionType` vs editing the existing one.

hls4ml/model/optimizer/__init__.py
```diff
@@ -33,6 +33,7 @@
 register_flow(
     'convert',
     [
+        'infer_precision_types',
```
jmitrevs, line 36: I would prefer having this towards the end of convert, or just leaving the call out of convert, since we call it in the optimize flow anyway. What is the purpose of doing it here? Here it messes up some of the conversion steps I have for qonnx, which only happen if the type is not set, since it doesn't want to override set types, and this inferring here sets the types.

vloncar: "optimize" is sort-of optional, it belongs to "convert" because after that stage the other optimizers don't expect "auto" to exist. If onnx has its own optimizers that run, why not group them into a flow and run before convert? the idea is that after "convert" it shouldn't matter where the model came from

jmitrevs: But we can move it later in concert, right?

vloncar: If you have onnx-specific optimizers that must go before, place them before this one.

jmitrevs (approved):

PR #811, opened 2023-06-14
Explicit pointwise Conv1D implementation for "Latency" strategy

jmduarte:
# Description

This is mostly for discussion and to let others test it out like @Duchstf. This PR adds an explicit pointwise Conv1D implementation, where the reuse factor (`RF`) is used to split the layer execution and reuse the existing module `RF` times

Original pointwise Conv1D: 
- `(in_width, n_chan) -> (in_width, n_filt)`

This PR splits it into `RF` calls of 
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- ...

The II ~ RF. To turn it on you have to configure `ConvImplementation` of the layer named `<layer>`
```python
config["LayerName"]["<layer>"]["ConvImplementation"] = "Pointwise"
```

Limitations: 
- Assumes `in_width` is divisible by `RF`
- Hardcoded explicit execution up to `RF = 120`. Could be automated with code generation.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] A new research paper code implementation

## Tests

See [test/pytest/test_pointwiseconv.py](https://github.com/jmduarte/hls4ml/blob/split_pointwise_conv_by_rf_rebase_latest/test/pytest/test_pointwiseconv.py)

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [x] I have added tests that prove my fix is effective or that my feature works.

PR #881, opened 2023-10-08
Pointwise Conv1D with code generation for "Latency" strategy (update of #811)

jmduarte:
# Description

Update of #811  with code generation. This PR adds an explicit pointwise Conv1D implementation, where the reuse factor (`RF`) is used to split the layer execution and reuse the existing module `RF` times

Original pointwise Conv1D: 
- `(in_width, n_chan) -> (in_width, n_filt)`

This PR splits it into `RF` calls of 
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- `(in_width/RF, n_chan) -> (in_width/RF, n_filt)`
- ...

The II ~ RF. It is on by default, but I think you should be able to use the standard conv1d implementation by skipping the optimizer.

Limitations: 
- Assumes `in_width` is divisible by `RF`

## Type of change

## Tests

See [test/pytest/test_pointwiseconv.py](https://github.com/jmduarte/hls4ml/blob/split_pointwise_conv_by_rf_codegen/test/pytest/test_pointwiseconv.py)

## Checklist


hls4ml/backends/fpga/fpga_backend.py
```diff
@@ -860,6 +860,64 @@ def generate_conv2d_line_buffer_fn(
 
         return generated_code
 
+    def generate_pointwise_conv1d_fn(self, layer_idx, reuse_factor=1):
```
vloncar, line 863: If we're placing this here, it's available for all backends, including quartus, but the rest of the changes doesn't seem to touch quartus?

hls4ml/backends/fpga/passes/codegen.py
```diff
@@ -49,3 +49,25 @@ def _generate_im2col_2d(self, node):
         )
 
         node.set_attr('line_buffer_codegen', Source(code_str))
+
+
+class GeneratePointwiseConv1D(OptimizerPass):
```
vloncar, line 54: hmm, I'm thinking if we should split this into a new file? and establish a naming convention. for example, current codegen.py could be im2col_codegen.py and this one pointwise_codegen.py. this will get cramped in the future (e.g., Chang will also use this for HGQ)

hls4ml/backends/vivado/passes/convolution_templates.py
```diff
@@ -56,6 +56,8 @@
     typedef {config_t} mult_config;
     template<unsigned K, unsigned S, unsigned W>
     using scale_index = nnet::{scale_index_type}<K, S, W>;
+    template<class data_T, class res_T, class CONFIG_T>
```
vloncar, line 59: Doesn't this mean that every Conv1D layer will have a pointer to a pointwise function that it may not even use?

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -71,8 +71,9 @@ def _register_layer_attributes(self):
 
         for layer in cnn_layers:
             attrs = self.attribute_map.get(layer, [])
-            # attrs.append(ConfigurableAttribute('conv_implementation', value_type=str, default='LineBuffer'))
-            attrs.append(ChoiceAttribute('conv_implementation', choices=['LineBuffer', 'Encoded'], default='LineBuffer'))
+            attrs.append(
+                ChoiceAttribute('conv_implementation', choices=['LineBuffer', 'Encoded', 'Pointwise'], default='LineBuffer')
```
vloncar, line 75: So this basically means that I can put `Pointwise` as the implementation for layers that are not pointwise? that's not user-friendly because if it causes the pointwise hls functions to be used they will fail an assert, which is invisible when calling from python.

hls4ml/templates/vivado/build_prj.tcl
```diff
@@ -161,7 +161,7 @@ if {$opt(reset)} {
 } else {
     open_solution "solution1"
 }
-catch {config_array_partition -maximum_size 4096}
+catch {config_array_partition -maximum_size 8192}
```
vloncar, line 164: I think this change should be discussed separately from this PR. We had 4k limit forever, increasing it allows the synthesis to run much longer, take more memory but it is not clear if this ultimately works and meets timing. I wonder why Xilinx set it to 1024 by default and then removed it in Vitis. Either way, I'd be more in favor of this being configurable as "you're on your own if you change it"

hls4ml/templates/vivado/nnet_utils/nnet_code_gen.h
```diff
@@ -25,6 +25,96 @@ template <class data_T, typename CONFIG_T> class FillConv2DBuffer {
...
+    static void pointwise_conv(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
+                               res_T res[CONFIG_T::out_width * CONFIG_T::n_filt],
+                               typename CONFIG_T::weight_t weights[CONFIG_T::n_chan * CONFIG_T::n_filt],
+                               typename CONFIG_T::bias_t biases[CONFIG_T::n_filt]) {
+        // To be implemented in subclasses
+    }
+};
+
+template <class data_T, class res_T, typename CONFIG_T>
+void pointwise_conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan / CONFIG_T::reuse_factor],
```
vloncar, line 39: Why is this here?

hls4ml/templates/vivado/nnet_utils/nnet_common.h
```diff
@@ -24,6 +24,7 @@ namespace nnet {
 // Common type definitions
 enum io_type { io_parallel = 0, io_stream };
 enum strategy { latency, resource };
+enum class conv_implementation { linebuffer = 0, encoded = 1, pointwise = 2 };
```
vloncar, line 27: Same question as before

hls4ml/templates/vivado/nnet_utils/nnet_conv1d.h
```diff
@@ -53,9 +53,19 @@ void pointwise_conv_1d_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
 
     #pragma HLS INLINE region
 
-    // Nothing special to be done for io_parallel implementation
     if (CONFIG_T::strategy == nnet::latency) {
-        conv_1d_latency_cl<data_T, res_T, CONFIG_T>(data, res, weights, biases);
+        if (CONFIG_T::implementation == conv_implementation::pointwise) {
+            // Use pointwise unrolled implementation
+            if (CONFIG_T::reuse_factor > 1) {
+                CONFIG_T::template pointwise_conv<data_T, res_T, CONFIG_T>::pointwise_conv(data, res, weights, biases);
```
vloncar, line 60: So a model with a kernel > 1 but set conv_implementation to pointwise, the code will be generated and since it doesn't have assert it will proceed to do an incorrect computation? It's nonsense to us, but I fear that misunderstanding of the docs or just bugs in conversion scripts may trigger this

test/pytest/test_pointwiseconv.py
```diff
@@ -100,6 +105,7 @@ def test_pointwiseconv2d(chans, padds, strides, backend, io_type, strategy):
             kernel_initializer='normal',
             use_bias=False,
             data_format=chans,
+            name='pointwise2d',
```
vloncar, line 108: general comment, not specific to this line: why don't we do this for 2d as well?

hls4ml/backends/fpga/fpga_backend.py
jmduarte: Moved to `vivado/vivado_backend.py`

hls4ml/backends/fpga/passes/codegen.py
jmduarte: split the new code into a new file `vivado/passes/pointwise_codegen.py` and renamed this one to `fpga/passes/im2col_codegen.py`

hls4ml/templates/vivado/build_prj.tcl
jmduarte: created a new configurable parameter `MaximumSize` that is by default 4096

hls4ml/templates/vivado/nnet_utils/nnet_code_gen.h
jmduarte: I was worried about some circular imports, but actually it seems ok... I moved it back to `nnet_cov1d_latency.h`

hls4ml/templates/vivado/build_prj.tcl
vloncar: I know I'm nitpicking here, but can we rename this to `ArrayPartitionLimit` or something more descriptive? MaximumSize doesn't provide information as to what it refers to

hls4ml/backends/vivado/vivado_backend.py
jmduarte: I agree this is bad, but I'm not sure the best way to fix it. Is it as simple as checking if a user tries to do this and exiting gracefully with a warning message?

hls4ml/templates/vivado/nnet_utils/nnet_conv1d.h
jmduarte: So there is an assert here: 
https://github.com/fastmachinelearning/hls4ml/blob/1dd2603558f8ceb6d16b449c67e52567650d3eaf/hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h#L88-L92

Where else should we add it?

hls4ml/backends/vivado/passes/convolution_templates.py
jmduarte: OK, I moved this only in the pointwise optimizer pass. I think this addresses the issue, but let me know if it can be done better.

test/pytest/test_pointwiseconv.py
jmduarte: Yes, we can easily add 2d... working on it.

hls4ml/backends/vivado/vivado_backend.py
vloncar: Perhaps I'm forgetting other cases here, but wasn't this shown to have better performance than existing implementation? Therefore one could argue we don't really need a switch to choose algorithms we know are worse. I'd make this default and not hide it behind a (currently undocumented) switch.

jmduarte:
Comparison of resources/latency: https://docs.google.com/spreadsheets/d/1fme-MpCr-WFArLka5Q07yC5gVUe7l4kh7MZdYzSmB5k/edit?usp=sharing

jmduarte:
pre-commit.ci autofix

hls4ml/backends/vivado/vivado_backend.py
jmduarte: made default with latest commits

hls4ml/templates/vivado/nnet_utils/nnet_common.h
jmduarte: pointwise used by default by latest commit. i think you can still override it by skipping to the optimizer (to be checked).

hls4ml/templates/vivado/nnet_utils/nnet_conv1d.h
jmduarte: resolved i think with the latest commits.

jmduarte:
pre-commit.ci autofix

jmitrevs:
I am slightly confused by the performance table above. Currently if io_parallel we don't have specific pointwise implementations and just the standard latency or resource convolutions are used. How much does this improve over that?

jmitrevs:
Also, so I understand the previous discussion, this is no longer selected as a separate implementation but is just io_parallel latency with size 1 kernel, right?

jmduarte:
> I am slightly confused by the performance table above. Currently if io_parallel we don't have specific pointwise implementations and just the standard latency or resource convolutions are used. How much does this improve over that?

The row labeled `Pointwise` is the latency, io_parallel implementation in this PR. The row labeled `LineBuffer` is the current latency, io_parallel  implementation for the same layer. Do you want any other comparisons? I can add them.

 > Also, so I understand the previous discussion, this is no longer selected as a separate implementation but is just io_parallel latency with size 1 kernel, right?

Yes, it is just activated by the pointwise optimizer that runs. So I think we can re-enable the previous behavior by skipping the optimizer.

hls4ml/backends/vivado/passes/convolution_templates.py
```diff
@@ -93,6 +93,11 @@ def format(self, node):
         else:
             params['fill_fn'] = 'FillConv1DBuffer'
 
+        if node.get_attr('filt_width') == 1 and node.model.config.get_config_value('IOType') == 'io_parallel':
```
jmitrevs, line 96: Is this needed? I assumed it was handled by the optimizer.

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -649,3 +649,61 @@ def init_garnet(self, layer):
     @layer_optimizer(GarNetStack)
     def init_garnet_stack(self, layer):
         self.init_garnet(layer)
+
+    def generate_pointwise_conv1d_fn(self, layer_idx, reuse_factor=1):
```
jmitrevs, line 653: Does it make sense to put this instead in `pointwise_codegetn.py`?

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -84,5 +84,84 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
...
+                #pragma HLS UNROLL
+                int index_mult = ii * CONFIG_T::n_filt * CONFIG_T::n_chan + ff * CONFIG_T::n_chan + cc;
+                int index_weight = cc * CONFIG_T::n_filt + ff;
+                int index_data = (ii * CONFIG_T::stride_width - CONFIG_T::pad_left) * CONFIG_T::n_chan + cc;
+
+                if ((ii * CONFIG_T::stride_width) < CONFIG_T::pad_left ||
+                    (ii * CONFIG_T::stride_width) >= (CONFIG_T::pad_left + CONFIG_T::in_width)) {
+                    mult[index_mult] = 0;
+                } else {
+                    mult[index_mult] = data[index_data] * weights[index_weight];
```
jmitrevs, line 129: Should this use `product` from `nnet_mult.h` so that it does the right thing when using binary weights?

jmitrevs (changes requested):

hls4ml/backends/vivado/passes/convolution_templates.py
jmduarte: In principle, maybe we don't have to check this explicitly again because this function will never be used unless it's a pointwise conv (and optimized as such). Do you suggest to remove it? I tried to follow the example about for other generated code like the fill_buffer, that does check for `io_parallel` at least.

hls4ml/backends/vivado/vivado_backend.py
jmduarte: Yes, done.

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
jmduarte: Yes, done. Also, in the vitis version.

jmduarte:
pre-commit.ci autofix

hls4ml/backends/vivado/passes/convolution_templates.py
jmitrevs: My suggestion was to remove it so if the pointwise optimizer was turned off, it would still work using the regular conv. Also, is the io_stream right?

jmduarte: I don't think this particular piece of code prevents us from using standard conv if the optimizer can be skipped. This just adds a parameter to the config that may or may not be used.

However, I did try skipping the optimizer with this snippet: 
https://gist.github.com/jmduarte/351871db7fdfbf530ec02aecaa95e756#file-test_deepset-py-L48
```python
config['SkipOptimizers'] = ['vivado:optimize_pointwise_conv', 'vitis:optimize_pointwise_conv']
```
and I find that the optimizer still runs even though it seems to be successfully removed from the optimizer list (!). If I manually deactivate it (by having the matching criteria always return False), then it is skipped. Do we know why this optimizer is always running?

vloncar: SkipOptimizer feature doesn't work since the introduction of flows. We'll need a different way of exposing the feature to the user eventually. Right now the only way is to set the flows through the config and make sure the optimizer isn't there. It's a bit hacky.

hls4ml/backends/vivado/passes/pointwise.py
```diff
@@ -4,13 +4,45 @@
...
+    template<class data_T, class CONFIG_T>
+    using fill_buffer = nnet::{fill_fn}<data_T, CONFIG_T>;
+    typedef {accum_t.name} accum_t;
+    typedef {bias_t.name} bias_t;
+    typedef {weight_t.name} weight_t;
+    typedef {config_t} mult_config;
+    template<unsigned K, unsigned S, unsigned W>
+    using scale_index = nnet::{scale_index_type}<K, S, W>;
+    template<class data_T, class res_T, class CONFIG_T>
+    using pointwise_conv = nnet::{pointwise_fn}<data_T, res_T, CONFIG_T>;
```
vloncar, line 42: In `Dense` layers we moved to using a function pointer like this that the main `dense()` function calls, eliminating the need for checks in HLS. Here it would also simplify the call hierarchy (no need for special handling of pointwise) and no need for this template.

hls4ml/backends/vivado/passes/pointwise_codegen.py
```diff
@@ -0,0 +1,84 @@
...
+    Args:
+        layer_idx (int): Index of layer ('index' attribute).
+        reuse_factor (int): Number of partitions to divide the input into.
+
+    Returns:
+        str: Generated C++ function
+    """
+
+    generated_code = (
+        "template<class data_T, class res_T, typename CONFIG_T>\n"
```
vloncar, line 18: Minor point, but we use single quotes, double quotes only in special cases

hls4ml/backends/vivado/passes/pointwise_codegen.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    return generated_code
+
+
+class GeneratePointwiseConv1D(OptimizerPass):
+    '''Generates code for pointwise 1D convolution'''
+
+    def match(self, node):
+        return isinstance(node, Conv1D) and node.model.config.get_config_value('IOType') == 'io_parallel'
```
vloncar, line 69: Why is there no check for `filt_width == 1` here? Otherwise we generate functions we don't use and is incorrect at that.

hls4ml/backends/vivado/passes/pointwise_codegen.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def transform(self, model, node):
+        node_class = node.__class__.__name__
```
vloncar, line 72: Minor point, but we have `node.class_name` for this purpose, but wouldn't make a big difference in the check here.

The bigger question is why the check at all? In what cases can it fail?

hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -85,5 +85,85 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
...
+    // Use a function_instantiate in case it helps to explicitly optimize unchanging weights/biases
+    #pragma HLS function_instantiate variable=weights,biases
+
+    // Parallel mode
+    #pragma HLS PIPELINE II=CONFIG_T::reuse_factor
+    #pragma HLS ARRAY_PARTITION variable=weights complete dim=0
+    #pragma HLS ARRAY_PARTITION variable=biases complete dim=0
+
+    // Limit multipliers to control parallelization
+    constexpr unsigned multiplier_limit = DIV_ROUNDUP(
```
vloncar, line 110: This was problematic before, and we moved to setting `multiplier_limit` in python and using it here. For consistency, we should not re-introduce old approaches.

hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -85,5 +85,85 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
...
+                acc[ii][ff] += mult[index_mult];
+            } // end channel loop
+        }     // end filter loop
+    }         // end output loop
+
+    // Cast to "res_t" type
+    for (int ii = 0; ii < CONFIG_T::out_width / CONFIG_T::reuse_factor; ii++) {
+        for (int ff = 0; ff < CONFIG_T::n_filt; ff++) {
+            #pragma HLS UNROLL
+            res[ii * CONFIG_T::n_filt + ff] = (res_T)(acc[ii][ff]);
```
vloncar, line 163: Do you need to call `cast()` here to ensure compatibility with all variants of `product`?

hls4ml/templates/vivado/nnet_utils/nnet_common.h
```diff
@@ -2,6 +2,7 @@
 #define NNET_COMMON_H_
 
 #include "ap_fixed.h"
+#include "nnet_helpers.h"
```
vloncar, line 5: `nnet_helpers.h` (mostly) doesn't contain synthesizable code and it is not intended to be included by other files apart from the testbench. And I don't see you using anything from it in the code that is introduced

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -84,5 +84,85 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
     }
 }
 
+template <class data_T, class res_T, typename CONFIG_T>
+void pointwise_conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan / CONFIG_T::reuse_factor],
```
vloncar, line 88: The files in Vitis/Vivado are mostly identical (apart from the very recent change in the final loop), perhaps we could remove the Vitis one (first we follow up with the changes)

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -84,5 +84,85 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
...
+    // Use a function_instantiate in case it helps to explicitly optimize unchanging weights/biases
+    #pragma HLS function_instantiate variable=weights,biases
+
+    // Parallel mode
+    #pragma HLS PIPELINE II=CONFIG_T::reuse_factor
+    #pragma HLS ARRAY_PARTITION variable=weights complete dim=0
+    #pragma HLS ARRAY_PARTITION variable=biases complete dim=0
+
+    // Limit multipliers to control parallelization
+    constexpr unsigned multiplier_limit = DIV_ROUNDUP(
```
vloncar, line 109: Same comment as in the other file

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
```diff
@@ -84,5 +84,85 @@ void conv_1d_latency_cl(data_T data[CONFIG_T::in_width * CONFIG_T::n_chan],
...
+                acc[ii][ff] += mult[index_mult];
+            } // end channel loop
+        }     // end filter loop
+    }         // end output loop
+
+    // Cast to "res_t" type
+    for (int ii = 0; ii < CONFIG_T::out_width / CONFIG_T::reuse_factor; ii++) {
+        for (int ff = 0; ff < CONFIG_T::n_filt; ff++) {
+            #pragma HLS UNROLL
+            res[ii * CONFIG_T::n_filt + ff] = (res_T)(acc[ii][ff]);
```
vloncar, line 162: Same comment as in the other file

vloncar (changes requested):
Looks very good, there are some minor tweaks to integrate cleaner. Works very well, which is more important :-)

hls4ml/templates/vivado/nnet_utils/nnet_common.h
jmduarte: If I remove that, I get this error on my EL 8 machine with g++ 8.5.0
```
In file included from firmware/nnet_utils/nnet_conv1d_latency.h:4,
                 from firmware/nnet_utils/nnet_code_gen.h:4,
                 from firmware/parameters.h:7,
                 from firmware/myproject.cpp:4:
firmware/nnet_utils/nnet_common.h: In function ‘T nnet::reduce(const T*, Op)’:
firmware/nnet_utils/nnet_common.h:37:39: error: there are no arguments to ‘floorlog2’ that depend on a template parameter, so a declaration of ‘floorlog2’ must be available [-fpermissive]
     static constexpr int leftN = pow2(floorlog2(N - 1)) > 0 ? pow2(floorlog2(N - 1)) : 0;
                                       ^~~~~~~~~
firmware/nnet_utils/nnet_common.h:37:39: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
firmware/nnet_utils/nnet_common.h:37:68: error: there are no arguments to ‘floorlog2’ that depend on a template parameter, so a declaration of ‘floorlog2’ must be available [-fpermissive]
     static constexpr int leftN = pow2(floorlog2(N - 1)) > 0 ? pow2(floorlog2(N - 1)) : 0;
```

hls4ml/backends/vivado/passes/pointwise_codegen.py
jmduarte: done in https://github.com/jmduarte/hls4ml/pull/22/commits/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03

hls4ml/backends/vivado/passes/pointwise.py
jmduarte: attempted in https://github.com/jmduarte/hls4ml/pull/22/commits/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03

hls4ml/backends/vivado/passes/pointwise_codegen.py
jmduarte: added in https://github.com/jmduarte/hls4ml/pull/22/commits/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03

hls4ml/backends/vivado/passes/pointwise_codegen.py
jmduarte: removed check in https://github.com/jmduarte/hls4ml/pull/22/commits/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03

hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h
jmduarte: updated to use the `multiplier_limit` from the config here: https://github.com/jmduarte/hls4ml/pull/22/commits/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03, but need to check the value is the same

hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h
jmduarte: added here:

https://github.com/fastmachinelearning/hls4ml/blob/d56dc7349e94ad8b6aaefe5b0f22385d3d9a9e03/hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h#L161

hls4ml/templates/vivado/nnet_utils/nnet_conv1d_latency.h
jmduarte: Do you mean both `nnet_conv1d.h` and `nnet_conv1d_latency.h`?

There are a few differences like the use of inline region, etc. Do you mean to test if just using the vivado versions works in vitis now?

hls4ml/templates/vitis/nnet_utils/nnet_conv1d_latency.h
jmduarte: ok, fixed the multiplier limit now... but it would be good to warn the user somehow if they use a `reuse_factor` that doesn't divide `in_width`

related: it would probably be beneficial to factorize the two into a `reuse_factor` (that limits the multipliers) and a `parallelization_factor` (that controls how many times the conv is split into separate calls)

JanFSchulte:
pre-commit.ci autofix

Issue #883, opened 2023-10-10
SeparableConv1d fail to synthesize


qberthet:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x] If there are steps to reproduce the problem, make sure to write them down below.
- [ ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
Project using `QSeparableConv1D` layers fail to synthesize.

## Details
While playing with the new SeparableConv1D feature, it appear that all attempts to pass the CSynth step fail with the same error.

### Steps to Reproduce

1. Clone the hls4ml repository
2. Checkout the master branch, with commit hash d36e2267c5330247453324aac3008c0f4807c891
3. Run the following script (Using Vivado 2020.1 setup):
```python
from keras.layers import Input
from keras.models import Model
from qkeras import *
import hls4ml

def get_model():
    # Define a dummy model with only one QSeparableConv1D layer
    input_layer = Input(shape=(32, 3))
    layer = QSeparableConv1D(
                filters=16,
                kernel_size=3,
                depthwise_quantizer=quantized_bits(16, 6, alpha=1),
                pointwise_quantizer=quantized_bits(16, 6, alpha=1),
                bias_quantizer=quantized_bits(16, 6, alpha=1)
            )(input_layer)
    model = Model(inputs=input_layer, outputs=layer)
    return model

model = get_model()

model.summary()

config = hls4ml.utils.config_from_keras_model(model, granularity="name")

# Configure the project to be SeparableConv1D compatible
config['Model']['Precision'] = 'ap_fixed<16,6>'
config['Model']['ReuseFactor'] = 1
config['Model']['Strategy'] = 'Latency'

# Use the Vivado backend (2020.1)
cfg = hls4ml.converters.create_config(backend='Vivado')
cfg['IOType'] = 'io_stream'
cfg['HLSConfig'] = config
cfg['KerasModel'] = model
cfg['OutputDir'] = 'hls4ml_prj'
cfg['Part'] = 'xcku115-flvb2104-2-i'

hls_model = hls4ml.converters.keras_to_hls(cfg)

hls_model.compile()

hls_model.build(reset=True, csim=False, synth=True)
```
### Expected behavior
Model should synthesize correctly.

### Actual behavior
[...]

PR #884, opened 2023-10-10
Fix config structure name in pragma for SeparableConv1D

qberthet:
# Description

Simple fix for #883

## Type of change

## Tests

Tested with example code provided in bug report #883 : Fail before applying this commit, works afterward.

## Checklist

### DIFF
--- a/hls4ml/backends/vivado/passes/convolution_templates.py
+++ b/hls4ml/backends/vivado/passes/convolution_templates.py
@@ -41,6 +41,8 @@ conv1d_config_template = """struct config{index} : nnet::conv1d_config {{
     static const unsigned out_width = {out_width};
     static const unsigned reuse_factor = {reuse};
     static const unsigned n_zeros = {nzeros};
+    static const unsigned multiplier_limit =
+        DIV_ROUNDUP(kernel_size * n_chan * n_filt, reuse_factor) - n_zeros / reuse_factor;
     static const bool store_weights_in_bram = false;
     static const unsigned strategy = nnet::{strategy};
     static const nnet::conv_implementation implementation = nnet::conv_implementation::{implementation};
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -213,7 +213,9 @@ class VivadoBackend(FPGABackend):
             f'vsynth={vsynth} '
             f'fifo_opt={fifo_opt}"'
         )
-        os.system(vivado_cmd)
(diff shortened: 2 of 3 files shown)


jmitrevs:
Is it possible to add a pytest to show that there is a problem and that this fixes it? We don't do synthesis in pytests but we do compile the HLS code.

qberthet:
I'll try to add this later today.

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
```diff
@@ -23,7 +23,7 @@ void depthwise_product(data_T data[CONFIG_T::kernel_size * CONFIG_T::n_chan], re
 
     #pragma HLS ARRAY_PARTITION variable=mult complete
 
-    #pragma HLS ALLOCATION operation instances=mul limit=CONFIG_T::multiplier_limit
+    #pragma HLS ALLOCATION operation instances=mul limit=CONFIG_T::mult_config::multiplier_limit
```
vloncar, line 26: This technically works, but we deliberately didn't use the `mult_config` in `depthwise_product`. In 2D the config has the `multiplier_limit` and we should use the same approach. So below [this ](https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/vivado/passes/convolution_templates.py#L43) line we should add to [these](https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/vivado/passes/convolution_templates.py#L146-L147) lines.

vloncar (changes requested):

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
qberthet: Ok, I was suspecting that there was something else to it, but not familiar with the code base yet. I'll start by adding a pytest that trig the issue, add the fix you suggest and force push to this PR.

qberthet:
@jmitrevs : I've added a test, but the problem only appear during CSynth, not CSim, so not sure if that will be ok (requires vivado). Also had to catch the return code of vivado_hls command to raise exception and make the test fail. Let me know if you prefer that I remove the two first commit to only have the fix.

@vloncar : your fix obviously does the job. Thanks!

qberthet:
And I just realize that my test is not setting the output folder correctly. Let me know if you want to keep these tests before I fix it.

vloncar (approved):

Issue #878, opened 2023-09-28
HLS4ML IN QUARTUS SUPPORT FOR CONV2D 


MODISENEISO:
Quick Summary 
Attempting to convert a trained Keras Model with Quartus as Backend. but get this error
![image](https://github.com/fastmachinelearning/hls4ml/assets/47188943/65e68f6f-7a60-4e17-b793-a7875e7ec8f2)

Detail
hls4ml v 0.7.1
1)The 3 layer hls4ml tutorial is able to convert to RTL with Quartus Backend.
2)The model I want to convert has Conv2D layers and Upsampling layers.
3)Vivado HLS is able to convert the same model without any errors.
4)When converting the same model with Quatus 20.1 backend the following error comes during the build p project -s step.
5)hl:~$ hls4ml build -p bratsquartus2 -s
Loading configuration from bratsquartus2/hls4ml_config.yml
Loading configuration from bratsquartus2/hls4ml_config.yml
i++   -march=Cyclone10GX --time quartus-hlssynt.log brats_test.cpp firmware/brats.cpp   -o brats-fpga
In file included from firmware/brats.cpp:2:
In file included from firmware/parameters.h:11:
firmware/nnet_utils/nnet_conv2d_stream.h:135:5: error: no matching function for call to 'shift_line_buffer_2d'
    nnet::shift_line_buffer_2d<data_T, CONFIG_T>(in_elem, line_buffer, shift_buffer);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
firmware/nnet_utils/nnet_conv2d_stream.h:199:13: note: in instantiation of function template specialization 'nnet::compute_output_buffer_2d<nnet::array<ac_fixed<16, 6, true, AC_TRN, AC_WRAP>, 8>, nnet::array<ac_fixed<16, 6, true, AC_TRN, AC_WRAP>, 4>, config11>' requested here
            compute_output_buffer_2d<data_T, res_T, CONFIG_T>(padds, res, line_buffer, kernel_window, weights, biases);
            ^
firmware/brats.cpp:103:11: note: in instantiation of function template specialization 'nnet::conv_2d_cl<nnet::array<ac_fixed<16, 6, true, AC_TRN, AC_WRAP>, 8>, nnet::array<ac_fixed<16, 6, true, AC_TRN, AC_WRAP>, 4>, config11>' requested here
[...]

PR #886, opened 2023-10-11
Add support for filt_height==1 for streaming quartus conv2d

jmitrevs:
# Description

The streaming Conv2D implementation in the quartus backend doesn't handle the `filt_height==1` case, as discovered in #878 and #855. We currently do not have a special pointwise streaming implementation for quartus, but nevertheless the default implementation should be able to handle the case, even if a bit less efficiently. This PR adds some C++ fixes for issues that clang and i++ complained about (but not g++).

## Type of change

## Tests

I just enabled streaming with the Quartus backend to the `test_pointwiseconv.py` test. As mentioned, there is no special pointwise implementation in this case, but the results should nevertheless be correct

## Checklist


jmitrevs:
It seems like the pytests ran out of space again. Trying to restart.

vloncar (approved):

MODISENEISO:
Hello if I run update hls4ml in CLI will the new merging be updated

Issue #885, opened 2023-10-11
Softmax with less than 10bits input generates error with default config


calad0i:
When the input bitwidth to a softmax layer is smaller than 10 bits, a warning message
 `WARNING: Higher bound(something) and lower(something-10) bound cannot be negative.`
is thrown more than 1000 times, and random results are generated from the softmax layer.

This issue presents with both Stable and Latency strategies.

The corresponding code where the warning was raised is https://github.com/fastmachinelearning/hls4ml/blob/d36e2267c5330247453324aac3008c0f4807c891/hls4ml/templates/vivado/nnet_utils/nnet_activation.h#L144
When the input's precision to the new softmax is smaller than 10 bits (table_size=1024 by default and not configurable in an obvious way), the cut to generate indices goes sub-zero, and the significant bits go random and produce random outputs. Adjusting `table_size` manually will fix the issue.

PR to fix comes later.

PR #887, opened 2023-10-13
Fix bit overflow with softmax

calad0i:
A# Description

> :memo: Please include a summary of the change.
> 
> * Please also include relevant motivation and context.
> * List any dependencies that are required for this change.

Fix for #885 : softmax working with less than 10 bits of input/table size. Tested with Vivado/Vitis/quartus.

## Type of change

## Tests

>

**Test Configuration**:

`pytest/test_softmax.py`

## Checklist

- [ ] My changes generate no new warnings. - Raise a warning when the added fix is triggered.

### DIFF
--- a/test/pytest/test_softmax.py
+++ b/test/pytest/test_softmax.py
@@ -8,15 +8,16 @@ from sklearn.metrics import accuracy_score
 import hls4ml
 
 test_root_path = Path(__file__).parent
-
-
-def normal_dist(shape):
-    return np.clip(np.random.normal(0, 8, shape), -32, 31)
+test_root_path = Path('/tmp/unit_test')
 
 
 @pytest.fixture()
 def generate_data(input_shape):
-    return normal_dist((1000, *input_shape))
+    shape = (5000, *input_shape)
+    d = np.random.normal(0, 2, shape)
+    modify_entries = np.random.randint(0, 1, shape) < 0.05
+    d[modify_entries] = d[modify_entries]*5+10
+    return np.clip(d, -32, 31)
 
 
 @pytest.mark.parametrize('backend', ['Vivado', 'Vitis', 'Quartus'])
@@ -46,7 +47,7 @@ def test_softmax(backend, strategy, generate_data, input_bits, input_shape, tabl
     cfg['LayerName']['softmax']['Strategy'] = strategy
     cfg['LayerName']['softmax']['inv_table_t'] = f_type
     cfg['LayerName']['softmax']['exp_table_t'] = f_type
-    cfg['LayerName']['softmax_input']['Precision']['result'] = f'ap_fixed<{input_bits}>'
+    cfg['LayerName']['softmax_input']['Precision']['result'] = f'fixed<{input_bits}>'
 
     odir = str(
         test_root_path
@@ -70,9 +71,9 @@ def test_softmax(backend, strategy, generate_data, input_bits, input_shape, tabl
 @pytest.mark.parametrize('io_type', ['io_parallel', 'io_stream'])
 def test_softmax_skipped(backend, io_type):
     X = np.random.rand(100, 10)
-    model = tf.keras.models.Sequential()
-    model.add(tf.keras.layers.Dense(14, input_shape=(10,), name='dense'))
-    model.add(tf.keras.layers.Activation(activation='softmax', name='softmax'))
+    dense = tf.keras.layers.Dense(14, input_shape=(10,), name='dense')
+    softmax = tf.keras.layers.Activation(activation='softmax', name='softmax')
+    model = tf.keras.models.Sequential([dense,softmax])
     model.compile()
 
     cfg = hls4ml.utils.config_from_keras_model(model, granularity='name')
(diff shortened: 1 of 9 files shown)


jmitrevs:
I wonder if the optimizer should really be in the backends directory. I think optimizers in `model/optimizers` should not concern themselves with backends.

jmitrevs (changes requested):
I wonder if the optimizer should really be in the backends directory. I think optimizers in `model/optimizers` should not concern themselves with backends.

calad0i:
> I wonder if the optimizer should really be in the backends directory. I think optimizers in `model/optimizers` should not concern themselves with backends.

Moved

test/pytest/test_softmax.py
```diff
@@ -10,50 +10,47 @@
...
-
-
-def high_accuracy_distribution(shape):
-    '''Start with a flat distribution, then pick a random member of each row to amplify'''
-    x = np.random.rand(*shape)
-    imax = np.random.randint(0, shape[1], size=shape[0])
-    x[:, imax] *= 10
-    return x
+def normal_dist(shape):
+    return np.clip(np.random.normal(0, 8, shape), -32, 31)
```
vloncar, line 14: Why are we getting rid of existing distributions?

test/pytest/test_softmax.py
```diff
@@ -10,50 +10,47 @@
...
-def test_softmax(backend, strategy, generate_data, input_shape, io_type, function):
+def test_softmax(backend, strategy, generate_data, input_bits, input_shape, table_bits, io_type):
     X = generate_data
     model = tf.keras.models.Sequential()
     model.add(tf.keras.layers.Activation(input_shape=input_shape, activation='softmax', name='softmax'))
     model.compile()
 
-    f_type = 'ac_fixed<18,8,true,AC_RND,AC_SAT>' if backend == 'Quartus' else 'ap_fixed<18,8,AP_RND,AP_SAT>'
+    f_type = (
+        f'ac_fixed<{table_bits},true,AC_RND,AC_SAT>' if backend == 'Quartus' else f'ap_fixed<{table_bits},AP_RND,AP_SAT>'
```
vloncar, line 43: I think a new standard for tests that use multiple backends is to just do `fixed<W,I>`

test/pytest/test_softmax.py
```diff
@@ -10,50 +10,47 @@
...
+    )
     cfg = hls4ml.utils.config_from_keras_model(model, granularity='name')
     cfg['LayerName']['softmax']['Strategy'] = strategy
     cfg['LayerName']['softmax']['inv_table_t'] = f_type
     cfg['LayerName']['softmax']['exp_table_t'] = f_type
+    cfg['LayerName']['softmax_input']['Precision']['result'] = f'ap_fixed<{input_bits}>'
```
vloncar, line 49: Related to the comment above, here you're actually relying on what is essentially a bug to correctly parse even in case of Quartus.

test/pytest/test_softmax.py
```diff
@@ -92,7 +89,6 @@ def test_softmax_skipped(backend, io_type):
     assert len(hls_layers) == 2
 
     # Verify hls4ml output is equal to Dense output
-    y_keras = model.predict(X)
-    y_hls4ml = hls_model.predict(X).reshape(y_keras.shape)
-    keras_trace = hls4ml.model.profiling.get_ymodel_keras(model, X)
-    np.testing.assert_allclose(y_hls4ml, keras_trace['dense'], rtol=0, atol=2e-2)
+    y_keras = model.layers[0](X).numpy()  # type: ignore
```
vloncar, line 92: Why do you need to do it like this?

test/pytest/test_softmax.py
calad0i: The existing distribution generation does not feel necessary, and I don't see a good reason to run tests twice on uniform and uniform+manual outliers. A gaussian will also give variables of many scales, giving some outliers by default.

test/pytest/test_softmax.py
calad0i: These are inherited from the old test file. Will `SAT` / `RND` flags be also auto-converted? Will change.

test/pytest/test_softmax.py
calad0i: The original code was redundant (and confusing to me): what is done here is essentially skipping the last layer in Keras. As there are only two, applying the first layer only is enough.
Will revert if you think the original one is better.

vloncar: I think we want just `model.predict(X)`. The section looks like it was cut from a different test. We should use the opportunity to clean up (also the surrounding comments)

test/pytest/test_softmax.py
calad0i: In this case I shall purge this part, should an equivalent test already exists somewhere.

test/pytest/test_softmax.py
calad0i: In this case I shall purge this part, should an equivalent test already exists somewhere.

jmitrevs (approved):

PR #898, opened 2023-10-25
Remove #INLINE pragma to fix SeparableConv1D with Vitis backed

### LINKED ISSUE #897: SeparableConv1D fail CSynth with Vitis backend
## Prerequisites

## Quick summary
Model using SeparableConv1D fail CSynth when using Vitis backend.

### Steps to Reproduce
2. Checkout the main branch, with commit hash: d03573888227da9b7527ce7a2d25403640ed4637
3. Run the following script:
```python
from keras.layers import Input
from keras.models import Model
from qkeras import *
import hls4ml

def get_model():
    # Define a dummy model with only one QSeparableConv1D layer
    input_layer = Input(shape=(32, 3))
    layer = QSeparableConv1D(
                filters=16,
                kernel_size=3,
                depthwise_quantizer=quantized_bits(16, 6, alpha=1),
                pointwise_quantizer=quantized_bits(16, 6, alpha=1),
                bias_quantizer=quantized_bits(16, 6, alpha=1)
            )(input_layer)
    model = Model(inputs=input_layer, outputs=layer)
    return model

model = get_model()

model.summary()

config = hls4ml.utils.config_from_keras_model(model, granularity="name")

# Configure the project to be SeparableConv1D compatible
config['Model']['Precision'] = 'ap_fixed<16,6>'
config['Model']['ReuseFactor'] = 1
config['Model']['Strategy'] = 'Latency'

# Use the Vivado backend (2020.1)
cfg = hls4ml.converters.create_config(backend='Vitis')
cfg['IOType'] = 'io_stream'
cfg['HLSConfig'] = config
cfg['KerasModel'] = model
cfg['OutputDir'] = 'hls4ml_prj'
cfg['Part'] = 'xcku115-flvb2104-2-i'

hls_model.compile()

hls_model.build(reset=True, csim=False, synth=True)
```
### Expected behavior
Successful CSynt.

### Actual behavior
CSytnh fail with the following error:
`ERROR: [HLS 214-272] In function 'void nnet::depthwise_product<ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>, ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>, config2_depthwise>(ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>*, ap_fixed<16, 6, (ap_q_mode)5, (ap_o_mode)3, 0>*, config2_depthwise::weight_t*, config2_depthwise::bias_t*)', Pragma conflict happens on 'INLINE' and 'PIPELINE'FUNCTION_INSTANTIATE' pragmas: same function (firmware/nnet_utils/nnet_sepconv_stream.h:13:0)`

qberthet:
Fixes #897

# Description

With Vitis backend, when `#pragma HLS function_instantiate` is used, `#pragma HLS INLINE` can only be used with the `off` option to forbid inlining. Using only `#pragma HLS INLINE` results in the error mentionned in #897.

From what I understand from https://docs.xilinx.com/r/2022.2-English/ug1399-vitis-hls/Function-Instantiation, removing this pragma does not means that inlining will not be performed when possible.

## Type of change

## Tests

As this is tested during CSynth, no PyTest are provided, but the test script in #897 can be used to test that the problem is indeed fixed.

## Checklist


vloncar:
This won't get inlined automatically by the compiler, meaning +1 on latency of the inner product for every output pixel due to the extra function call. It's the `function_instantiate` pragma that we should remove. That pragma is a leftover from the wild '10s of Vivado HLS where tricks like that were actual optimizations (note that there will only ever be one instance of this function since the weights array is fixed, so it is useless)

qberthet:
If `function_instantiate` is removed, there is then a clash with `#pragma HLS PIPELINE ...`. If this one is removed, the issue is solved, but again, not sure about the impact on resulting rtl.

vloncar:
Looking at the surrounding code there's already a `inline recursive` pragma before each call to this function. Does that get ignored? Is inlining itself a problem (that messes with the pipeline of the `depthwise_conv_1d_buffer_cl` that ultimately calls this) or is the issue the existence of additional pipeline directive in the `depthwise_product` (which should be ignored with a warning, but is promoted to an error in Vitis)? I think we should test if this works better inlined or not, and if not, prevent inlining by removing those pragmas (and perhaps even keeping the one in `depthwise_product` but with `off`). Just not with `function_instantiate` :-)

qberthet:
To be honest, I have no clue about all this. I just have the impression from the error message that the multiple pragma are simple mutually exclusive with Vitis (obviously no issue with Vivado).

I am willing to dig a bit in all this and do some tests, but not sure when I will have the time to do it..

JanFSchulte:
The synthesis issue fixed by this PR is not present in newer versions of Vitis and 2022.1 is not supported by hls4ml for a variety of reasons. Closing this PR as no longer necessary.

PR #908, opened 2023-10-31
Quartus multi out with stream fix

calad0i:
A# Description

Quartus writer writes uncompilable programs when there are multiple outputs with io_stream. This PR fixes that.

## Type of change

## Tests

`test/pytest/test_multiout_network.py`

**Test Configuration**:

## Checklist

### DIFF
--- a/test/pytest/test_multiout_network.py
+++ b/test/pytest/test_multiout_network.py
@@ -1,0 +1,53 @@
+import os
+import random
+from pathlib import Path
+
+import numpy as np
+import pytest
+import tensorflow as tf
+from keras.layers import Dense
+from tensorflow import keras
+
+from hls4ml.converters import convert_from_keras_model
+
+test_root_path = Path(__file__).parent
+
+
+@pytest.fixture(scope='module')
+def model():
+    seed = 42
+    os.environ['RANDOM_SEED'] = f'{seed}'
+    np.random.seed(seed)
+    tf.random.set_seed(seed)
+    tf.get_logger().setLevel('ERROR')
+    random.seed(seed)
+
+    inp = keras.Input(shape=(10,))
+    x = Dense(10)(inp)
+    y = Dense(10)(inp)
+    model = keras.Model(inp, [x, y])
+    return model
+
+
+@pytest.fixture(scope='module')
+def data():
+    rng = np.random.RandomState(42)
+    X = rng.normal(0, 1, (1000, 10))
+    X = np.clip(X, -16, 15)
+    return X
+
+
+@pytest.mark.parametrize('backend', ['Vivado', 'Quartus', 'Vitis'])
+@pytest.mark.parametrize('io_type', ['io_parallel', 'io_stream'])
+def test_multi_clone(model, data, backend: str, io_type: str):
+    output_dir = str(test_root_path / f'hls4mlprj_multiout_network_{backend}_{io_type}')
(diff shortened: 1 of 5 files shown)


jmitrevs:
Does this work correctly if the multiple outputs have a different type? I am worried that they get forced to `result_t`, though I didn't check.

jmitrevs:
It is these lines that make me concerned about the output types when there are multiple outputs:
https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/graph.py#L472-L477
https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/graph.py#L610-L613

vloncar:
@jmitrevs Those lines were also problematic in our attempts to split the writing into multiple IPs. There's no practical reason for this feature to exist, it is merely to be human readable from the very ancient times, and in my (limited) tests everything still works if they are removed

jmitrevs:
I just tried:
```python
def test_multi_clone(model, data, backend: str, io_type: str):
    output_dir = str(test_root_path / f'hls4mlprj_multiout_network_{backend}_{io_type}')
    hls_config = config_from_keras_model(model, "name", default_precision='fixed<32,5>')
    hls_config["LayerName"]["dense_linear"]["Precision"]["result"] = 'fixed<35,5>'
    hls_config["LayerName"]["dense_1_linear"]["Precision"]["result"] = 'fixed<40,5>'
    hls_config["LayerName"]["dense"]["Precision"]["result"] = 'fixed<35,5>'
    hls_config["LayerName"]["dense_1"]["Precision"]["result"] = 'fixed<40,5>'
    model_hls = convert_from_keras_model(
        model, backend=backend, output_dir=output_dir, hls_config=hls_config, io_type=io_type
    )
    model_hls.compile()
    r_hls = model_hls.predict(data)
    r_keras = [x.numpy() for x in model(data)]

    assert np.allclose(r_hls[0], r_keras[0], atol=1e-5, rtol=0)
    assert np.allclose(r_hls[1], r_keras[1], atol=1e-5, rtol=0)
```
and both outputs are `result_t` of `fixed<35,5>`. I could have made a mistake in the code (and I am not sure why I had to define the linear result like that, too), but the behavior doesn't seem quire right.

calad0i:
I changed the two places in `graph.py`, and now the output variables are not overwriting each other. It does not seem to affect other tests in my tests. Though, maybe this should go into a separate PR.

jmitrevs:
There is still some manual seeding left, if I looked at the code correctly.

jmitrevs:
@calad0i, can you remove the remaining seed setting?

jmitrevs (approved):

test/pytest/test_multiout_network.py
```diff
@@ -0,0 +1,53 @@
...
+
+import numpy as np
+import pytest
+from keras.layers import Dense
+from tensorflow import keras
+
+from hls4ml.converters import convert_from_keras_model
+
+test_root_path = Path(__file__).parent
+test_root_path = Path('/tmp/hls4mlprj_multiout_network')
```
jmitrevs, line 11: We don't usually put the results in /tmp/

jmitrevs:
Generally other that the test_root_path issue, I think this is good. After that is fixed we can merge it.

jmitrevs (approved):

PR #906, opened 2023-10-31
Fix for #905

calad0i:
Fix for #905 

- [x] Bug fix (non-breaking change that fixes an issue)

## Tests

`test/pytest/test_stream_multi_clone.py`

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [x] I have added tests that prove my fix is effective or that my feature works.

PR #911, opened 2023-11-05
Add precision inherition for when generating stream clone

calad0i:
A# Description

Based on #906
When creating cloned streams, use the precision of the stream being cloned. Currently, the precision is `model_default` without any special tweak.

## Type of change

## Tests

`test/pytest/test_stream_clone.py`

Merged with tests from #906.

## Checklist


test/pytest/test_stream_clone.py
```diff
@@ -0,0 +1,111 @@
...
+from tensorflow import keras
+
+from hls4ml.converters import convert_from_keras_model
+
+test_root_path = Path(__file__).parent
+
+
+@pytest.fixture(scope='module')
+def model_clone_precision_inherition():
+    seed = 42
```
jmitrevs, line 18: We should use pytest-randomly for setting the seed.

hls4ml/backends/fpga/passes/clone.py
```diff
@@ -79,13 +77,17 @@ def transform(self, model, node):
...
+                clone_layer: Clone = model.make_node(
                     Clone,
                     'clone_' + node.name,
                     attrs,
                     [output],
                     [output + '_cpy' + str(i + 1) for i in range(len(output_map[output]))],
                 )
+                for i in range(len(output_map[output])):
+                    key = output + '_cpy' + str(i + 1)
+                    clone_layer.attributes[key].type = node.attributes['result_t']
```
jmitrevs, line 90: This is copying the object reference, not the object, right? Is that fine? One could argue that they should really be the same, but I wanted to check.

jmitrevs: (I think this is probably good since it would be more confusing to have multiple copies of a named type, so I think I will resolve it soon unless someone objects in the meantime.)

hls4ml/__init__.py
```diff
@@ -22,3 +22,8 @@ def reseed(newseed):
         torch.manual_seed(newseed)
     except ImportError:
         print('\nPyTorch seed not set')
+        seed = 42
```
jmitrevs, line 25: I think this is leftover code from testing

hls4ml/__init__.py
```diff
@@ -22,3 +24,5 @@ def reseed(newseed):
         torch.manual_seed(newseed)
     except ImportError:
         print('\nPyTorch seed not set')
+
+    os.environ['RANDOM_SEED'] = f'{newseed}'
```
jmitrevs, line 28: Why do we need to set the environment variable?

jmitrevs (approved):

Issue #912, opened 2023-11-05
Inconsistant behaviour of `accum_t` in `io_stream` and `io_parallel` for `pooling` layers with `Vivado` backend


calad0i:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [ ] If there are steps to reproduce the problem, make sure to write them down below.
- [ ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
For pooling operations with `vivado` (or `vitis`) backend and  `io_parallel`, `accum_t` is defined but not used. With `io_stream` though, the intermediate values when reducing the pool are stored in an array of type `accum_t`.

Worth mentioning here that `quartus` not define `accum_t` at all, but uses `value_type` for intermediate values.

### Expected behavior
Behaviour of `accum_t` should be consistent with different io types, (and backends, ideally). 

### Actual behavior
As stated.

### Possible fix
- purge `accum_t` all at once, as the necessary precision can be derived from input precision: directly input precision for max pool, and `(u)fixed<b+2*ceil(log2(N)), i+1*ceil(log2(N))>`, with `b,i` being width and int width of input precision, for the average pool.

- or, let all implementations use `accum_t` to store intermediate values when reducing the pool. However, the user would need to set the precision manually in this case.

PR #914, opened 2023-11-07
Add support for HGQ proxy model

calad0i:
# Requires #973 and `python>=3.10` now.
# Description

This PR adds support for a special layer type `FixedPointQuantizer` used in [calad0i/HGQ/hls4ml-integration](https://github.com/calad0i/HGQ/tree/hls4ml-integration) to facilitate conversion from HGQ model to hls4ml model graph. ModelGraph converted from proxy-model is meant to be bit-accurate, until `fp32` (or `tf32`) cannot emulate the required precision during tensorflow model inference. 

The layer has three major functions:
1. When heterogeneous activation quantization is used, do value masking (with code generation). If not, the layer is removed.
2. Overrides precision settings for other layers. All precision settings are embedded into these layers in a proxy model.
3. Though not used in hls4ml, when `fp32` can represent the required precision, the proxy model can be used to emulate hls models' output, **with or without** overflows with `SAT` or `WRAP` (other overflow modes not tested).

In the future, the proxy model mechanism can also be used with Qkeras models, but this part is not yet implemented.

#912 could still break bit-accuracy for pooling layer with `io_stream`. Temporary fix available at #917 (included), superseded by #855 when available.

**New** (09/01/24):

- Rebased to current master and squashed commits for basic HGQ proxy (also supports QKeras partially) support.

- Allowing converting any unary activation function to a LUT, ensuring bit-accuracy for all unary activation functions. Performance in resource and timing may be better or worse depending on the input data format.

**TODO**

If the python version for hls4ml bumps to `>=3.10`, we can make HGQ an optional dependency for hls4ml and re-use tests from there. We can remove the re-defining of `FixedPointQuantizer` in hls4ml also.

Independent of this PR, `Conv2D` in Quartus with `3x3` filter size with `io_parallel` seems to be broken at the moment, some tests are **expected to fail** at the moment. The cause seems to be related to the winograd implementation.

~~This PR dependents on #887 #906 #907 #908 #909 #911~~ All merged now.

## Type of change

## Tests

`test/pytest/test_proxy_model.py`

**Test Configuration**:

Requires models add in fastmachinelearning/example-models#11.

## Checklist


jmitrevs:
I had issues like this, of the tests running to the limit, with the qonnx branch. I am rerunning now, but it may be better to merge fewer tests. I had to change it for the qonnx branch.

jmitrevs:
In the oneAPI branch, we have `  N_TESTS_PER_YAML: 4` on line 10 of .gitlab-ci.yml. Maybe try that here?

hls4ml/backends/fpga/passes/hgq_proxy_model.py
```diff
@@ -0,0 +1,107 @@
...
+    name: str, shape: tuple[int, ...], k: np.ndarray, b: np.ndarray, i: np.ndarray, RND: str, SAT: str, backend: str
+) -> str:
+    """Generate heterogenous quantization mask function, ONLY works for IOType=io_parallel"""
+    assert k.shape[0] == b.shape[0] == i.shape[0] == 1
+    assert backend.lower() in ('quartus', 'vivado', 'vitis'), f'Backend {backend} not tested'
+    Ks, Bs, Is = k[0], b[0], i[0]
+    Ks, Bs, Is = np.broadcast_to(Ks, shape), np.broadcast_to(Bs, shape), np.broadcast_to(Is, shape)
+    Ks, Bs, Is = Ks.ravel(), Bs.ravel(), Is.ravel()
+    masks = []
+    to_fixed = to_acfixed if backend.lower() == 'quartus' else to_apfixed
```
jmitrevs, line 31: This doesn't work for catapult (already in main branch) or future oneAPI backends. I would recommend trying to use the standard type converters instead of doing your own.

hls4ml/backends/fpga/passes/hgq_proxy_model.py
```diff
@@ -0,0 +1,107 @@
...
+
+from hls4ml.backends import Backend
+from hls4ml.backends.template import FunctionCallTemplate
+from hls4ml.model.layers import Layer
+from hls4ml.model.optimizer import OptimizerPass
+from hls4ml.model.optimizer.passes.hgq_proxy_model import FixedPointQuantizer, UnaryLUT
+from hls4ml.model.types import Source
+
+
+def to_apfixed(k, b, i, RND, SAT):
```
jmitrevs, line 11: These should not be necessary. The standard type conversions should handle this automatically via https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/fpga/fpga_types.py#L28 and similar.

calad0i: Is there a unified type string converter (backend agnostic) I can grab from, or do I still need to do it for every different backends?

hls4ml/model/optimizer/passes/hgq_proxy_model.py
```diff
@@ -0,0 +1,151 @@
...
+        self.add_output_variable(shape, dims)
+        self.set_attr('n_in', inp.size())
+        self.table = self.attributes['table']
+        self.table_size = self.attributes['table_size']
+
+        table_t = to_hls4ml_fixed(self.attributes['table_t'])
+        self.add_weights_variable(name='table', var_name='table{index}', precision=table_t, data=self.table)
+
+
+def to_hls4ml_fixed(fixed: str):
```
jmitrevs, line 40: Aren't there standard already-used functions that implement this? I am worried about ac_fixed vs ac_fixed differences here.

hls4ml/model/optimizer/passes/hgq_proxy_model.py
```diff
@@ -0,0 +1,151 @@
...
+    matched = re_parse_fixed.match(re_purge_prefix.sub('', fixed))
+    assert matched is not None, f'Cannot parse {fixed}'
+    signed = matched.group(1) != 'u'
+    b, i, *args = matched.group(2).split(',')
+    b, i = int(b), int(i)
+    args = [arg.upper() for arg in args]
+    new_type = FixedPrecisionType(b, i, signed, *args)
+    # For some reason, __class__ is overwritten in hls4ml
```
jmitrevs, line 48: I wonder if that's the conversion to a backend version, so that it knows how to print it properly. (I am just guessing now.)

calad0i: Unfortunately, this need to go through another round of meta programming (done in `TransformTypes` optimizer) to got printed well, thus unlikely can be used at the spot of code generation (final string produced at the spot).

hls4ml/backends/fpga/passes/hgq_proxy_model.py
jmitrevs: I believe the way it works by default is via the TransformTypes (e.g. https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/vivado/passes/transform_types.py). That determines the text-based implementation for internal types. See if there's a way to use the same machinery somehow.

hls4ml/model/optimizer/passes/hgq_proxy_model.py
calad0i: Ok, glad to see that the backend specific types are no longer generated on the fly. Though, the current serialization strategy seems to be a bit problematic: Say one defines `SAT` without rounding strategy, it will generate `ap_(u)fixed<w,i,AP_SAT>`, which doesn't do what it intends to do.... can be fixed quite easily though.

To do it properly, probably, we will the backends declaring their required type string generator (converter, whatever). As I went to through the backends, I did not spot them at the moment, and it seems that everything is still handled through an optimizer pass. This will go to a separate PR as it would require changing working flows irrelevant to this.

calad0i:
Rebased to current head to cleanup history

jmitrevs (approved):

PR #924, opened 2023-11-15
Update default FPGA part number from KU115 to VU13P

jmduarte:
# Description

- Update default FPGA part number from KU115 to VU13P, as more commonly used by hls4ml users (e.g., for CMS Phase-2 Level-1 Trigger)

## Type of change

- [x] Documentation update
- [x] Other: Default configuration change

## Tests

- Checking in CI

## Checklist

### DIFF
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -178,7 +178,7 @@ class VivadoBackend(FPGABackend):
     def create_initial_config(self, part='xcvu13p-flga2577-2-e', clock_period=5, io_type='io_parallel'):
         config = {}
 
-        config['Part'] = part
+        config['Part'] = part if part is not None else 'xcku115-flvb2104-2-i'
         config['ClockPeriod'] = clock_period
         config['IOType'] = io_type
         config['HLSConfig'] = {}
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -175,10 +175,10 @@ class VivadoBackend(FPGABackend):
     def get_writer_flow(self):
         return self._writer_flow
 
-    def create_initial_config(self, part='xcku115-flvb2104-2-i', clock_period=5, io_type='io_parallel'):
+    def create_initial_config(self, part='xcvu13p-flga2577-2-e', clock_period=5, io_type='io_parallel'):
         config = {}
 
-        config['Part'] = part if part is not None else 'xcku115-flvb2104-2-i'
+        config['Part'] = part if part is not None else 'xcvu13p-flga2577-2-e'
         config['ClockPeriod'] = clock_period
         config['IOType'] = io_type
         config['HLSConfig'] = {}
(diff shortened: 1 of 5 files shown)


hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -175,10 +175,10 @@ def get_default_flow(self):
     def get_writer_flow(self):
         return self._writer_flow
 
-    def create_initial_config(self, part='xcku115-flvb2104-2-i', clock_period=5, io_type='io_parallel'):
+    def create_initial_config(self, part='xcvu13p-flga2577-2-e', clock_period=5, io_type='io_parallel'):
         config = {}
 
-        config['Part'] = part if part is not None else 'xcku115-flvb2104-2-i'
+        config['Part'] = part if part is not None else 'xcvu13p-flga2577-2-e'
```
jmitrevs, line 181: I guess this existed before, but why does part have a default value in the function parameters and then there's the check if it's None? Is that if someone explicitly assigns None to part? I wonder if just `config['Part'] = part` is better here.

jmduarte: Yeah, I think that's the only case it would be None. So if we want to remove this catch, we can change it for Intel as well: https://github.com/fastmachinelearning/hls4ml/tree/2b46ebf7bb3e5847871870da6e0cd190e00bde93/hls4ml/backends/vitis#L130

vloncar: Not sure, but is this related to the accelerator backend that sets board and not part? The same check exists in [there](https://github.com/fastmachinelearning/hls4ml/blob/2b46ebf7bb3e5847871870da6e0cd190e00bde93/hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py#L132) as well, maybe we get rid of this style altogether if it doesn't have specific purpose

jmduarte: Seeing what happens without this catch. 

I also realized why we had something like this. I think it comes from the `VivadoAccelerator` backend.

In that case, `part=None` by default
https://github.com/fastmachinelearning/hls4ml/blob/2b46ebf7bb3e5847871870da6e0cd190e00bde93/hls4ml/backends/vivado_accelerator/vivado_accelerator_backend.py#L98-L109

and it's overwritten to be consistent with the `board`
https://github.com/fastmachinelearning/hls4ml/blob/2b46ebf7bb3e5847871870da6e0cd190e00bde93/hls4ml/backends/vivado_accelerator/vivado_accelerator_config.py#L9-L26

jmitrevs: So do we need to revert the change, or is this handled correctly?

jmitrevs: I don't really see how our change would make something fail if it worked before, but I am not positive. What's our VivadoAccelerator pytest coverage?

jmduarte: will fix in a new issue/PR

jmitrevs (approved):

Issue #932, opened 2023-12-02
Keras Reshape Layer is Built with Error


MrFaith2001:
## Quick summary
I want to synthesize a model with reshape and concatenate layers through hls4ml, but when I build the generated HLS file, there will be a problem of undefined variables, do I need to manually set some additional configuration?
## Details
My model consists of two inputs and an output, both of which pass through the corresponding reshape layer, and finally merge through the concatenate layer.

### Steps to Reproduce
1. Run conversion on model file with code
My model file is here [https://drive.google.com/file/d/1q0JdEZhn2AYalodmhS665kSUdEVZccOC/view?usp=drive_link](model_file)

```
  import os
  
  os.environ["PATH"] += os.pathsep + "/data2t1/dataset/Xilinx_Public/Vivado/2019.2/bin"
  
  from tensorflow.keras.models import load_model
  import hls4ml
  from print_dict import print_dict
  from tensorflow import keras
  
  import tensorflow as tf
  from tensorflow.keras.models import Model
  
  model_name = "Ret_re5re6_in_box_con4_out"
  model = load_model('./model/{}.h5'.format(model_name))
  model.summary()
  model.compile()
  
  import plotting
  import matplotlib.pyplot as plt
  from sklearn.metrics import accuracy_score
  import hls4ml
  from pathlib import Path
  
  hls_config = hls4ml.utils.config_from_keras_model(model, granularity='name')
  hls_config['Model']['Precision'] = 'ap_fixed<32,8>'
  hls_config['Model']['Strategy'] = 'Latency'
  for Layer in hls_config['LayerName'].keys():
      hls_config['LayerName'][Layer]['Precision'] = 'ap_fixed<32,8>'
      hls_config['LayerName'][Layer]['Strategy'] = 'Latency'
  
  print("-----------------------------------")
  print("Configuration")
  plotting.print_dict(hls_config)
  print("-----------------------------------")
  
  hls_model = hls4ml.converters.convert_from_keras_model(model, 
                                                         io_type='io_stream', 
                                                         backend='Vivado',
                                                         output_dir='{}'.format(model_name), 
                                                         project_name='{}'.format(model_name),
                                                         part='xcvu13p-fhga2104-1-e', 
                                                         hls_config=hls_config
                                                      )
  
  hls_model.compile()
  hls_model.build(csim=False, synth=True)
```

### Expected behavior
I want to synthesize and get a two-input and one-output project directly.
### Actual behavior
Error info:
```
[...]

PR #934, opened 2023-12-03
better repalce_node fn

calad0i:
# Description

Make `ModelGraph.replace_node` method work with nodes work with replacing nodes, which itself/parents/children have multi-input/outputs.

#932 will be fixed with this PR.

This fix is rather long but keeps the previous behavior in most cases. A much shorter version will be simply doing

```python
        new_node.inputs = old_node.inputs
        new_node.outputs = old_node.outputs
```

but the final variable names will be different from the current version.

## Type of change

## Tests

Integrated with `test_repack_precision`, renamed to `test_repack_stream`

## Checklist


test/pytest/test_repack_stream.py
```diff
@@ -0,0 +1,62 @@
+from pathlib import Path
+
+import numpy as np
+import pytest
+from tensorflow import keras
+
+from hls4ml.converters import convert_from_keras_model
+
+# test_root_path = Path(__file__).parent
+test_root_path = Path('/tmp')
```
jmitrevs, line 10: This needs to be reverted.

jmitrevs:
So the order of the nodes would stay the same, right? Is there a possibility to reorder the nodes? (I am not saying it is needed; just trying to understand.)

jmitrevs:
I will tentatively put this as 0.8.1 milestone, but I would be fine deferring to 1.0.

calad0i:
> So the order of the nodes would stay the same, right? Is there a possibility to reorder the nodes? (I am not saying it is needed; just trying to understand.)
> 

They should stay the same for the current implementation (copy in order w/ replacing a single k/v pair).

What do you mean by reorder in this case? Shuffling the nodes will break the graph in general I guess.

jmitrevs:
I was thinking if there was a case where the new node expected the order of inputs to be different than the old node. It's more hypothetical, without a specific case in mind. ONNX nodes often have multiple inputs with specific roles assigned to specific inputs, but we usually transform the special inputs early.

jmitrevs:
Do the tests ever produce a case when the length or the inputs or outputs is not one? I tried looking with a debugger and didn't see it in the tests, but I could have missed it.

test/pytest/test_repack_stream.py
```diff
@@ -0,0 +1,61 @@
...
+    out = keras.layers.Concatenate(name='concat')([r1, r2])
+    model = keras.Model([inp1, inp2], out)
+
+    hls_config = {'Model': {'Precision': 'ap_ufixed<8,8>', 'ReuseFactor': 1}, 'Strategy': strategy}
+    model_hls = convert_from_keras_model(
+        model,
+        io_type='io_stream',
+        backend='Vivado',
+        hls_config=hls_config,
+        output_dir=str(test_root_path / f'{backend}_{strategy}'),
```
jmitrevs, line 52: Generally the names have been of the style, hl4mlprj_repack_{backend}_{strategy}. The way it is currently, if you run pytest locally it's hard to refer back to the results.

test/pytest/test_repack_stream.py
```diff
@@ -0,0 +1,61 @@
...
+    layer_conf = {
+        'inp': {'Precision': 'fixed<20,10>'},
+        'reshape': {'Precision': 'fixed<20,10>'},
+        'conv': {'Precision': 'fixed<20,10>'},
+    }
+
+    hls_config = {'Model': {'Precision': 'fixed<2,1>', 'ReuseFactor': 1}, 'LayerName': layer_conf}
+
+    # Repack only happens in io_stream
+    model_hls = convert_from_keras_model(model, hls_config=hls_config, io_type='io_stream')
```
jmitrevs, line 27: Should name the output directory in the proper style.

jmitrevs:
By the way, to answer my question about the replaced nodes having one input and output only in the test, I think that's true, but this fix properly handles the case when the node after the replaced node has multiple inputs, which the test demonstrates. So after properly setting the output directories, I think I can merge this PR and make 0.8.1.

jmitrevs (approved):

PR #937, opened 2023-12-07
Fix for multiple inputs that may get out of order

jmduarte:
Fix for multiple inputs that may be in a different order than the order in which the model uses them. Should I add a specific test for this?

## Tests

This breaks in the current main
```python
import hls4ml
from tensorflow.keras.models import Model
from tensorflow.keras.layers import Dense, Input, Add

input_1 = Input(shape=(10), name='input_1')
input_2 = Input(shape=(10), name='input_2')
x = Dense(10, name='dense_1')(input_2)
outputs = Add(name='add_1')([x, input_1])
model = Model(inputs=[input_1, input_2], outputs=outputs, name='model_1')
model.summary()

config = hls4ml.utils.config_from_keras_model(model, granularity='model')
hls_model = hls4ml.converters.convert_from_keras_model(model,
                                                       hls_config=config,
                                                       io_type='io_parallel',
                                                       output_dir='test_fix',
                                                       part='xcvu13p-flga2577-2-e',
                                                       clock_period=5)
hls_model.compile()
```

with the error message:

```
RuntimeError: Currently only support the case when input variables and input layer names match
Input layers = ['input_1', 'input_2'], input_vars = ['input_2', 'input_1']
```

## Checklist

### DIFF
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -343,7 +343,7 @@ class ModelGraph:
         input_layers = inputs if inputs is not None else [layer_list[0]['name']]
         output_layers = outputs if outputs is not None else [layer_list[-1]['name']]
         self.inputs = self._find_output_variable_names(layer_list, input_layers)
-        if self.inputs != input_layers:
+        if sorted(self.inputs) != sorted(input_layers):
             raise RuntimeError(
                 "Currently only support the case when input variables and input layer names match\n"
                 + f"Input layers = {input_layers}, input_vars = {self.inputs}"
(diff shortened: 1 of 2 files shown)


jmitrevs:
I think I like the way this fix works better than the old way. There is just the minor pre-commit issue that needs fixing.

vloncar:
I agree this looks good. A test would be nice, and I don't think it needs to be separate. We can piggyback on an existing test, for example [here](https://github.com/fastmachinelearning/hls4ml/blob/main/test/pytest/test_merge.py#L21). Reversing `in1` and `in2` will cause failures without this fix. Along with the change perhaps a comment saying why was this done would help so we don't forget in the future.

jmduarte:
pre-commit.ci autofix

jmitrevs (approved):

PR #955, opened 2024-01-11
Beginnings of the oneAPI backend

jmitrevs:
# Description

This is the beginning of the oneAPI backend, and is just included here for easier sharing and discussion. It is not ready for merging. Currently only `io_parallel` Dense and activation layers are supported.

Some things to discuss are:
1.  The usage of `std::array` for sending data on the pipe, and also between layers. It is generally easier to send `std::array` than a struct containing C arrays, I believe. (Note, `std::array` is passed as `nnet::array`, so it's possible to change in one place the array definition.)
2.  The addition of `oneapi_types.py` instead of adding the types in `fpga_types.py`. I think  we want to split `fpga_types.py`. Otherwise catapult types will soon also come there.
3.  This uses cmake to build the libraries and testbench. The setup is based on the examples in https://github.com/oneapi-src/oneAPI-samples
4. Currently only oneAPI 2024.0.0 or later is supported. You have to have it installed.

There is more information on the status at the following link:
https://www.icloud.com/keynote/0e6hRDG4E4PazKCJkJ-ur3yTg#oneAPIDev

## Type of change

## Tests

oneAPI has been added as a backend in many tests.

## Checklist


hls4ml/converters/keras/core.py
```diff
@@ -62,6 +62,10 @@ def parse_activation_layer(keras_layer, input_names, input_shapes, data_reader):
 
     if layer['class_name'] != 'Activation':
         layer['activation'] = layer['class_name']
+
+    if layer['activation'] == 'elu':
+        layer['class_name'] = 'ELU'  # always use ELU type for elu, even if passed as activation
```
JanFSchulte, line 67: Do we need this also for the other converters?

jmitrevs: For oneAPI we probably would need it everywhere. If I remember correctly, this has to do with oneAPI kernels not liking function overloading, but I would have to check. It may be worth enabling the pytorch tests that would exercise this.

jmitrevs: Or maybe not. If I remember correctly, for Keras 2 API you could either use an ELU layer (https://keras.io/2.17/api/layers/activation_layers/elu/) or an Activation layer (https://keras.io/2.17/api/layers/core_layers/activation/) with an elu activation function, and this was so that they always became the same node type in hls4ml. I didn't think hls4ml needed the two different types of ELUs.

jmitrevs: I think I need to add 'oneAPI' to `test_pytorch_api.py` to test for this. I'll try that out.

jmitrevs: Actually, the activations succeed without a problem. It is MaxPooling1D with pad of 1 that hangs (while avg pooling works just fine).

jmitrevs:
I think the optimizer to clone outputs needs updates to work with oneAPI.

hls4ml/converters/keras/core.py
jmitrevs: This pooling with padding issue was fixed by [8c827b8](https://github.com/fastmachinelearning/hls4ml/pull/955/commits/8c827b83af689d0efb6a1292efdaed888703e42c)

jmitrevs:
Actually, I believe I was wrong about the optimizer to clone outputs. It was migrated. I added a clone outputs test, and I removed what looks to be the redundant `test_stream_multi_clone.py`

jmitrevs:
I updated some oneAPI-specific documentation and docstrings. I did not update the status, however, I figure we do that globally before the next release, since I think we also want to remove the "experimental" tag from Vitis.

JanFSchulte (approved):

vloncar (approved):
Looks great. Many thanks Jovan!

PR #961, opened 2024-01-29
Leaky Quantized ReLU fix

DrWatt:
# Description
The `quantized_relu` activation layer included in the QKeras library lets the user set the `negative_slope` option which practically change the activation function from the usual ReLU to a LeakyReLU. This change was not perceived by hls4ml, meaning that that information was lost when the HLS_model was created by the library. 
In order to fix this behaviour I have added the QLeakyReLU in the list of supported layers in `model/layers.py` using the `ParametrizedActivation` implementation mimicking the `LeakyReLU` already present and so following a similar path of implementation to the "non-leaky" `quantized_relu`.
Other changes were made to `model/profiling.py` and `utils/config.py` to make them compatible with the new layer.

A couple of other fixes were made in `backends/vivado_accelerator/vivado_accelerator_config.py` (a missing "casting" function when comparing complex objects with strings) and in `model/types.py` (adding the `ap_` to the hls fixed precision type due to errors raised by the vivado compiler)

## Type of change

## Tests

The `quantized_relu` with a `negative_slope` different from the default one was added in the pytest routine covering qkeras' layers. 
A specific test script has been also added to the pytest directory. The results from the new implementation are asserted to be equal to the QKeras layer with a relative tolerance of 0.00001.

## Checklist


hls4ml/model/types.py
```diff
@@ -300,7 +300,7 @@ def saturation_mode(self, mode):
     def __str__(self):
         args = [self.width, self.integer, self.rounding_mode, self.saturation_mode, self.saturation_bits]
         args = ','.join([str(arg) for arg in args if arg is not None])
-        typestring = '{signed}fixed<{args}>'.format(signed='u' if not self.signed else '', args=args)
+        typestring = 'ap_{signed}fixed<{args}>'.format(signed='u' if not self.signed else '', args=args)
```
jmitrevs, line 303: We don't have 'ap_' or 'ac_' prefixes in our internal typestings. This change should be reverted

hls4ml/utils/config.py
```diff
@@ -79,8 +79,8 @@ def _get_precision_from_quantizer(quantizer):
         rnd = "AP_RND_CONV"
         overflow = "AP_SAT"
         if quantizer['class_name'] in ('quantized_relu', 'quantized_relu_po2'):
-            signed = False
-            integer -= 1
+            signed = True
```
jmitrevs, line 82: Why do you think this should be signed?

jmitrevs: It should only be signed for leaky, not for the others.

hls4ml/utils/config.py
```diff
@@ -79,8 +79,13 @@ def _get_precision_from_quantizer(quantizer):
         rnd = "AP_RND_CONV"
         overflow = "AP_SAT"
         if quantizer['class_name'] in ('quantized_relu', 'quantized_relu_po2'):
-            signed = False
-            integer -= 1
+            if quantizer['config']['negative_slope'] != 0.0:
+                print(quantizer)
```
vloncar, line 83: get rid of the print statement and commented out code

hls4ml/writer/vivado_accelerator_writer.py
```diff
@@ -38,8 +38,14 @@ def write_axi_wrapper(self, model):
                 newline += f'static const unsigned N_IN = {inp.size()};\n'
                 newline += f'static const unsigned N_OUT = {out.size()};\n'
                 if self.vivado_accelerator_config.get_interface() == 'axi_stream':
-                    newline += f'typedef {inp_axi_t} T_in;\n'
-                    newline += f'typedef {out_axi_t} T_out;\n'
+                    if 'fixed' in str(inp_axi_t):
```
vloncar, line 41: unlrelated to this PR

test/pytest/q_leaky_relu_test.py
```diff
@@ -0,0 +1,63 @@
...
+        (quantized_relu(10, 5, negative_slope=0.5)),
+        (quantized_relu(10, 5, negative_slope=0.25)),
+    ],
+)
+@pytest.mark.parametrize('backend', ['Vivado', 'Vitis', 'Quartus'])
+@pytest.mark.parametrize('io_type', ['io_parallel', 'io_stream'])
+def test_quantizer(randX_1000_1, quantizer, backend, io_type):
+    '''
+    Test a single quantizer as an Activation function.
+    Checks the type inference through the conversion is correct without just
```
vloncar, line 40: Not sure I understand this sentence :-)

test/pytest/q_leaky_relu_test.py
```diff
@@ -0,0 +1,63 @@
...
+        (quantized_relu(8, negative_slope=0.125)),
+        (quantized_relu(8, 4, negative_slope=1.0)),
+        (quantized_relu(10, negative_slope=0.25)),
+        (quantized_relu(10, 5, negative_slope=0.5)),
+        (quantized_relu(10, 5, negative_slope=0.25)),
+    ],
+)
+@pytest.mark.parametrize('backend', ['Vivado', 'Vitis', 'Quartus'])
+@pytest.mark.parametrize('io_type', ['io_parallel', 'io_stream'])
+def test_quantizer(randX_1000_1, quantizer, backend, io_type):
```
vloncar, line 37: This should have a proper name, and could be fused with the existing qkeras test file. Or split along with the other qactivation tests. But whatever the solution ends up being it should be properly named with `test_` prefix.

test/pytest/test_qkeras.py
```diff
@@ -272,6 +272,7 @@ def randX_1000_1():
         (quantized_relu(8, 4)),
         (quantized_relu(10)),
         (quantized_relu(10, 5)),
+        (quantized_relu(10, 5, negative_slope=0.25)),
```
vloncar, line 275: Aren't you already testing this in the other test?

vloncar (changes requested):
Looks good, needs cleanup. Thanks Marco!

hls4ml/model/types.py
DrWatt: I have removed the `ap_` here and fixed the vivado_accelerator writer which gave issues with the compiler due to the `my_project_axi.hpp` not using the hls type definition but directly the internal typestrings. However this fix has been removed from this branch to separate the fixes.

hls4ml/utils/config.py
DrWatt: Implemented an `if` statement which makes the activation signed only if the negative slope is different from 0.

.gitignore
```diff
@@ -14,3 +14,4 @@ docs/autodoc/*
 hls4mlprj_*
 *~
 *.ipynb_checkpoints/
+*.bak
```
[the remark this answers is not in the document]
DrWatt: Exception removed, my bad, it was here for the byproducts of running precommit manually.

hls4ml/backends/vivado_accelerator/vivado_accelerator_config.py
```diff
@@ -68,16 +68,16 @@ def __init__(self, config, model_inputs, model_outputs):
         if out_axi_t not in ['float', 'double']:
             self.output_type = self._next_factor8_type(config.backend.convert_precision_string(out_axi_t))
 
-        if self.input_type == 'float':
+        if str(self.input_type) == 'float':
```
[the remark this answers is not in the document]
DrWatt: I have removed all the fixes concerning the vivado_accelerator backend and writer (small stuff), should I open another PR for those?

hls4ml/converters/keras/qkeras.py
```diff
@@ -147,6 +146,10 @@ def parse_qactivation_layer(keras_layer, input_names, input_shapes, data_reader)
         layer['slope_prec'] = FixedPrecisionType(width=2, integer=0, signed=False)
         layer['shift_prec'] = FixedPrecisionType(width=2, integer=0, signed=False)
         layer['activation'] = activation_config['class_name'].replace('quantized_', 'hard_')
+    elif activation_config['class_name'] == 'quantized_relu' and activation_config['config']['negative_slope'] != 0:
+        layer['class_name'] = 'QLeakyReLU'
```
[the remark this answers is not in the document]
DrWatt: You are right, I wanted to be clear when writing the code for my sake but indeed using `LeakyReLU` makes the fix simpler.

hls4ml/utils/config.py
DrWatt: Print and commented code removed

test/pytest/q_leaky_relu_test.py
DrWatt: I am sorry, it does not make any sense, I think it was a copy and paste mishap.

test/pytest/test_qkeras.py
DrWatt: In order to create the test suite requested for the PR I added the line in the already present qkeras test, however due to problems with the environment requested for the pytests, some issues not concerning my fix came up. So, I have written a separate script just to show that the "new" layer passes an accuracy test. If the `test_qkeras.py` test passes when called in the automatic workflow, my other script can be removed.

hls4ml/model/profiling.py
```diff
@@ -589,8 +589,9 @@ def get_ymodel_keras(keras_model, X):
         name = layer.name
         if (
             hasattr(layer, "activation")
-            and layer.activation.__name__ != "linear"
+            and layer.activation is not None
```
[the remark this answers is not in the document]
DrWatt: I have found out the if the QBatchNormalization layer is used without specifying an activation function, the default `layer.activation` is None, which breaks the if statement when the `linear` check is performed. Now that I am thinking about this again, I am not sure if another case should be added in this situation by forcing a linear activation.

jmitrevs:
Sorry for the delay, but what is the status of this? It looks to me like maybe I can approve and merge it.

jmitrevs (approved):

vloncar:
I was testing this and the changes work, however the PR includes unnecessary changes and the test doesn't do anything (data is all positive, so it passes even without the changes in the converter). I distilled the changes and made a proper test in https://github.com/vloncar/hls4ml/tree/qrelu_negative_slope

vloncar:
I made #987 as a continuation of this PR

PR #964, opened 2024-02-08
Change fractional (and others) to be a property, move quantizers 

jmitrevs:
# Description

This PR has two somewhat unrelated changes. I made the second after the first, so I don't have the second standalone (but can if necessary), and it's easy to extract just the first since it's the first commit. The changes are:

1.  Currently `FixedPrecisionType` and `IntegerPrecisionType` have the values `width`, `integer`, and `fractional` set as values stored in the classes. Though they are set properly initially, the values are mutable, so it's possible to edit the values after the fact. It's up to whoever modifies them to make sure they remain consistent. I think that's error-prone. For example, one could modify `width` in `IntegerPrecisionType` but leave `integer` alone, leaving a malformed class. What this PR does is make `fractional` a property of `FixedPrecisonType`, always returning `width` - `integer`, and a property of `IngegerPrecisionType`, always returning `0`. Similarly `integer` is a property of  `IntegerPrecisionType`, just returning `width`. They don't have setters. I also modified the saturation and rounding properties to return the default value if they are set to None, and added similar properties to IntegerPrecisionType but unchangeable, since they are not changeable in the type.
2. Move quantizers out of `types.py` into a new `quantizers.py` file. For qonnx we will add additional quantizers, so my thinking is that it's better to have them in a separate file. But this is not a major issue, so I would be fine reverting it if there are objections.

## Type of change

- [x] Other (Specify)
 Code restructuring

## Tests

There should be no changes so the old tests should confirm that I didn't introduce an error.

## Checklist


hls4ml/model/types.py
```diff
@@ -237,20 +81,36 @@ class IntegerPrecisionType(PrecisionType):
...
+    def rounding_mode(self):
+        return RoundingMode.TRN
+
+    @property
+    def saturation_mode(self):
+        return SaturationMode.WRAP
+
+    @property
+    def saturation_bits(self):
+        return None
```
jmitrevs, line 113: Should this be 0 instead of None?

vloncar: yes. good catch! i left it while debugging :man_facepalming:

hls4ml/backends/fpga/fpga_types.py
```diff
@@ -71,6 +75,10 @@ def definition_cpp(self):
             self._saturation_mode_cpp(self.saturation_mode),
             self.saturation_bits,
         ]
+        if args[3] == 'AC_TRN' and args[4] == 'AC_WRAP' and args[5] == 0:
+            # This is the default, so we won't write the full definition for brevity
+            args[3] = args[4] = args[5] = None
```
jmitrevs, line 80: Does ac_fixed actually have saturation bits? It might be tangental, but I didn't see it in the documentation or a cursory look at the source code. If it's not supported, we should maybe add a warning if it's set. (I don't fully understand how those are used.)

vloncar: ok, I'll add the warning under this check to inform the user

jmitrevs:
I think this looks good. Since I initially submitted this PR I can't approve it, but if you are happy with it, go ahead and merge it.

Issue #957, opened 2024-01-12
Move transpose based on backend from hls4ml/model/layers.py to backend-specific areas


jmitrevs:
## Prerequisites

- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.

## Quick summary

Backed-specific code should generally not exist in `hls4ml/model/layers.py`, so the backend-specific transposing of weights should not be part of the standard internal representation (IR).

## Details
For `Conv2DBatchnorm` in the standard IR there is a different transpose made based on the backend:

https://github.com/fastmachinelearning/hls4ml/blob/eb70a61d0da77d880de055065e3837de5846cc72/hls4ml/model/layers.py#L560-L569

The standard IR should not depend on the backend. This logic needs to be moved to backend-specific code.

PR #965, opened 2024-02-08
Remove double transpose for Conv2DBatchnorm

jmitrevs:
# Description

In the `layers.py` `Conv2DBatchnorm` makes a transpose if it is the resource strategy, but only for `Vivado` and `VivadoAccelerator` backends. Note that `Conv2DBatchnorm` inherits from `Conv2D`, and `Conv2D` makes the same transpose in:

https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/vivado/passes/resource_strategy.py#L29-L32

I think therefore the transpose is doubly-applied. I am therefore removing the backend-specific transpose in the `layers.py`, which regardless does not belong there.

## Type of change

## Tests

I assume there's a `Conv2DBatchnorm` test that will catch this change, but maybe not.

**Test Configuration**:

## Checklist

### DIFF
--- a/hls4ml/model/layers.py
+++ b/hls4ml/model/layers.py
@@ -557,16 +557,8 @@ class Conv2DBatchnorm(Conv2D):
     def initialize(self):
         super().initialize()
         folded_weights, folded_bias = self._get_folded_weights()
-        if self.model.config.is_resource_strategy(self) and self.model.config.backend.name in [
-            'Vivado',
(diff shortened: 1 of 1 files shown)


jmitrevs:
This was motivated by issue #957

jmitrevs:
Conv2DBatchnorm seems buggy in general. We may want a more serious look at it.

PR #967, opened 2024-02-09
Try to fix sphinx problem by restricting tensorflow-model-optimization

jmitrevs:
# Description

The sphinx build failed. This tries to restrict incompatible versions

## Tests

Check the sphinx build

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ install_requires =
     qkeras
     tabulate
     tensorflow
+    tensorflow-model-optimization<=0.7.5
 python_requires = >=3.8
 include_package_data = True
 scripts = scripts/hls4ml


jmitrevs:
@jmduarte, @vloncar, is this the type of fix you think is appropriate for sphinx? This is actually for everyone that install hls4ml.

vloncar:
After some more digging, this is due to the split from keras and TF and the way this is handled in TF 2.15 onwards. Apparently, TFMOT works only with keras 2.x from TF (`tf.keras`) and after the split this is provided by the [tf_keras](https://pypi.org/project/tf-keras/) package. TFMOT added this to `requrements.txt` but not to `setup.py` so it doesn't get installed on fresh environments. I opened tensorflow/model-optimization#1117 to track this. I propose we wait for the resolution of that one in a timely manner, and if not, we go with this workaround.

jmitrevs:
Do we need tf_keras in all cases? Maybe we should add it to our requirements.

If we add tf_keras then we don't need to restrict tensorflow-model-optimization, right?

jmitrevs:
I am going to try the alternate solution to see if it works.

vloncar:
tf_keras only exists for 2.14.1 and 2.15. Not sure if this will force the tensorflow to that version or if not, will it affect precious tensorflow versions. I also didn't figure out what would happen if I install keras 3. This situation is a bit of a mess :slightly_smiling_face:

jmitrevs:
The alternate version may be more problematic based on your comments, so I won't push it. (I do think it works in my setup, though.) So we wait to see if it gets solved upstream first?

jmitrevs:
Ideally this is a *temporary* fix that will be reverted.

PR #947, opened 2023-12-21
Generalize pooling accumulator inference

jmduarte:
Currently, pooling uses a wider data type for accumulation inferred based on the input type and the number of elements. However, we only implemented this for `ap_fixed<W,I>` and `ap_int<W>` and the template matching fails for `ap_uint<W>` and `ap_ufixed<W,I,Q,O>` and `ap_fixed<W,I,Q,O>` (if `Q` and `O` are not the default values). This PR (hopefully) makes it so it matches all those additional types.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Tests

Example before this fix:
profiling_global_average_pooling1d (1)

Example after this fix: 
profiling_global_average_pooling1d (2)

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have installed and run `pre-commit` on the files I edited or added.
- [ ] I have added tests that prove my fix is effective or that my feature works.

@Brainz22 @ddiaz006 @AnthonyAportela @danprim

PR #917, opened 2023-11-09
[DRAFT] Fix pooling accum_t autoset & avoid global override

calad0i:
A# Description

No test at the moment - Fix a series of strange `accum_t` behaviors for pooling layers in the current version. Potential conflict with #855, would need to converge first before finalizing changes.

Strange behaviors:
- `model_default` precision overridden
- Extra integer bits were added, while the number of fp bits decreased 

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)

## Tests

Not yet available.

**Test Configuration**:

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [ ] I have added tests that prove my fix is effective or that my feature works.

PR #973, opened 2024-02-21
Latency Pooling Header Updates

calad0i:
## Precision override moved to Requires #855 now. Expect tests to fail without.

Include and superceeds #917. On the same issue: #947

This PR does the following, for the `vivado` and `vitis` backends.

- make avg pooling layers in `io_parallel` mode respect to `accum_t`.
- refactoring codes for pooling layers for `io_parallel`.
  - Performance validated to be equal or better. 
  - The new implementation avoids a few potential overflowing in `tmp` and `scale` in the original avg pool implementation. 
~~- Avoid overriding global variable names when automatically overriding precisions; derive `accum_t` properly from impending layer's `result_t`.~~ Moved to #855 

With the current setting, pooling layers will be bit-accurate if the impending layer's `result_t` is configured properly.

Synthesis on some toy models table

## Type of change

## Tests

Will be covered in #914 (WIP).

## Checklist


calad0i:
pre-commit.ci autofix

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -375,10 +375,38 @@ def init_depconv2d(self, layer):
 
     def _set_pooling_accum_t(self, layer, pool_size):
         extra_bits = ceil_log2(pool_size)
+        if 'inputs' in layer.attributes:
+            # Functional
+            input_layer_name = layer.attributes['inputs'][0]
+        else:
+            # Sequential
```
jmitrevs, line 382: I don't fully understand why the if-else is requried. Can't you always query for the previous node?

jmitrevs:
I wonder if this is part of the inferring of precisions that's in the works: #855. Let me try to update #855 (by making pull requests to it) so we can merge it in soon.

hls4ml/backends/vivado/vivado_backend.py
calad0i: If we have a functional model that has A->B, A->C, with C being the pooling layer. As the order of B and C is arbitrary, if we query the precious node, we can get B, the sibling of the current node, instead of A.

jmitrevs: I don't quite follow. Are the node inputs not correctly filled in some cases?

calad0i: Not sure for now, but they were not as of Nov 2023. Only for the functional model converted graph, the parent names are filed; for sequential models, the fields were missing.
Could check this again quickly.

calad0i: Confirmed.

jmitrevs: But look at the jettagging from the hls4ml-tutorial:
```python
model = Sequential()
model.add(Dense(64, input_shape=(16,), name='fc1', kernel_initializer='lecun_uniform', kernel_regularizer=l1(0.0001)))
model.add(Activation(activation='relu', name='relu1'))
model.add(Dense(32, name='fc2', kernel_initializer='lecun_uniform', kernel_regularizer=l1(0.0001)))
model.add(Activation(activation='relu', name='relu2'))
model.add(Dense(32, name='fc3', kernel_initializer='lecun_uniform', kernel_regularizer=l1(0.0001)))
model.add(Activation(activation='relu', name='relu3'))
model.add(Dense(5, name='output', kernel_initializer='lecun_uniform', kernel_regularizer=l1(0.0001)))
model.add(Activation(activation='softmax', name='softmax'))
```
and looking at the node object for `fc2` you can see that the inputs are `['relu1]`:

jmitrevs: image

jmitrevs: So you can use `get_input_node` or something similar to get the node. https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/layers.py#L182-L192

calad0i: Just checked and cannot get the inputs with current hls4ml master.... weird.

jmitrevs: Note that it's `self.inputs`, not `self.attributes['inputs']`.

jmitrevs: (on the Layer object)

calad0i: Confirmed, modified PR.

jmitrevs:
There seem to be some real test failures associated with this PR. @calad0i, can you take a look?

hls4ml/templates/vivado/nnet_utils/nnet_pooling.h
```diff
@@ -114,32 +95,19 @@ void pooling1d_cl(data_T data[CONFIG_T::n_in * CONFIG_T::n_filt], res_T res[CONF
...
             data_T pool[CONFIG_T::pool_width];
             #pragma HLS ARRAY_PARTITION variable=pool complete dim=0
-            // Keep track of number of pixels in image vs padding region
-            unsigned img_overlap = 0;
-            // Loop over pool window x
-            for (int jj = 0; jj < CONFIG_T::stride_width; jj++) {
-                if (ii + jj < CONFIG_T::pad_left || ii + jj >= (padded_width - CONFIG_T::pad_right)) {
-                    // Add padding
-                    pool[jj] = pad_val<data_T, CONFIG_T::pool_op>();
-                    if (CONFIG_T::count_pad)
```
jmitrevs, line 126: It seems like the new code ignores the count_pad option. Is it fine to get rid of it?

vloncar: This was introduced to ensure compatibility with PyTorch. Not sure if it can be removed without breaking bit-perfect agreement with pytorch

jmitrevs: The count_pad option seems to also be ignored in general in streaming pooling.

jmitrevs: Actually, padding is not currently supported in streaming, so it's fine that it's ignored.

calad0i:
Fixed vitis and mnist tests. Pytorch test directly crashed, need some debugging to see what is wrong.

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -376,12 +376,27 @@ def init_depconv2d(self, layer):
...
+
+        accum_t.name = f'{layer.name}_accum'
+        # accum_t.name was likely model_default, change to avoid override parameters for other by chance
+
         accum_t = layer.get_attr('accum_t')
-        accum_t.precision.width += extra_bits * 2
-        if isinstance(accum_t.precision, FixedPrecisionType):
-            accum_t.precision.integer += extra_bits
+
+        if pool_op == 'max':
```
jmitrevs, line 392: This should not be set here. The user can override it, in which case we should follow what the user requests. Otherwise, if this was set to "auto", then the automatic precision inference (#855) will do something similar. We can check if it does the right thing there and fix it if needed (but let me check that my updates that went in that PR first before adding code to it.)

jmitrevs: This is the current version that I have:  https://github.com/jmitrevs/hls4ml/blob/qonnx-1p0/hls4ml/model/optimizer/passes/infer_precision.py. I will try to extract it and make a PR to 855 with it within the next 2 hours or so.

jmitrevs: I made a PR to Vladimir's branch:  https://github.com/vloncar/hls4ml/pull/55 . You can probably do the same. I don't think I explicitly did anything for pooling, but there is the merge Average and Add that probably have the same logic in my PR.

calad0i: Opened PR at [#56](https://github.com/vloncar/hls4ml/pull/56). This PR will be updated if that one got merged.

hls4ml/backends/vivado/vivado_backend.py
```diff
@@ -376,37 +373,14 @@ def init_depconv2d(self, layer):
         )  # TODO Once we have SeparableConv implementation for io_parallel this should be set properly
         layer.set_attr('implementation', layer.model.config.get_conv_implementation(layer).lower())
 
-    def _set_pooling_accum_t(self, layer, pool_size):
```
jmitrevs, line 379: We shouldn't merge this change before we have the subsequent change in auto precision inference. Right now it's removing features that already exist (and catapult customers may be using) without replacing it right away. For Vitis and Vivado it is fine to not add something that doesn't yet exist, but one has to be more careful about removing already existing features. If we want to decouple things, we should move the catapult update to a separate PR.

calad0i: The current `_set_pooling_accum_t` is flawed (overrides `default_t` in most cases, doesn't always fix overflow as well; multiple pooling layers will results in a huge default model bitwidth thanks to the `+=` operation), and I fixed it in the last iteration of this PR.

This PR will not work without proper `accum_t` set, and that's why I fixed it right here in the last iteration. As you suggested, I integrated it with #855, and I have marked #855 as required by this PR.

jmitrevs: Generally, unless they are code maintenance-type changes or purely adding new features, Catapult backend changes need Siemens approval.

jmitrevs: We are hoping to merge #855 soon. The last changes failed the pytests, so after chaging "ave" to "average" we are trying again. Vivado should not have the overrides since it never had them. It should start with the new way of doing things from the beginning.

jmitrevs: I would probably make the Catapult change a separate PR, though. We do eventually want that.

calad0i: As #855 does not remove that override, I will remove the Vivado override and leave the Siemens one as-is.

jmitrevs: I hadn't appreciated that there was already some auto precision setting for Vivado. Nevertheless, I believe removing it is the right thing to do since we have the auto precision. I will announce it.

calad0i:
`test/pytest/test_weight_writer.py` is not updated in this PR though...
pre-commit.ci autofix

jmitrevs:
Don't worry about the test_weight_writer.py

jmitrevs:
I am not sure if our pytest environment is functional. I can try running them again, but they may fail for infrastructure reasons

jmitrevs:
The `test_qkeras.py::test_qgru[Vivado]` failure we can ignore. For cnn_mnist, why does this error cause the failure? We mentioned about possibly removing that test, but I want to understand why it's failing. Would setting the accumulator to auto fix the issue?

calad0i:
> The `test_qkeras.py::test_qgru[Vivado]` failure we can ignore. For cnn_mnist, why does this error cause the failure? We mentioned about possibly removing that test, but I want to understand why it's failing. Would setting the accumulator to auto fix the issue?

Correct. The reason to fail is exactly overflow in the avg pool, which had an overriding mechanism but is now removed.
Adding a line `hls_config['LayerName']['avg_pool']['Precision']['accum'] = 'auto'` in the test and name the layer properly will fix it.

jmitrevs:
Let's add the fix to the test and maybe merge this? I think making auto a default for some layer types could be a different PR, and I think it's really a change in the `config_from_keras_model` functionality.

jmitrevs:
We can fix the failing `test_cnn_mnist.py` with:
```
(fastml310) jovans-mac:pytest jmitrevs$ git diff
diff --git a/test/pytest/test_cnn_mnist.py b/test/pytest/test_cnn_mnist.py
index ab3365f2..a8b77862 100644
--- a/test/pytest/test_cnn_mnist.py
+++ b/test/pytest/test_cnn_mnist.py
@@ -70,6 +70,8 @@ def test_mnist_cnn(keras_model, mnist_data, backend, io_type, strategy):
     hls_config = hls4ml.utils.config_from_keras_model(keras_model, granularity='name', backend=backend)
     hls_config['Model']['Strategy'] = strategy
     hls_config['LayerName']['softmax']['Implementation'] = 'stable'
+    hls_config['LayerName']['average_pooling2d']['Precision']['accum'] = 'auto'
+    hls_config['LayerName']['max_pooling2d']['Precision']['accum'] = 'auto'
     output_dir = str(test_root_path / f'hls4mlprj_cnn_mnist_{backend}_{io_type}_{strategy}')
 
```

jmitrevs (approved):

PR #975, opened 2024-02-23
remove special vitis version of nnet_dense_resource.h

jmitrevs:
# Description

The reasons to have a Vitis version of `nnet_dense_resource.h` seem to have been fixed, and the Vivado version uses much fewer resources. For example, in a 2D CNN model used in a DUNE study, reverting to the old Vivado version makes the BRAM usage estimation go from 6060 to 206, as estimated by Vitis HLS 2023.2 HLS synthesis. (Vivado estimate comparison to come later.)

The minimal Vitis HLS version for this is yet to be determined. I will test it for 2022.2 at least and report below

## Type of change.

## Tests

We should test that the standard pytests still pass, and offline I am testing synthesis. I will report the results below.

## Checklist


jmitrevs:
I confirmed that this works well with 2022.2, as well (at least for synthesis). The Vivado implementation in 2023.2 has a bram usage of 175, lower than the synthesis estimate. So I think we need to remove the special vitis version.

PR #979, opened 2024-03-12
Update QONNX parsing for 1.0

jmitrevs:
# Description

This change updates the ONNX parser and adds support for QONNX. It replaces PR https://github.com/fastmachinelearning/hls4ml/pull/832. It only supports ONNX that has been cleaned by the [qonnx](https://github.com/fastmachinelearning/qonnx) package, including converting convolutions to be channels-last and changing Gemm to MatMul and Add.

In QONNX Quant nodes can act on constants as well as the datapath. To make handling this easier, we explicitly put constants in the initial graph. There are also some helper nodes like MatMul and Conv that are introduced to support the explicit constant nodes. After the convert flow, no special ONNX nodes remain in the graph, though.

Generally Quant nodes that have power-of-2 scales and no zero-offset get converted to fixed data types either by setting the types of constants or adding a linear activation that is usually merged into preceding nodes. Non-power-of-2 scales result in ApplyAlpha nodes beings added to scale and unscale, with propagation across some layers. This can be further optimized and has generally been tested less.

This includes the changes from PR #855 with a few updates that will be backported and discussed there. Therefore, this PR needs to wait till that PR is merged, which is why I am making it draft. 

Note:  for the `config_from_onnx_model` I made the default granularity be "name" because that enables automatic precision inference, which you need for QONNX. The way that I did that is that I set `config['Model']['Precision']` to the default (e.g. `fixed<16,6>`), but all the precisions filled by `config['Model']` are `auto`. These can be overriden if, for example, the accumulator becomes too wide. In general, though, they are set by the `infer_precision.py` optimizer.

Binary networks are not yet supported.

More information can be found in this presentation:
https://www.icloud.com/keynote/025yxvgBx8IF2m3Iso6HosqPw#QONNX_Ingestion_0p1

## Type of change

## Tests

The pytest, `test_qonnx.py`, is the main test, building some models from the [QONNX model zoo](https://github.com/fastmachinelearning/qonnx_model_zoo)

## Checklist


sei-jgwohlbier:
Hi, I'm not sure if this is the right place for this, but I'm trying to synthesize the `cnv-2w2a` example from the `test_qonnx.py` file. It has been running for nearly two days now without a sign of progressing. Have you synthesized this? Do you have an idea of how long it took? Thanks!

Output
```
****** Vivado(TM) HLS - High-Level Synthesis from C, C++ and SystemC v2019.2 (64-bit)
  **** SW Build 2708876 on Wed Nov  6 21:39:14 MST 2019
  **** IP Build 2700528 on Thu Nov  7 00:09:20 MST 2019
    ** Copyright 1986-2019 Xilinx, Inc. All Rights Reserved.

source /tools/Xilinx/Vivado/2019.2/scripts/vivado_hls/hls.tcl -notrace
INFO: Applying HLS Y2K22 patch v1.2 for IP revision
INFO: [HLS 200-10] Running '/tools/Xilinx/Vivado/2019.2/bin/unwrapped/lnx64.o/vivado_hls'
INFO: [HLS 200-10] For user 'hls4ml-user' on host 'etc-gpu-09.etc.sei.cmu.edu' (Linux_x86_64 version 4.18.0-513.24.1.el8_9.x86_64) on Fri Jun 21 14:33:16 UTC 2024
INFO: [HLS 200-10] In directory '/home/hls4ml-user/work/hls4mlprj_qonnx_cnv-2w2a_Vivado'
Sourcing Tcl script 'build_prj.tcl'
INFO: [HLS 200-10] Creating and opening project '/home/hls4ml-user/work/hls4mlprj_qonnx_cnv-2w2a_Vivado/myproject_prj'.
INFO: [HLS 200-10] Adding design file 'firmware/myproject.cpp' to the project
INFO: [HLS 200-10] Adding test bench file 'myproject_test.cpp' to the project
INFO: [HLS 200-10] Adding test bench file 'firmware/weights' to the project
INFO: [HLS 200-10] Adding test bench file 'tb_data' to the project
INFO: [HLS 200-10] Creating and opening solution '/home/hls4ml-user/work/hls4mlprj_qonnx_cnv-2w2a_Vivado/myproject_prj/solution1'.
INFO: [XFORM 203-101] Allowed max sub elements number after partition is 4096.
INFO: [XFORM 203-1161] The maximum of name length is set into 80.
INFO: [HLS 200-10] Setting target device to 'xcvu13p-flga2577-2-e'
INFO: [SYN 201-201] Setting up clock 'default' with a period of 5ns.
INFO: [SYN 201-201] Setting up clock 'default' with an uncertainty of 0.625ns.
***** C SIMULATION *****
INFO: [SIM 211-2] *************** CSIM start ***************
   Generating csim.exe
INFO: Unable to open input/predictions file, using default input.
-1.4662 -1.68291 -0.174961 -0.68965 -0.22011 -1.15919 -0.915391 -0.969569 -0.987628 -1.29464 
INFO: Saved inference results to file: tb_data/csim_results.log
INFO: [SIM 211-1] CSim done with 0 errors.
***** C SIMULATION COMPLETED IN 0h1m0s *****
***** C/RTL SYNTHESIS *****
INFO: [SCHED 204-61] Option 'relax_ii_for_timing' is enabled, will increase II to preserve clock frequency constraints.
INFO: [HLS 200-10] Analyzing design file 'firmware/myproject.cpp' ... 
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:69:75
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:69:81
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:73:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:73:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:77:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:77:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:81:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:81:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:85:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:85:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:93:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:93:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:97:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:97:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:101:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:101:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:105:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:105:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:113:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:113:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:117:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:117:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:121:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:121:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:125:93
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:125:99
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:130:90
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:130:96
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:134:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:134:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:138:90
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:138:96
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:142:94
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:142:100
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:146:90
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:146:96
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:148:95
WARNING: [HLS 214-113] Either use an argument of the function or declare the variable inside the dataflow loop body: firmware/myproject.cpp:148:101
WARNING: [HLS 200-471] Dataflow form checks found 38 issue(s) in file firmware/myproject.cpp
INFO: [HLS 200-111] Finished Linking Time (s): cpu = 00:13:53 ; elapsed = 00:14:19 . Memory (MB): peak = 892.973 ; gain = 463.164 ; free physical = 21561 ; free virtual = 168305
INFO: [HLS 200-111] Finished Checking Pragmas Time (s): cpu = 00:13:53 ; elapsed = 00:14:19 . Memory (MB): peak = 892.973 ; gain = 463.164 ; free physical = 21561 ; free virtual = 168305
INFO: [HLS 200-10] Starting code transformations ...
```

hls4ml/converters/onnx/core.py
```diff
@@ -37,7 +29,7 @@ def parse_gemm_layer(reader, node, inputs_map, input_shapes, graph, config):
     'Softmax',
     'Softsign',
     'Softplus',
-    'Clip',
+    # 'Clip',
```
JanFSchulte, line 32: Remove commented code?

hls4ml/converters/onnx/core.py
```diff
@@ -53,70 +45,89 @@ def parse_gemm_layer(reader, node, inputs_map, input_shapes, graph, config):
     'Softmax': 'Softmax',
     'Softsign': 'Activation',
     'Softplus': 'Activation',
-    'Clip': 'Clip',
+    # 'Clip': 'Clip',
```
JanFSchulte, line 48: Remove commented code?

hls4ml/converters/onnx/merge.py
```diff
@@ -21,22 +33,10 @@ def parse_merge_layer(reader, node, inputs_map, input_shapes, graph, config):
         layer['op'] = layer['class_name'].lower() + f'{rank}d'
         layer['axis'] = get_onnx_attribute(node, 'axis')
 
-        # Calculate output shape
-        new_dim = sum(
-            [x.type.tensor_type.shape.dim[layer['axis']].dim_value for x in graph.value_info if x.name in node.input]
-        )
-        output_shape[layer['axis']] = new_dim
-
-    elif layer['class_name'] == 'Add':
```
JanFSchulte, line 30: Just for my understanding, what is the reason that BiasAdd is no longer supported?

hls4ml/model/optimizer/passes/batchnorm_opt.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def transform(self, model, node):
+        """
+        Remove Constant from the BatchNormalization node parameters (but not input[0])
+
+        TODO:  Currently the quantizers are not actually used by the underlying layer.
+        """
+
+        if not (len(node.inputs) == 5 and all(node.inputs)):
+            raise ValueError(f'All {len.node.inputs} BatchNormOnnnx inputs need to be defined')
```
JanFSchulte, line 27: Not sure I understand the error message here, shouldn't it just read "All 5 BatchNormOnnnx inputs need to be defined" since it will also throw the error if all inputs are defined, but their number does not equal 5?

hls4ml/model/optimizer/passes/batchnorm_opt.py
```diff
@@ -0,0 +1,269 @@
...
+
+        attributes = {k: node.attributes.get(k, None) for k in _base_attributes}
+
+        gamma_node = node.get_input_node(node.inputs[1])
+        if not isinstance(gamma_node, Constant):
+            raise TypeError('Only consant gammas supported')
```
JanFSchulte, line 33: typo, "constant"

hls4ml/model/optimizer/passes/batchnorm_opt.py
```diff
@@ -0,0 +1,269 @@
...
+        # remove the batch norm node
+        model.remove_node(node, rewire=True)
+
+        return True
+
+
+class FuseConsecutiveBatchNormalization(OptimizerPass):
+    """
+    OptimizerPass to merge consecutive BatchNormalization layers,
+    only if the earlier one does not have quantization specified
```
JanFSchulte, line 151: The code below seems to match also in the case when the current node does not have quantization specified, not just the earlier one. Does this description need to be updated?

hls4ml/model/optimizer/passes/batchnorm_opt.py
```diff
@@ -0,0 +1,269 @@
...
+        prev_node = node.get_input_node(node.inputs[0])
+
+        prev_map = prev_node.get_output_use_map()
+        if len(prev_map[prev_node.outputs[0]]) > 1:
+            return False
+
+        # # Not sure why this part is needed
+        # node_map = node.get_output_use_map()
+        # if len(node_map[node.outputs[0]]) > 1:
+        #     return False
```
JanFSchulte, line 196: remove commented code?

hls4ml/model/optimizer/passes/bn_fuse.py
```diff
@@ -1,38 +1,97 @@
...
 
 class FuseBatchNormalization(OptimizerPass):
+    """
+    OptimizerPass to merge BatchNormalization layers,
+    only if the earlier one does not have quantization specified
+
+    Note:  Consider restricting this to ApplyAlpha.  Batch Normalization quantization seems to be ignored.
+
+    Note:  This optimizer may not be safe if weights are updateable. May need to turn off.
+    """
```
JanFSchulte, line 16: This seems to be the doc string from ` FuseConsecutiveBatchNormalization` above and should probably be replaced with one appropriate to this function.

hls4ml/model/optimizer/passes/bn_fuse.py
```diff
@@ -1,38 +1,97 @@
...
         parent_map = parent_node.get_output_use_map()
-        node_map = node.get_output_use_map()
-        if len(parent_map[parent_node.name]) > 1 or len(node_map[node.name]) > 1:
+        if len(parent_map[parent_node.outputs[0]]) > 1:
             return False
 
+        # # Not sure why this part is needed
+        # node_map = node.get_output_use_map()
+        # if len(node_map[node.outputs[0]]) > 1:
+        #     return False
```
JanFSchulte, line 55: remove commented code.

hls4ml/model/optimizer/passes/conv_to_convxd.py
```diff
@@ -0,0 +1,89 @@
...
+        if len(node.inputs) == 3:
+            bias_node = node.get_input_node(node.inputs[2])
+
+        # creating the attributes
+        attributes = {k: node.attributes.get(k, None) for k in _base_attributes}
+
+        # The ConvxD nodes expect the weight data to be in a different format, not (M, k1.., C)
+        if node.attributes['n_dim'] == 1:
+            newtype = Conv1D
+            attributes['weight_data'] = np.transpose(weight_data, (1, 2, 0))
```
JanFSchulte, line 58: Out of curiosity, why do we need to transpose the weights here when the model is supposed to have been processed by `qonnx-to-channels-last` already?

JanFSchulte (approved):
I left a bunch of very minor comments, otherwise this seems ready for merge from my point of view.

hls4ml/converters/onnx/core.py
jmitrevs: Clip will not be hard to add, and we probably should do it, but I'll remove it for now.

hls4ml/converters/onnx/core.py
jmitrevs: Removed for now.

hls4ml/converters/onnx/merge.py
jmitrevs: It is handled as just a regular merge with add.

jmitrevs: Merge with constant becomes a batchnorm, and it can then be fused with a dense layer if there are no conflicts.

jmitrevs: I did wonder if adds and scales should be kept separate, but there's a whole infrastructure setup already for scale+add, which is what batch norms become in the end. (We don't support properly quantized batch-norms, where the mean, etc., is quantized, by the way, even for QKeras.)

hls4ml/model/optimizer/passes/batchnorm_opt.py
jmitrevs: I think they need to be defined and valid. The values are not optional, according to my understanding of the ONNX description. This optimizer reads those inputs, which should be Constant nodes (potentially quantized--that's why the quant-constant folding is scheduled earlier), and replaces them with '', to not shift the index of the inputs. So after this optimizer, `any(node.inputs[1:])` should return false.

hls4ml/model/optimizer/passes/batchnorm_opt.py
jmitrevs: Fixed.

hls4ml/model/optimizer/passes/batchnorm_opt.py
jmitrevs: I may have been referring to the output variable not having a special quantization required (i.e. it's `UnspecifiedPrecisionType`). I did play around with the requirements after the fact, so I should maybe update the description.

jmitrevs: I do wonder if the matching is too loose. It basically is saying that if at least one bias or scale is unquantized, then the resulting merged values are unquantized. This doesn't increas the number of unquantized scales of biases.

hls4ml/model/optimizer/passes/bn_fuse.py
jmitrevs: deleted

hls4ml/model/optimizer/passes/bn_fuse.py
jmitrevs: Attempted to update.

hls4ml/model/optimizer/passes/conv_to_convxd.py
jmitrevs: I think the weight data is still not in the same order in ONNX as expected by hls4ml, which follows the Keras order. It's not the channels index that is moved but the feature map index. Let me know if it's not correct, because it was a bit found by experimentation.

JanFSchulte: I am not very familiar with how this is handled in QONNX, but at first glance I don't see the weights being transposed in https://github.com/fastmachinelearning/qonnx/blob/main/src/qonnx/transformation/channels_last.py, so you are probably right. I just don't understand the logic in QONNX then :-D

hls4ml/model/optimizer/passes/batchnorm_opt.py
JanFSchulte: I was wondering more about the number of inputs. We check explicitly that there are 5 of them, but the error message will not tell you that the number is wrong if that check fails. So if there are 4 inputs for some reason, the check will fail, but the error message will tell you that "All 4 BatchNormOnnnx inputs need to be defined". So my comment was more about how the error message should tell you why the check actually failed.

hls4ml/model/optimizer/passes/batchnorm_opt.py
JanFSchulte: I really have no clue what the best configuration here would be. But I would be inclined to just keep it as is for and see if it causes problems down the road.

hls4ml/converters/onnx/merge.py
JanFSchulte: Thanks for the explanation. Sounds ok to me.

hls4ml/model/optimizer/passes/batchnorm_opt.py
jmitrevs: I see. I need to change the error message.

vloncar:
The tests included are based on QONNX model zoo which doesn't check all the features that are introduced. Can we expand the test suite somehow with synthetic models? This can be created with `onnx.helper`, ost of the ONNX test units are based on this approach. Alternatively this can be done with [`onnxscript`](https://opensource.microsoft.com/blog/2023/08/01/introducing-onnx-script-authoring-onnx-with-the-ease-of-python/). Perhaps even PyTorch to onnx and QONNX's (Q)Keras converter would suffice.

jmitrevs:
Some bugs found. Changing to draft till I fix them.

jmitrevs:
The sep_conv bug is fixed with #1069.

jmitrevs:
I added more tests from example models. Note--one is skipped because I think the qonnx channels-last conversion doesn't take into account enough activations, with https://github.com/fastmachinelearning/qonnx/pull/148 opened for it.

PR #986, opened 2024-03-28
Add namespaces and optional writer config

vloncar:
# Description

As observed in the emulator in CMSSW, there are some issues with global namespace being polluted when multiple versions of the same model are loaded. The accepted solution was to use namespaces. This PR follows up in hls4ml by adding the ability to generate namespace and optionally not write .txt files (also requested by the emulator people) and not to write .tar.gz (everyone hates this). The new configurable options are placed under `WriterConfig` as it made no sense to shove them under `HLSConfig`. In the future with the new config scheme we will re-evaluate this.

## Type of change

- [x] Documentation update

## Tests

There's a test in `test_writer_config.py` for the 3 new config options

## Checklist


jmduarte:
After discussing this with @thesps @aloeliger, we realized that the definition of `nnet_utils` may also collide (to be checked), which we want to prevent in principle due to potential differences in hls4ml versions, etc., so I think we should also try to protect those in this PR

thesps:
Could we also protect the `nnet_utils` function definitions? They are currently all in the namespace `nnet`, but we want to allow that each model/project can use a different hls4ml version/implementation, so it would be good to change that to the same namespace that applies to the weights.

vloncar:
It would be better if we worked on understanding the root of the problem rather than patching up in anticipation of unexpected behavior. Making the nnet_utils part of the namespace won't solve the issue if function loading is the problem, as the interface fundamentally depends on the existence of `create_model` and `destroy_model` functions in every loaded library and you can't put those in namesepaces. We have not seen this failing which is expected behavior according to the standard. We have also not seen bad casts which suggests types don't clash. We have only seen this in initialization of static arrays. I was unable to reproduce this and create a MWE, making me think CMSSW environment is more specific and we should understand what goes on there better. Or we just forget about all of this emulator business and go back to the solution of pasting the code of all models in CMSSW codebase if we want to be safe.

jmitrevs:
Can we fix the conflict so that we can merge this?

jmduarte:
pre-commit.ci autofix

jmduarte (approved):
Looks good. @vloncar just to check: this PR changes the default behavior to longer writing weight text files intentionally, correct?

@jmitrevs this can be merged once the jenkins test passes

hls4ml/model/graph.py
```diff
@@ -56,8 +56,8 @@ def __init__(self, config):
         else:
             self.writer_config = {
                 'Namespace': None,
-                'WriteWeightsTxt': True,
-                'WriteTar': False,
+                'WriteWeightsTxt': False,
+                'WriteTar': True,
```
vloncar, line 60: I think this should be the other way around. The tar is mostly (*) useless, the txt actually speeds up compilation (Csim) a lot so that should remain true by default. Only once you're fully ready and want a clean project to turn into an emulator in CMSSW would you change txt to false.

(*) Mostly because I rely on this feature for the sysnthesis service framework, but I can manage without.

jmduarte: Swapped here: 6a48180

jmitrevs:
There are synthesis failures that need to be investigated before we merge this.

vloncar:
There are two issues I saw. One for VivadoAccelerator writer that doesn't check if the .tar.gz exists before trying to remove it, and the second issue is that the tests rely on the existence of .tar.gz files for creating directories. Simple solution was to force the hls4ml script to always write them. The script is going away soon anyway, as soon as we switch the synthesis tests away from it.

vloncar:
The latest commit fcd6f7a from @jmduarte fixes the synthesis tests. I was just about to push the same change myself :smile:

jmitrevs:
The jenkins failure could well be because correlator2 is being updated. We should retry it once jenkins is back up.

PR #987, opened 2024-03-28
Support negative_slope in quantized_relu

vloncar:
# Description

This is a cleaned up version of #961 with a working test case added. All credit to @DrWatt


vloncar:
pre-commit.ci autofix

hls4ml/model/profiling.py
```diff
@@ -589,6 +589,7 @@ def get_ymodel_keras(keras_model, X):
         name = layer.name
         if (
             hasattr(layer, "activation")
+            and layer.activation is not None
             and layer.activation.__name__ != "linear"
```
jmitrevs, line 593: I think this line needs to move below the check of what type the layer is (i.e. swap 593 and 594)

jmitrevs (approved):

PR #991, opened 2024-04-05
Vitis accelerator

axiotisk:
# Description
The Vitis Accelerator Backend builds upon the foundation laid by the Vitis backend and streamlines the generation process for PCIe accelerators using the Vitis Accelerator Flow.
Features:
* This backend inherits from the Vitis backend, ensuring compatibility with existing workflows and projects.
* Converts the input of the top-level design from AXI Stream to memory-mapped and the output from memory-mapped to AXI Stream.
* Automates the generation of host code and the necessary makefile for kernel compilation. The host code and build process is expected to evolve before the merge of this MR

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Other (Specify)

## Tests

The backend has been tested with the hls4ml getting started tutorial example.

**Test Configuration**:
The Vitis version used for the validation is 2022.2.
The functionality of the project  was tested on a VCK5000 accelerator board.

## Checklist


hls4ml/model/graph.py
```diff
@@ -865,6 +865,14 @@ class TraceData(ctypes.Structure):
         else:
             return output, trace_output
 
+    def hardware_predict(self, x, **kwargs):
```
qberthet, line 868: This method has been added to enable performing predictions directly on the FPGA from the Python code. It feels a bit intrusive to add this backend-specific code to the hls4ml core. Another approach could be to modify `predict()` to allow backend-specific overloading. So, `model.hardware_predict(x)` could become `model.predict(x, target='hw')`, but this also requires some modification of the existing core code. Could an hls4ml dev provide advice on the best approach here? (@vloncar, @jmitrevs?). Thanks!

vloncar: I would be in favor of naming this somewhat different (`predict_hw` for example) and moving the exception to backends (in `FPGABackend` should be enough to cover those that don't/cannot support it).

Some longer term idea here would be that we would have 3 ways of doing preciction: `predict_emu` (emulation, the current one), `predict_sim` (simulation via pyverilator) and `predict_hw` (real deal), with the `predict` being `predict_emu` by default with maybe a switch for user to control which one is called if it's just `predict(x)`.

qberthet:
As testing of this PR have been mentioned in the minutes of the last dev meeting, the most recent work on the host code provided with the VitisAccelerator have been pushed to ensure testing of the latest version (Also rebased on current main).

bo3z:
I've done a first pass, going through the hls4ml-core changes. Most of the comments are minor, just to make sure the code is consistent with the rest of the hls4ml codebase. In the following days, I'll also try out the VitisAccelerator on a local set-up with a U55C / U250 and try to review the accelerator-specific (templates, build files etc.) changes. 

Overall, a very nice addition to the hls4ml codebase and seems very orthogonal to all the other functionality, so shouldn't be many issues with merging it soon.

qberthet:
Thanks for the review. There is probably some room for improvement, so please comment on your testing experience. We intend to do a polishing pass, mostly to provide a more seamless integration from the Python code, but maybe this can be done in a subsequent PR if the current PR is deemed usable enough.

bo3z:
I just tried testing the VitisAccelerator backend on Alveo u55c and Alveo u250, but there were some issues: 

- The biggest issue are timing violations: On both the u55c and u250, there is very large WNS; around -3ns to -5ns. I tried synthesising with clock periods of 4ns and 5ns; both with 27% uncertainty. Also, tried lowering the batch size to 1 (hoping it simplifies the logic and reduces congestion). Finally, I tried both with and without hw_quant. Overall, all of these cases and across boards had significant timing violations; which is a bit unexpected. To me this seems like some missing constraint in the build process or similar. I've commonly seen timing violations on the u55c around the HBM, but they are usually much smaller (-0.5ns) and can be fixed by some floor-planning and passing more advanced Vivado directives. 

- I had to change the platform for u250 to: xilinx_u250_gen3x16_xdma_4_1_202210_1, because I got the error from Vitis "Platform not found". However, a quick google of the one in this PR: xilinx_u250_xdma_201830_2, does find it. I am wondering whether there are several versions of the u250?

- On the u250; when I changed the platform, the placer in implementation failed. There was a constrain (I guess generated by hls4ml), that forces the model kernel onto SLR0; however the specific model couldn't fit into SLR0. The model was the jet tagging model, so not too large; but I think we should avoid such explicit placements of kernels to SLRs, as it can be quite hard to estimate the resource usage of a model before actual synthesis. Per-SLR placement should probably be left to more advanced users who have issues meeting timing, in my opinion.

bo3z:
So in response to the above comment: the significant timing issues are only for io_parallel; io_stream has no such issues.

qberthet:
Considering the overlap between this PR and the new VitisUnified backend PR (#1376), the developers agreed in today’s meeting to merge the two efforts so that both SoC and accelerator support are included before merging into main. As this may take some time, this PR will be periodically rebased and maintained until an equivalent solution is merged.

PR #956, opened 2024-01-12
Stage initial set of changes for the Catapult backend

dgburnette:
A# Description
This is the initial merge of a Catapult backend for HLS4ML. This merge includes all of the python changes and the top-level template files but not the nnet_utils/* files at this point in time. 

> :memo: Please include a summary of the change.
>
> * Please also include relevant motivation and context.
> * List any dependencies that are required for this change.

## Type of change

For a new feature or function, please create an issue first to discuss it
with us before submitting a pull request.

Note: Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] A new research paper code implementation
- [ ] Other (Specify)

## Tests

> :memo: Please describe the tests that you ran to verify your changes.
>
> * Provide instructions so we can reproduce.
> * Please also list any relevant details for your test configuration.

**Test Configuration**:

## Checklist

- [ ] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have installed and run `pre-commit` on the files I edited or added.
- [ ] I have added tests that prove my fix is effective or that my feature works.

PR #1000, opened 2024-04-16
fix unwanted tested file change in #956

calad0i:
Revert the change (typo) made to `test/pytest/test_cnn_mnist.py` by #956, which breaks tests when using `pytest-xdist`.


jmitrevs:
I'll pre-emptively approve this because it really looks like a typo, but we should wait for confirmation.

jmitrevs (approved):

PR #1010, opened 2024-05-03
Make im2col default option for quartus

calad0i:
A# Description

With using the Quartus backend, with conv2d with kernel size 3x3 or conv1d with kernel size 3, the framework makes the default implementation `Winograd` instad of `im2col`. As the current `Winograd` implementation does not seem to play nicely with low bitwidth and hard to achieve bit-accuracy (e.g., with automatic precision setting), I would suggest not enable it by default.

For the Winograd impl, type `uint8_t` from `<cstdint>` is used. This header is not included by default in many environments, explicit `#include <cstdint>` is added to corresponding files. 

## Type of change

- [ ] Documentation update
- [ ] Other (Specify)

## Checklist

- [ ] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).

### DIFF
--- a/hls4ml/templates/quartus/firmware/nnet_utils/nnet_conv1d_resource.h
+++ b/hls4ml/templates/quartus/firmware/nnet_utils/nnet_conv1d_resource.h
@@ -3,6 +3,7 @@
 
 #include "nnet_common.h"
 #include "nnet_dense.h"
+#include <cstdint>
 
 namespace nnet {
 
--- a/hls4ml/backends/quartus/quartus_backend.py
+++ b/hls4ml/backends/quartus/quartus_backend.py
@@ -274,7 +274,7 @@ class QuartusBackend(FPGABackend):
         # - combination - at compile-time, the decision between Winograd and im2col is made
         # - im2col - specifically use im2col
         # - Winograd - use Winograd, if possible
-        layer.set_attr('implementation', layer.model.config.get_layer_config_value(layer, 'Implementation', 'combination'))
+        layer.set_attr('implementation', layer.model.config.get_layer_config_value(layer, 'Implementation', 'im2col'))
 
         layer.set_attr(
             'n_partitions', 1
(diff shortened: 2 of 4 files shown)


jmitrevs:
I don't have a strong opinion on this. Let's see what others think.

hls4ml/templates/quartus/firmware/nnet_utils/nnet_conv1d_resource.h
```diff
@@ -3,6 +3,7 @@
 
 #include "nnet_common.h"
 #include "nnet_dense.h"
+#include <cstdint>
```
vloncar, line 6: What do we use from this header?

vloncar:
I'm fine with this type of change. But last I remember, im2col often didn't even synthesize so we added some heuristic to at least try winograd if possible. @bo3z might remember better.

bo3z:
> I'm fine with this type of change. But last I remember, im2col often didn't even synthesize so we added some heuristic to at least try winograd if possible. @bo3z might remember better.

Generally, some designs would fail HLS compilation and it just seemed that Winograd was more successful. This could be tied to a specific version of Intel HLS or some implementation detail we had at the time. If I recall correctly, the heuristic was such that if it's 3x3 (and no stride, padding etc.) we set it to use Winograd. But it's well known that Winograd doesn't perform well at extreme quantisations (which granted were never tested at the time).

hls4ml/templates/quartus/firmware/nnet_utils/nnet_conv1d_resource.h
calad0i: This header provides the `uint8_t` used in that file, and is not included by default in all platforms. Though, I am not sure if using `uint8_t` provides any performance benefit as that index should probably be optimized out when compiling to hdl...

jmitrevs (approved):

PR #1012, opened 2024-05-13
SepConv1d/2d for io_parallel with Latency strategy

vloncar:
# Description

This PR adds implementation of `SeparableConv1D/2D` for `io_parallel` and `Latency` strategy. It's based on existing code-generation for `Conv1D/2D`.

## Type of change

## Tests

Added tests in `test_sepconv1d.py` and `test_sepconv2d.py`. 1D tests had a nonsense test with a kernel size of 1, I changed that to 2 to be in line with the 2D tests. 

## Checklist

### DIFF
--- a/hls4ml/backends/fpga/passes/codegen.py
+++ b/hls4ml/backends/fpga/passes/codegen.py
@@ -1,4 +1,4 @@
-from hls4ml.model.layers import Conv1D, Conv2D
+from hls4ml.model.layers import Conv1D, Conv2D, SeparableConv1D, SeparableConv2D
 from hls4ml.model.optimizer import OptimizerPass
 from hls4ml.model.types import Source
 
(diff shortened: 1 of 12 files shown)


vloncar:
@dgburnette This could be followed up in Catapult. I noticed that you don't use the code-gen im2col implementations of conv1d/2d so I didn't want to mess with it

hls4ml/backends/fpga/passes/codegen.py
```diff
@@ -7,16 +7,27 @@ class GenerateConvIm2col(OptimizerPass):
...
+        )
 
     def transform(self, model, node):
-        node_class = node.__class__.__name__
-        if '1D' in node_class:
-            self._generate_im2col_1d(node)
-        elif '2D' in node_class:
-            self._generate_im2col_2d(node)
+        node_class = node.class_name
+        if 'Separable' in node_class:
```
jmitrevs, line 17: If we just had DepthwiseConv2D (https://keras.io/api/layers/convolution_layers/depthwise_convolution2d/) would this work? We nominally support it.

vloncar: It will work. For im2col algorithm the input matrix is the same, so DepthwiseConv1D/2D will be caught by the subsequent `elif` statement and from there on treated as the regular Conv1D/2D for creation of `fill_buffer` function. It is then correctly called in the new HLS. I've also added an explicit test for depthwise convolution

jmitrevs (approved):

PR #1008, opened 2024-04-30
fixes for depth_multiplier in SeparableConv2d and DepthwiseConv2d

dgburnette:
A# Description

The model python attached (model_asic.py) demonstrated that the shape printed by hls4ml (e.g. [None,32,32,8]) did not agree with the python model.summary() (e.g. [None,32,32,40)).
It turns out to be an incomplete handling of depth_multiplier in the SeparableConv2d and DepthwiseConv2d implementations.

## Type of change

For a new feature or function, please create an issue first to discuss it
with us before submitting a pull request.

Note: Please delete options that are not relevant.

- [x ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] A new research paper code implementation
- [ ] Other (Specify)

## Tests

import os
import shutil
import sys
import hls4ml
import yaml
import pickle
import argparse
import numpy as np
import tensorflow as tf
import json
from tensorflow import keras
from tensorflow.keras.models import model_from_json
from tensorflow.keras import regularizers
from tensorflow.keras.constraints import max_norm
from qkeras import *
from sklearn.metrics import accuracy_score
 
# Usage:
#    python3 model.py
#      will build the model, perform analysis, generate HLS C++ and then synthesize it
 
## Function to create depthwise conv2d Neural Network model. For depthwiseconv2d, the kernel size will always be (1,1)
def create_model():
    model = tf.keras.Sequential()
    model.add(DepthwiseConv2D(kernel_size=(1,1), padding='valid', depth_multiplier=5, strides=1, dilation_rate=1, use_bias=True, input_shape=(32,32,8)))
    #Compiling the model
    model.compile(loss='sparse_categorical_crossentropy',
             optimizer='adam',
             metrics=['accuracy'])
    return model
 
## Function to save model architecture, weights, and configuration
def save_model(model, name=None):
    if name is None:
        name = model.name
    model.save(name + '.h5')
    model.save_weights(name + '_weights.h5')
    with open(name + '.json', 'w') as outfile:
        outfile.write(model.to_json())
    return
 
if __name__ == '__main__':
 
    print("============================================================================================")
    print("HLS4ML Version: ",hls4ml.__version__)
 
    parser = argparse.ArgumentParser(description='Configure and convert the model for Catapult HLS')
[...]

PR #1022, opened 2024-06-13
Add an optimizer to replace SeparableConv by Depthwise + Conv (pointwise)

jmitrevs:
# Description

When looking at automatic type inference, reuse factor setting, stream buffer optimization, and eventual oneAPI implementation with task sequences, it became clear that treating separable convolutions as two layers instead of one was easier. The different layers can have different accumulator precisions, reuse factors, etc.

This optimizer converts a `SeperableConv*D` layer to a `DepthwiseConv*D` layer followed by a `Conv*D` layer for the pointwise convolution. (For backends that have an explicit pointwise implementation, a subsequent optimizer changes the `Conv*D` to `PointwiseConv*D`.) Layer-wise configurations are also created for the new depthwise and pointwise convolutions so that type inference can be done on the individual layers. Hence, this optimizer should be run before the automatic type inference. (The qonnx PR #979 adds a number of other optimizers than also need to run before the type inference, so this will be a common feature.)

~~In this PR I added parameters but did not remove any. In particular, reuse factor and accumulator type are ambiguous, and unused in the new implemenation, being split between depthwise and pointwise reuse factors and accumulators. However, if this optimizer is disabled, the old scheme should still work, with care by the user.~~

Decided it was cleaner to remove the unused parameters.

~~I believe this PR also adds support for multiplier factors other than 1, but it's untested. It was motivated by #1008 .~~

Multiplier factors other than 1 will come in a separate PR. This implements some parsing but at the end an assert is added for the Vitis/Vivado templates asserting that only a multiplier factor of 1 is currently supported.

## Type of change

Updated implementation that

- [x] Other (Specify) - more maintainable implementation

## Tests

This should not cause changes to the standard depth_multiplier=1 separable convolutions not using automatic type inference, so the default tests should be fine. The automatic type inference will be tested in a following PR that makes auto the default.

## Checklist


jmitrevs:
I am somewhat torn as to whether to remove the regular accum_t and reuse_factor, which are not used if the separable is split, which is the default.

jmitrevs:
Comment for reviewers:  Note the changes in graph.py for easier passing of the configuration from one layer to child layers after transformations. This is used also in the qonnx PR. It's probably useful to check here.

jmitrevs:
There are some template changes in the Vivado backend from when I thought I would make this PR handle multiplier factors != 1. In particular, I changed n_chan to n_filt in some places, which for multiplier_factor == 1 are identical. I also added an assert that the multiplier factor == 1. I did not put those in other backends, though all only work with multiplier factor == 1. I figured that would be taken up by the multiplier_factor != 1 PR.

JanFSchulte (approved):

PR #1037, opened 2024-07-18
FIFO depth optimizer for Vitis backend

steltze:
_FIFO depth_ optimizer for the Vitis backend using the built-in cosim FIFO profiling feature of Vitis HLS
- Inserted the new optimizer to the Vitis backend flows
- Implemented the optimizer. The steps are similar to the Vivado optimizer, only the profiling and FIFO depth parsing parts change
- Updated the _Vitis Writer_ to overwrite the `project.tcl` file with Vitis attributes
- Updated the `build_prj.tcl` file to skip the *add_vcd_instructions_tcl* function in the case of Vitis backend
- Added a pytest that checks if the depths decreased after the optimization and the cosim with the optimized depths passes, without detecting any deadlocks

## Type of change

- [x] Documentation update

## Tests

- Used Vitis HLS 2023.2
- Dummy dense model example. However, optimized depths are the same as the initial ones set by hls4ml

```python
import hls4ml
from tensorflow.keras.layers import Dense
from tensorflow.keras.models import Sequential
model = Sequential()
model.add(Dense(64, input_shape=(16,), name='fc1', activation='relu'))
model.add(Dense(32, name='fc2', activation='relu'))
model.add(Dense(32, name='fc3', activation='relu'))
model.add(Dense(5, name='fc4', activation='softmax'))

config = hls4ml.utils.config_from_keras_model(model, default_precision='ap_fixed<8, 4>')
config['Flows'] = ['vitis:fifo_depth_optimization']
hls4ml.model.optimizer.get_optimizer('vitis:fifo_depth_optimization').configure(profiling_fifo_depth=200_000)

output_dir = 'hls4mlprj_fifo_depth_opt_vitis'
hls_model = hls4ml.converters.convert_from_keras_model(model,
                                                       io_type='io_stream',
                                                       hls_config=config,
                                                       output_dir=output_dir,
                                                       part='xc7z020clg400-1',
                                                       backend='Vitis')

hls_model.build(reset=False, csim=False, synth=True, cosim=True)
hls4ml.report.read_vivado_report(output_dir)
```
- Tested the model above with Vivado backend to ensure nothing broke
- 10k param U-Net model. After running cosim with the optimized depths, no deadlocks are detected by Vitis

**Test Configuration**:

## Checklist

- [x] My changes generate no new warnings. There is a warning on `dense_resource`, " 'Resource pragma' is deprecated, use 'bind_op/bind_storage pragma' "
- [x] I have added tests that prove my fix is effective or that my feature works. **Fixed** 
~~There is a small problem with the branched model and adds a linear layer in the input causing a very large FIFO between that layer and the depthwise convolution.~~


steltze:
hey @vloncar @jmitrevs do you think a unit test is required? To avoid delaying the CI pipeline, I could add a dummy model that verifies synthesis and that the flow is executed. Otherwise, I can add a complete unit test that also checks the FIFO depth results

vloncar:
I think it would be good to have a full test, but until we improve the synthesis testing pipeline, maybe mark it with `@pytest.mark.skip(reason='Skipping synthesis tests for now')`. This way we could run it locally when we review your code and it won't cause issues in the CI.

steltze:
To the reviewers: when we agree on how we want to test this, I can open an issue to also include the Vivado/VivadoAccelerator FIFO optimization in the tests, as a future PR, as some changes in the initial implementation of those backends will be needed

steltze:
After discussing, some more todos:
- [ ] Update the hls4ml documentation by including an example of this feature with the Vitis backend
- [ ] Include a UNet model on the pytest

hls4ml/backends/vitis/passes/fifo_depth_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+    vars_to_profile = {
+        output_variable_name: output_variable
+        for output_variable_name, output_variable in model.output_vars.items()
+        if ("VivadoStreamVariable" in str(type(output_variable)))
+        and output_variable != model.get_output_variables()[0]
+        and output_variable != model.get_input_variables()[0]
+    }
+
+    # initialize all the fifos to `profiling_fifo_depth` so that they will be automatically implemented in BRAMs and so
+    # they will be profiled. Alternatively, "config_dataflow -override_user_fifo_depth profiling_fifo_depth" can be
```
nghielme, line 31: Is there any particular reason why you didn't opt for the modification of `build_prj.tcl`?

hls4ml/backends/vitis/passes/fifo_depth_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+    path_to_new_test_bench = f"{model.config.get_output_dir()}/{model.config.get_project_name()}_new_test.cpp"
+
+    newline = ""
+    second_part_of_testbench = False
+    with open(path_to_old_test_bench) as old_test_bench:
+        file_iterator = iter(old_test_bench)
+        for line in file_iterator:
+
+            if "// hls-fpga-machine-learning insert zero" in line:
+                newline += indent + indent + "const unsigned BATCH_SIZE = 2;\n"
```
nghielme, line 60: To my understanding `BATCH_SIZE` is not appropriate here, you are calling `N` times the same function for profiling purposes. `PROFILING_ITERATIONS`?

hls4ml/backends/vitis/passes/fifo_depth_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+        Returns:
+            bool: The execution state of the Optimzer Pass
+        """
+
+        # use `large_fifo_depth = 0` to keep the default fifo depth
+        # consider changing 100_000 either with a very very large value > of any total bram storage space
+        # or via vitis 2023.2 c-simulation
+        profiling_fifo_depth = getattr(self, "profiling_fifo_depth", 100_000)
+
+        if not isinstance(profiling_fifo_depth, int) or profiling_fifo_depth < 0:
```
nghielme, line 229: Is `depth=0` acceptable?

test/pytest/test_optimization/test_fifo_depth.py
```diff
@@ -0,0 +1,195 @@
...
+def test_successful_execution_of_dummy_keras(backend):
+    """Test the correct execution of the FIFO depth optimizer."""
+    run_fifo_depth_optimization_keras(backend, profiling_fifo_depth=200_000, io_type='io_stream')
+
+
+def get_tiny_unet_model():
+    """
+    Load tiny unet model, already channels-last and cleaned
+    """
+    dl_file = str(example_model_path / "onnx/branched_model_ch_last.onnx")
```
nghielme, line 148: I think this one is a model with a single branch, is not a UNet model. `tiny_unet_ch_last.onnx` should be the UNet one

nghielme (commented):
In general it looks good to me 👍

hls4ml/backends/vitis/passes/fifo_depth_optimization.py
steltze: the outcome is the same so I just implemented the easiest option, rather than changing the tcl script

test/pytest/test_optimization/test_fifo_depth.py
steltze: I chose this one since it's smaller and the test will run for a shorter period

nghielme: Ok, makes sense. Just modify the comment for consistency then 👍

steltze: done. fixed all comments

hls4ml/writer/vitis_writer.py
```diff
@@ -24,10 +24,36 @@ def write_nnet_utils_overrides(self, model):
...
+        f.write('variable project_name\n')
+        f.write(f'set project_name "{model.config.get_project_name()}"\n')
+        f.write('variable backend\n')
+        f.write('set backend "vitis"\n')
+        f.write('variable part\n')
+        f.write('set part "{}"\n'.format(model.config.get_config_value('Part')))
+        f.write('variable clock_period\n')
+        f.write('set clock_period {}\n'.format(model.config.get_config_value('ClockPeriod')))
+        f.write('variable clock_uncertainty\n')
+        f.write('set clock_uncertainty {}\n'.format(model.config.get_config_value('ClockUncertainty', '12.5%')))
```
jmitrevs, line 46: Why do we put a clock uncertainty of 12.5% here?  Also in general, would it be better to have a project.tcl template that we edit?

jmitrevs (changes requested):

hls4ml/writer/vitis_writer.py
steltze: @jmitrevs this should be the default one set at also at the vivado backend from the [writer](https://github.com/fastmachinelearning/hls4ml/blob/898d53d0d4438b0e8abb7a6aebb4bf7e6a0a8835/hls4ml/writer/vivado_writer.py#L717)

nghielme: Hi @vloncar ,
thanks for testing. I am also testing in parallel it and I noticed that `project.tcl` is generated assigning the wrong name to `project_name` variable (`myproject` vs `myproject_axi`). Have you noticed the same as well and added in your changes?

nghielme: as I wrote somewhere else in this PR, I think it would be beneficial for to have a test in which a simple model is created and synthetised with Vivado (so that all the Vivado synthesis errors can be spotted as well)

vloncar (approved):

vloncar:
This looks good and ready to be merged. One change I made (which I think is long overdue) is to ensure multiple iterations of the top function are called when user doesn't provide any input. All co-simulation results with io_stream are invalid when only one input is passed, not just the FIFO depth but latency and II. This PR originally suggested to call the top function twice, but in the past I have observed that for some ResNet blocks I needed to run co-sim 4 times (3 times to see the buffers fill up, 4th time to be sure they stopped increasing). I placed 5 to be sure. For large models this may take a lot of time, so we shouldn't go to much larger numbers as a default.

PR #1042, opened 2024-07-24
clean up mult-dimensional dense

jmitrevs:
# Description

Multidimensional Dense was broken for cases where any dimensions had size 1, and for Quartus parallel. I believe it was also broken for the resource strategy.

I made the change for `ReplaceMultidimensionalDenseWithConv` to convert `Dense` to `Conv1D` or `2D` instead of directly to `PointwiseConv1D` or `2D`, since pointwise is not required to be implemented in all cases. The conv to pointwise optimizer is backend-specific, while this is a backend-agnostic optimizer. This keeps the separation clear. Of course, the conv to pointwise optimizer usually runs after this, so in the end you still end up with a pointwise convolution (if it exists).

Also, the reshaping of the weights is now done once, in the `ReplaceMultidimensionalDenseWithConv` optimizer, instead of incorrectly there (I think it was correct originally but from before some code modernization) and also in the pointwise optimizers.

## Type of change

## Tests

The existing `test_multi_dense.py` was expanded to cover the broken cases.

## Checklist

### DIFF
--- a/test/pytest/test_multi_dense.py
+++ b/test/pytest/test_multi_dense.py
@@ -13,6 +13,8 @@ test_root_path = Path(__file__).parent
 @pytest.mark.parametrize(
     'backend, strategy',
     [
+        ('Vivado', 'Latency'),
+        ('Vivado', 'Resource'),
         ('Vitis', 'Latency'),
         ('Vitis', 'Resource'),
         ('Quartus', 'Resource'),
--- a/test/pytest/test_multi_dense.py
+++ b/test/pytest/test_multi_dense.py
@@ -11,46 +11,32 @@ test_root_path = Path(__file__).parent
 
 
 @pytest.mark.parametrize(
-    'backend, io_type',
+    'backend, strategy',
     [
-        ('Quartus', 'io_parallel'),
-        ('Vivado', 'io_parallel'),
-        ('Vitis', 'io_parallel'),
-        ('Vivado', 'io_stream'),
-        ('Vivado', 'io_stream'),
-        ('Vitis', 'io_stream'),
+        ('Vitis', 'Latency'),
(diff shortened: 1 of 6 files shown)


test/pytest/test_multi_dense.py
```diff
@@ -11,46 +11,32 @@
...
-    'backend, io_type',
+    'backend, strategy',
     [
-        ('Quartus', 'io_parallel'),
-        ('Vivado', 'io_parallel'),
-        ('Vitis', 'io_parallel'),
-        ('Vivado', 'io_stream'),
-        ('Vivado', 'io_stream'),
-        ('Vitis', 'io_stream'),
+        ('Vitis', 'Latency'),
```
JanFSchulte, line 16: We don't need to test Vivado anymore?

jmitrevs: I wonder if I got ahead of myself. I was thinking to keep the tests shorter, since Vivado support will soon be on the way out, and it shares things with Vitis, but maybe it shouldn't be just yet. The tests aren't very long. I can add it.

dgburnette (approved):

test/pytest/test_multi_dense.py
JanFSchulte: I was mostly curious if we are starting to phase out Vivado. Still catching up on the status and plans for things :) I guess it makes sense, so I don't think you necessarily need to add them.

jmitrevs: I think the plan was that after 1.0 we make Vitis the default backend and restructure the code a bit. I think it makes sense to add Vivado. (There's a similar Quartus -> oneAPI evolution that will happen.)

JanFSchulte (approved):

PR #1051, opened 2024-08-13
Add functionality to use granularity option also for pytorch models

JanFSchulte:
So far the pytorch parser's `config_from_pytorch_model` has the `granularity` argument, but complete ignores it. This PR mostly copies the treatment of this parameter from the keras parser to add this functionality. 

This requires to slightly rework in the pytorch parser to allow to extract the layer structure already in `config_from_pytorch_model`, which in turn requires to move the `input_shape` from `convert_from_pytorch_model` to `config_from_pytorch_model`. 

Removing the need to explicitly pass this shape and have it inferred from the model itself is on my todo list. 

This PR also requires adding torch to the list of requirements. 

## Type of change

## Tests

Tested on a model and using the `granularity='name'` option and adjusting the precision of a layer, this precision gets successfully propagated to the hls4ml project. 

## Checklist


JanFSchulte:
pre-commit.ci autofix

hls4ml/utils/config.py
```diff
@@ -321,6 +323,77 @@ def config_from_pytorch_model(
...
+                if name.endswith('_t'):
+                    name = name[:-2]
+                if attr.default is None:
+                    precision_cfg[name] = default_precision
+                else:
+                    precision_cfg[name] = str(attr.default)
+            else:
+                if attr.default is not None:
+                    layer_config[attr.config_name] = attr.default
+
```
jmitrevs, line 369: One bug we recently discovered in the keras parser is that the default reuse factor is not set properly. I didn't try here, but for keras if we set the backend to a real backend, the per-layer reuse factor was always set to 1. See the addition I made in #1016 to fix that. (This PR doesn't change the default to auto, but nevertheless, the reuse factor part is a bug fix, which maybe I should have extracted and not made part of #1016).

hls4ml/converters/__init__.py
```diff
@@ -303,6 +301,7 @@ def convert_from_pytorch_model(
 
     model_config = hls_config.get('Model', None)
     config['HLSConfig']['Model'] = _check_model_config(model_config)
+    config['InputShape'] = hls_config.get('InputShape', None)
```
vloncar, line 304: I think we shouldn't have a default of `None` for input shape because this will propagate further and then lead to errors which won't make it clear what is the original cause

hls4ml/utils/config.py
```diff
@@ -321,6 +323,77 @@ def config_from_pytorch_model(
     model_config['Strategy'] = 'Latency'
 
     config['Model'] = model_config
+    config['PytorchModel'] = model
+    config['InputShape'] = input_shape
```
vloncar, line 327: I would add a check if the passed input shape makes sense. Later on if it doesn't it's not easy to figure out why.

setup.cfg
```diff
@@ -32,6 +32,7 @@ install_requires =
     tabulate
     tensorflow
     tensorflow-model-optimization<=0.7.5
+    torch
```
vloncar, line 35: Do we need this? This will force all installations of hls4ml to have both tf and torch, and since these have huge dependencies themselves it may bork user's environment with the partial updates. Ideally we would have hls4ml with optional extras installed to support various frontends (e.g., `pip install hls4ml[torch]`), but this may not be easy with the current codebase.

Also, if we go down this route, we should remove the checks for existence of pytorch on the system in `converters/__init__.py`

hls4ml/utils/config.py
```diff
@@ -284,6 +285,7 @@ def config_from_pytorch_model(
 
     Args:
         model: PyTorch model
+        input_shape (list): The shape of the input tensor. First element is the batch size, needs to be None
```
vloncar, line 288: We have a nice opportunity here to get rid of the first `None` in the shape. Just ask users to not put it, but then put it in ourselves :wink:

setup.cfg
JanFSchulte: I think we can get away with not making torch a requirement by moving the `from hls4ml.converters.pytorch_to_hls import parse_pytorch_model` from the `converters/__init__.py` directly into `config.py`. The only reason why I had put the import there is that it's there for keras. The question is, is there a reason to place these imports into the `converters/__init__.py`? If not, I can just move it and get rid of the requirement.

hls4ml/utils/config.py
JanFSchulte: Thanks for the pointer. Do I understand correctly that the fix is these two lines? https://github.com/fastmachinelearning/hls4ml/pull/1016/files#diff-8ec43234900e263f95104d90f7969c22a9d314817d20ce4d1f23f1a4354fc91aR191-R192

hls4ml/converters/__init__.py
JanFSchulte: Good point, we're now raising an exception if that parameter is not found. No point in continuing.

hls4ml/utils/config.py
JanFSchulte: Good idea, that was always very ugly. Done.

hls4ml/utils/config.py
JanFSchulte: I have been thinking about that. In Pytorch it seems like the exact input shape is not determined by the model architecture, so it's not possible to completely infer it during parsing. We we can still check some general features, like the number of dimensions of the input, based on the type of the first layer. I'll implement something along those lines.

vloncar: I actually meant much simpler, just check if input shape is a list/iterable and not `None`. Because technically you can pass `None` and you'll get a strange error later. Since this is one of the top user-facing functions, it makes sense we do some user input validation here so we have more confidence later

JanFSchulte: Ah, got it. I changed it to enforce that the input shape is a tuple for a single input or a list of tuples for multiple inputs.

JanFSchulte:
The test failure is due to no space left on device for one of the pytests. Otherwise things are working without making torch a requirement for the tool :)

hls4ml/utils/config.py
jmitrevs: Yes.

hls4ml/utils/config.py
jmitrevs: In general I think we need to be careful about the batch dimensions. For ONNX it's usually just 1 (but can really be any value, e.g. 10), not None. I think we try to get rid of the batch dimension pretty quickly.

vloncar (approved):

PR #1061, opened 2024-09-09
remove padding attribute

jmitrevs:
# Description

This PR removes the "padding" attribute from hls4ml's IR. Given that it we specify the actual pad values in the IR, this is redundant, and it is also keras-specific, so it becomes unnatural to set in pytorch and onnx, since you can get paddings that do not map exactly.

## Type of change

- [x] Other (Specify):  remove keras-specific redundant feature

## Tests

This is supposed to produce no changes, so original tests should continue working. That functions as a test.

## Checklist


vloncar:
I verified that the tests run ok, they failed due to a timeout. We'll address this in a separate PR.

vloncar (approved):

PR #1067, opened 2024-09-24
Remove extras flow

vloncar:
# Description

Vivado, Quartus and Catapult backends have an "extras" flow that is used for all optimizers that are not part of existing flows and should be empty in normal operation. It is only intended as a development/debug feature and was created before the flows become so granular. Meanwhile it became a hinderance and can trip up genuine development/debug effort. This PR removes the `extras` flow and replaces it with a warning.

## Type of change
- [x] Other (Specify) - Change in internal behavior, invisible to users

## Tests

There are numerous ways to test this, as a developer one can just ensure one optimizer is not registered to a backend flow to see the warning. This can be done by commenting out any of the existing optimizers in any backend flows or creating a new optimizer in `passes` directory.

Creating a test-case for this is not really useful and is very hacky, as it requires "reloading" a backend, first by removing the registered optimizers and flows, then registering the new dummy optmizer and finally reloading the backend in the `backend_map`. I suggest we don't do this.

## Checklist

### DIFF
--- a/hls4ml/backends/catapult/catapult_backend.py
+++ b/hls4ml/backends/catapult/catapult_backend.py
@@ -153,9 +153,8 @@ class CatapultBackend(FPGABackend):
         ]
 
         if len(extras) > 0:
-            extras_flow = register_flow('extras', extras, requires=[init_flow], backend=self.name)
-        else:
-            extras_flow = None
+            for opt in extras:
+                print(f'WARNING: Optimizer "{opt}" is not part of any flow and will not be executed.')
 
         ip_flow_requirements = [
             'optimize',
@@ -164,10 +163,8 @@ class CatapultBackend(FPGABackend):
             quantization_flow,
             optimization_flow,
             catapult_types_flow,
-            extras_flow,
             template_flow,
         ]
-        ip_flow_requirements = list(filter(None, ip_flow_requirements))
 
         self._default_flow = register_flow('ip', None, requires=ip_flow_requirements, backend=self.name)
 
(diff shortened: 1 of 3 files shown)


hls4ml/backends/catapult/catapult_backend.py
```diff
@@ -153,9 +153,8 @@ def _register_flows(self):
         ]
 
         if len(extras) > 0:
-            extras_flow = register_flow('extras', extras, requires=[init_flow], backend=self.name)
-        else:
-            extras_flow = None
+            for opt in extras:
+                print(f'WARNING: Optimizer "{opt}" is not part of any flow and will not be executed.')
```
jmitrevs, line 157: Is it better to print a warning or to do warnings.warn()? I am fine either way.

jmitrevs (approved):

hls4ml/backends/catapult/catapult_backend.py
vloncar: There's a nice description about `warnings` vs `logging` in python documentation:
> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
> A logger’s [warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning) method if there is nothing the client application can do about the situation, but the event should still be noted

By this logic we should use `warnings.warn()` instead of our way of "logging" with `print()`. I'll update the PR.

On a related note, I was thinking that after we merge large PRs for 1.0 we do a simple pass and change all `print()` statements to `logging.info()` or `logging.warn()` and eventually use `logging.debug()` throughout the code to help with development/debugging. It was never a good time to do this as it affected large unmerged codebases, but it seems after 1.0 things will finally sufficiently quiet down for this to be done.

PR #1068, opened 2024-09-29
Raise exception on compile errors

vloncar:
# Description

In some cases when the compilation fails, the `build_lib.sh` script still returns 0 and the actual error is "lost" (as in, not visible in all environments, like Jupyter). We can force the script to fail on any error with `set -e` command, as done in first commit. The other two commits are to improve the code path that is touched by this. `build_lib.sh` should be executable and this prompted a change in writer. While there, I updated the code to use the `pathlib` library instead of os.path for a cleaner code. We should do a separate PR to change the rest of the writer (also in other backends eventually). Finally there's a change to use `subprocess` instead of `os.system` since this is the recommended way for newer Python.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) - sort of.

## Tests

Existing tests are enough, they all call `compile()` in some way.

## Checklist

### DIFF
--- a/hls4ml/backends/fpga/fpga_backend.py
+++ b/hls4ml/backends/fpga/fpga_backend.py
@@ -1,6 +1,7 @@
 import math
 import os
 import re
+import subprocess
 from bisect import bisect_left
 from collections.abc import Iterable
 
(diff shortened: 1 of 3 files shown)


vloncar:
pre-commit.ci autofix

vloncar:
pre-commit.ci autofix

jmitrevs:
`test_sr.py` fails. Does the symbolic backend need some updates?

vloncar:
I keep forgetting the SR backend exists :smile:  Should be working for all backends now.

jmitrevs (approved):

PR #1069, opened 2024-09-29
Expose alpha and theta type for parametrized activations

jmitrevs:
# Description

The type of the parameter for parametrized activations were not configurable before, being set the same as either the input of result. Generally there is no reason for that requirement, so this makes the type explicitly configurable,

## Type of change

## Tests

The standard activations tests (slightly modified) provide the test.

## Checklist

### DIFF
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -84,6 +84,9 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
         if node_class in ['SimpleRNN', 'LSTM', 'GRU']:
             return self._infer_rnn_precision(node, types_to_infer)
 
+        if node_class in ['ParametrizedActivation']:
+            return self._infer_par_act_precision(node, types_to_infer)
+
         # What about quantized activation layer? Setting it to 'auto' manually will break it here. We should prevent
         # this in config_from_* functions
 
@@ -555,5 +558,16 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
                 self._infer_default_type(node, f'{weightvar}_t')
                 node.weights[weightvar].update_precision(node.types[f'{weightvar}_t'].precision)
                 inferred_types.append(f'{weightvar}_t')
+
+        return inferred_types
+
+    def _infer_par_act_precision(self, node, types_to_infer):
+        inferred_types = []
+
+        # for now, only set if for threshold relu
+        if 'param_t' in inferred_types and self.get_attr('activation').lower() == 'thresholdedrelu':
+            in_type = node.get_input_variable().type.precision
+            node.attributes['param_t'].type = in_type
+            inferred_types.append('param_t')
 
         return inferred_types
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -564,7 +564,9 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
     def _infer_par_act_precision(self, node, types_to_infer):
         inferred_types = []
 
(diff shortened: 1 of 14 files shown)


jmitrevs:
Note:  this PR is needed to fix test_qonnx.py::test_sep_conv[Vitis]  in #979.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -557,3 +560,14 @@ def _infer_rnn_precision(self, node, types_to_infer):
                 inferred_types.append(f'{weightvar}_t')
 
         return inferred_types
+
+    def _infer_par_act_precision(self, node, types_to_infer):
+        inferred_types = []
+
+        # for now, only set if for threshold relu
```
vloncar, line 567: Why not ELU while we're at it?

example-models
vloncar, line 1: Technically, this got in by accident from the QONNX PR

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: I agree with that, my comment was mostly that the comment in the code suggests this is only applicable to thresholded relu while it actually works for three layers (leaky, threholded and elu)

example-models
vloncar: I think it is fine to remain. We'll update that pointer anyway.

hls4ml/model/optimizer/passes/infer_precision.py
vloncar: I think that the idea of configurable threshold is good in priciple. However I think this way we risk of having an inconsistency between parametrized activations. Thresholded relu gets to have the same type as input, which is the current behavior anyway, but leaky and elu get the default (which may be different from what is the input to the current layer), and prelu is not even considered so it gets the default that way. Why not handle all of them to have a consistent behavior (whatever that is, may be input precision for continuity, may be default precision)? Also, why use input precision when we can be smart and figure out the required precision given that you have access to the `activ_param` to figure out exactly how many bits you need to represent that number/array.

jmitrevs: They have a fundamentally different function. Threshold compares the input values to a threshold, so it makes sense for it to be the same as the input. One can tune it if preferred, but there is an implicit connection with the input type. The others are scaling factors, so they have no connection to the input type at all. It just doesn't make sense to make them the same as the input type. There is a fundamental difference. I did consider adding prelu to the match on line 87, but then `_infer_par_act_precision` would just ignore it and have it set the default anyway, so I didn't add it to the match. But I am not fundamentally opposed to that.

jmitrevs: I think logically, for comparisons the same type makes sense, so it seems like a good setting, I believe better than the default. For scaling factors, we have no guidance, so the default precision makes sense.

jmitrevs: But the suggestion about looking at the `activ_param` is interesting. I have to think about it. One can set the range, but not necessarily the number of bits.

vloncar: For a scale (and threshold) that is a scalar the guidance is simple, it's a number for which we can figure out the best precision to store (without using too many decimal bits to match the original stored in float). Does get trickier for the case of PReLU where we would need a smarter way to decide what precision suits best for the most of the array. I undestand your logic for the current behavior. I'm fine with merging this as-is and do a "smarter" way as a follow-up

vloncar (approved):

hls4ml/model/optimizer/passes/infer_precision.py
jmitrevs: Actually, I think all activations can be updated. The output type can be unsigned for relu, restricted in range for sigmoid and tanh.

Issue #859, opened 2023-08-29
Add Resource, io_parallel, depth_multiplier>1 and Quartus support for (Q)DepthwiseConv


jmitrevs:
## Details
As discussed in #834, (Q)DepthwiseConv2D and DepthwiseConv1D only exist for `io_stream` and `Latency` strategy for Vivado and Vitis. Also only `depth_multiplier=1` is supported.

### New behavior

We should add support for `Resource` strategy, `io_parallel`, `depth_multiplier>1` and Quartus support for (Q)DepthwiseConv.

### Motivation
This will provide better support for depthwise convolution.

### Parts of hls4ml being affected
The separable and depthwise convolution implementations.

steltze:
hi @jmitrevs @vloncar I am interested in the `Resource` strategy for DepthwiseConv2D as I want to use it for Separable convolution layers. Anything in mind on how to implement it?

vloncar:
You would need to implement a `depthwise_product(...)` for resource strategy in `nnet_sepconv_stream.h` that currently is hardcoded to fail an assertion. This product is not the same as matrix-vector multiplication that is at the core of most layers, rather, it's a simpler elementwise product.

The current, latency-oriented implementation of `depthwise_product` was based on `dense_latency` so you can compare the two to see how it was changed. You would essentially have to do the same to produce an elementwise product that has a similar flow as `dense_resource`. This should be simpler than current `dense_resource` implementation which has three possible scenarios that may not be needed here.

PR #1079, opened 2024-10-11
Depthwise 1D and 2D Resource strategy for io_stream

steltze:
# Description

- Related to #859
- Added `const` to the depth variable of the FIFOs between the Depthwise and the Pointwise convolutions to fix a Vitis HLS (2023.2) c-synthesis warning

Similar to the dense resource implementation, there are 3 separate implementations depending on the *reuse factor* and *number of channels*

The common functionality is similar to the dense_resource layer
- calculate the `block_factor`
- initialize the accumulator array with the biases

- `in_index`: data-weight index, always incremented by the `reuse factor` 
- `out_index`: output index

1. `rf < n_chan`

the `out_index` follows the `in_index` and there is just a check for `out_index` if it's out of bounds

2. `rf >= n_chan` and `rf % n_chan == 0`

the `out_index` does not change in the `block_factor` loop, it's only incremented in the `rufactor` loop. It should be the least resource-demanding implementation

3. `rf > n_chan`

the `out_index` is incremented with the remainder `rf % n_chan` and there is just a check if `out_index` is out of bounds

## Type of change

## Tests

- Apart from the numerical correctness in the pytests comparing Keras to hls4ml, executed successfully the c-synthesis
- Checked the numerical correctness and c-synthesis of a 10k parameter UNet that has 20 Separable 2D Convolutions

## Checklist


steltze:
Some results

jmitrevs:
Thanks. I should mention that there is also an effort to develop depthwise (and seperable, which gets split) for oneAPI, which will replace the Quartus backend eventually. It's in a PR to the oneAPI PR.

test/pytest/test_depthconv1d.py
```diff
@@ -21,18 +24,22 @@
...
-        ('Vivado', 'io_stream'),
-        ('Vitis', 'io_stream'),
-        ('Catapult', 'io_stream'),
+        # ('Vivado', 'io_parallel', 'latency'),
+        # ('Vitis', 'io_parallel', 'latency'),
+        # ('Vivado', 'io_stream', 'latency'),
+        # ('Vitis', 'io_stream', 'latency'),
+        ('Vivado', 'io_stream', 'resource'),
+        ('Vitis', 'io_stream', 'resource'),
+        # ('Catapult', 'io_stream', 'latency'),
```
jmitrevs, line 35: This is a debugging setup that got committed. We should uncomment the useful checkps

test/pytest/test_depthconv2d.py
```diff
@@ -22,18 +24,22 @@
...
-        ('Vivado', 'io_stream'),
-        ('Vitis', 'io_stream'),
-        ('Catapult', 'io_stream'),
+        # ('Vivado', 'io_parallel', 'latency'),
+        # ('Vitis', 'io_parallel', 'latency'),
+        # ('Vivado', 'io_stream', 'latency'),
+        # ('Vitis', 'io_stream', 'latency'),
+        ('Vivado', 'io_stream', 'resource'),
+        ('Vitis', 'io_stream', 'resource'),
+        # ('Catapult', 'io_stream', 'latency'),
```
jmitrevs, line 35: Same issue here. The others should be uncommented.

jmitrevs:
Updated to the latest main branch since I think it fixes the issue that caused the test failure.

steltze:
Unrolling over different `out_index`es (internal loop) might produce better QoR. Pending to test this hypothesis

nghielme:
What is blocking this PR to be merged?

xtreme8000:
With this patch my Vivado HLS 2019.2 fails with following output, regardless of resource or latency mode:
```sh
ERROR: [HLS 200-70] '#pragma HLS ARRAY_PARTITION variable=&acc type=complete' is not a valid pragma.
ERROR: [HLS 200-70] '#pragma HLS ARRAY_RESHAPE variable=&weights type=block factor=block_factor' is not a valid pragma.
ERROR: [HLS 200-70] '#pragma HLS ARRAY_RESHAPE variable=&data type=block factor=block_factor' is not a valid pragma.
ERROR: [HLS 200-70] '#pragma HLS ARRAY_PARTITION variable=&acc type=complete' is not a valid pragma.
ERROR: [HLS 200-70] '#pragma HLS ARRAY_PARTITION variable=&acc type=complete' is not a valid pragma.
```

The issue can be solved by removing `type=` from these pragmas as done elsewhere in the code base.

test/pytest/test_depthconv1d.py
steltze: @jmitrevs because of the addition of the reuse factor and input size options, the tests increase exponentially. I could include those options only to the `resource` strategy tests to reduce the total number of tests but it won't be easy to maintain. Any ideas?

steltze:
thanks @xtreme8000! fixed

test/pytest/test_depthconv1d.py
JanFSchulte: Do we need all these reuse factor and input size options, i.e. how likely is it that the code would break in different ways for the different cases? I think we will probably need to live with testing only 1 or 2 or these combinations every time, otherwise this would get quite out of hand.

steltze: Reduced the number of tests to the minimum in order to cover all the cases of the resource implementation. Tests won't probably break unless something changes in the `rf` validation logic. When I first started the implementation, `sepconv` layer had a single `rf` which caused problems with `dense-resource`, now that `sepconv` is split, it works ok.

There was also an idea only to use `rf`s that trigger the 2nd case (rem0), instead of the 3rd one for better QoR.

hls4ml/templates/vitis/nnet_utils/nnet_sepconv1d_stream.h
```diff
@@ -86,7 +86,7 @@ void separable_conv_1d_cl(hls::stream<data_T> &data, hls::stream<res_T> &res,
     #pragma HLS DATAFLOW
 
     hls::stream<dw_res_T> depthwise_res;
-    unsigned res_depth = CONFIG_T::depthwise_config::out_width;
+    const unsigned res_depth = CONFIG_T::depthwise_config::out_width;
```
vloncar, line 89: Does using `const` here bring any benefit? And since this is a newer compiler, why not `constexpr`?

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
```diff
@@ -4,13 +4,219 @@
...
+                                            typename CONFIG_T::bias_t biases[CONFIG_T::n_chan]) {
+
+    const int nin = CONFIG_T::kernel_size * CONFIG_T::n_chan;
+    const int nout = CONFIG_T::n_chan;
+    const int rufactor = CONFIG_T::reuse_factor;
+    const int multfactor = MIN(nin, rufactor);
+    const int multiplier_limit = DIV_ROUNDUP(nin, multfactor);
+    const int block_factor = DIV_ROUNDUP(nin, rufactor);
+
+    assert((multiplier_limit == block_factor) && "This function is correct only for RF <= N_IN");
```
vloncar, line 22: Would be great if this message made it more clear what `N_IN` is. Same for `N_OUT` in others. We should update all functions to make it more clear, not just `resource` ones.

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
```diff
@@ -4,13 +4,219 @@
...
+                                            typename CONFIG_T::bias_t biases[CONFIG_T::n_chan]) {
+
+    const int nin = CONFIG_T::kernel_size * CONFIG_T::n_chan;
+    const int nout = CONFIG_T::n_chan;
+    const int rufactor = MIN(CONFIG_T::reuse_factor, nin);
+    const int block_factor = DIV_ROUNDUP(nin, rufactor);
+    assert((rufactor > nout) && "This function is correct only for RF > N_IN");
+
+    #pragma HLS function_instantiate variable=weights,biases
+    //#pragma HLS RESOURCE variable=weights core=RAM_2P_BRAM Commenting out the deisgnation HLS seems to choose correctly
```
vloncar, line 156: We should get rid of these comments across the file.

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
```diff
@@ -4,13 +4,219 @@
...
 #include "nnet_common.h"
 #include "nnet_conv_stream.h"
-
 namespace nnet {
 
 template <class data_T, class res_T, typename CONFIG_T>
-void depthwise_product(data_T data[CONFIG_T::kernel_size * CONFIG_T::n_chan], res_T res[CONFIG_T::n_chan],
-                       typename CONFIG_T::weight_t weights[CONFIG_T::kernel_size * CONFIG_T::n_chan],
-                       typename CONFIG_T::bias_t biases[CONFIG_T::n_chan]) {
+void depthwise_product_resource_rf_lt_nchan(data_T data[CONFIG_T::kernel_size * CONFIG_T::n_chan],
```
vloncar, line 10: For the sake of making this file clearer, I propose we move all depthwise product functions into a separate file, `nnet_depthwise_product.h`. Don't forget that you'll have to then add this file in sepconv include list in `convolution_templates.py`

test/pytest/test_depthconv1d.py
```diff
@@ -13,6 +13,8 @@
 strides_options = [(1), (2)]
 kernel_options = [(2), (3)]
 bias_options = [False]
+rf_options = [1, 16, 24]  # each rf corresponds to one of the three cases of depthwise resource for io_stream
+input_size_options = [16]
```
vloncar, line 17: Why 16? This only makes the test run longer. Can we choose a smaller number?

test/pytest/test_depthconv2d.py
```diff
@@ -8,12 +8,13 @@
 
 test_root_path = Path(__file__).parent
 
-padds_options = ['same', 'valid']
+padds_options = ['same']
```
vloncar, line 11: Why remove `valid`?

vloncar (dismissed):
Looks very good so far, needs minor changes.

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
steltze: Should I change the non-resource functions in this PR or open a separate one?

hls4ml/templates/vivado/nnet_utils/nnet_sepconv_stream.h
steltze: Done. I think changing `convolution_templates.py` is not needed?

test/pytest/test_depthconv1d.py
steltze: Reduced to `3` and `4` respectively in order to make sure all 3 cases of the implementation are triggered

vloncar:
The only thing I see lacking here is the use of function pointers for depnse instead of the current `if (strategy == latency) do_latency_dense() else do_resource_dense()` approach. We moved the rest of the codebase towards that model as it allows the dense function to be generated and implemented in special ways (HGQ and `resource_unrolled` use this). I've made the changes needded and put them in [vloncar/stelios_depthwise_resource](https://github.com/vloncar/hls4ml/tree/stelios_depthwise_resource). 

After extensive testing of both approaches, I've made a few observations:
- In some rare instances function pointer adds a clock cycle to the main loop of depthwise function. I didn't manage to figure out why, the schedule viewer shows both approaches resulting in identical sequence of operations, but with function pointer one having one operation after `depthwise_product` function scheduled in the next cycle for no apparent reason. I'm thinking to follow this up with AMD. 
- Sometimes the 3rd case (`rf > n_chan and rf % n_chan != 0`) fails to meet timing. This is regardless of the approach used. We could consider enforcing RF that avoids it
- Depending on the size of the layer/input, the `pragma HLS INLINE recursive` just prior to calling `depthwise_resource` may have a detrimental effect (going from added cycles to missing timing). Again, regardless of the approach. FWIW, this also affects regular Conv1D/2D.

Since the issues are specific to a use case, I'm thinking to merge the function pointer changes into this and then merge it to main. (there's also a trivial change to tests to avoid creating directory within a directory.)  If the issue appears, it is trivial to manually tweak around it to avoid the use of function pointer and/or tweak the inline pragma. The 3rd case can also be avoided by the user if they observe that issue. If no one objects I'll do this tomorrow.

vloncar:
This is now ready. Let's just wait to see if I messed up the tests before we merge.

vloncar (approved):

Issue #1076, opened 2024-10-08
bug in Leaky_relu instantiation


AnouarITI:
I have a model that includes a leaky_relu layer. The leaky relu function requires three input parameters (input, alpha, output) however, during instantiation, it only generates a function with two input parameters (input, output).

This is the Leaky Relu function in activation.h:
```
template <class data_T, class param_T, class res_T, typename CONFIG_T>
void leaky_relu(data_T data[CONFIG_T::n_in], param_T alpha, res_T res[CONFIG_T::n_in]) {
    #pragma HLS PIPELINE

    data_T datareg;
    for (int ii = 0; ii < CONFIG_T::n_in; ii++) {
        datareg = data[ii];
        if (datareg > 0)
            res[ii] = datareg;
        else
            res[ii] = alpha * datareg;
    }
}
```
in myproject() function Leaky relu is instantiated as follow:
`nnet::leaky_relu<layer12_t, layer15_t, leaky_relu_config15>(layer12_out, layer15_out)`

Could this be fixed please?

JanFSchulte:
Hi, 

can you please be more specific about your setup and model? Which version of hls4ml are you using? Which io_type and backend are you trying to use? 
I tried to replicate this issue but for me it seems to be working correctly with the current state of the master branch.

bo3z:
There have been some changes in hls4ml on the data types of activations. Can you please confirm you have the latest version of hls4ml. You can either install it via pip or clone it from github and install it using `pip install -e .` . I suspect what's going on here is that you have some mismatch of files from different versions.

JanFSchulte:
Version 0.6.0 is certainly a bit old. As I said, it seems to be working fine in the current master, so I would indeed encourage you to test with either the master or install version 0.8. via pip.

zsrabbani:
I have the same problem. I already have the updated version of 0.8.1 but still have the same error.

jmitrevs:
Can you give an example that we can run that shows the problem?

zsrabbani:
**### Model:**
def resnet_block(input_data, filters, conv_size):
    shortcut = input_data  # Store the original input for the shortcut connection

    # Pre-activation (Batch Normalization and Activation before Convolution)
    x = BatchNormalization()(input_data)
    x = Activation('leaky_relu')(x)

    # Bottleneck 1x1 Convolution (reduce channels)
[...]

PR #1085, opened 2024-10-20
update keras activation parsing, especially leaky relu

jmitrevs:
# Description

Issue #1076 showed that hls4ml didn't parse leaky_relu passed as a parameter (`activation='leaky_relu'`) in Keras properly. One could argue that it was not explicitly described in v2 of the API documentation, but it was in the code, and it is officially described in v3 of the API, so we should support it.

As an aside, I don't think we support relu with an alpha parameter properly--maybe I should add a check to change relu to leaky_relu in that case.

I noticed also that the `Activation` layer handler was not called for activations passed as arguments. Special cases had to be handled in two places, once in the Activation layer handler, once in the activations passed as arguments section. This I believe is error-prone, and elu passed as an argument would break the oneAPI backend if I don't duplicate code. I think it is better to unify and always call the Activation layer handler, so that is what I did instead.

## Type of change

## Tests

I added a leaky_relu passed as argument test, and it's also important not to break the other tests.

## Checklist

### DIFF
--- a/hls4ml/converters/keras_to_hls.py
+++ b/hls4ml/converters/keras_to_hls.py
@@ -297,26 +297,18 @@ def parse_keras_model(model_arch, reader):
         layer_list.append(layer)
         if 'activation' in layer and layer['class_name'] not in activation_layers + recurrent_layers:  # + qkeras_layers:
             act_layer = {}
+            act_details = layer['activation']
             # Workaround for QKeras activations passed as an argument
-            if isinstance(layer['activation'], dict):
-                act_details = layer['activation']
+            if isinstance(act_details, dict):
                 act_layer['class_name'] = 'QActivation'
                 act_layer['config'] = {
                     'name': layer['name'] + '_' + act_details['class_name'],
                     'activation': act_details,
                 }
-                act_layer, output_shape = layer_handlers['QActivation'](act_layer, None, [output_shape], reader)
             else:
-                act_layer['name'] = layer['name'] + '_' + layer['activation']
-                act_layer['activation'] = layer['activation']
-                if 'activ_param' in layer:
-                    act_layer['activ_param'] = layer['activ_param']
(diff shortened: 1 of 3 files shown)


jmitrevs:
I also added support for the new v3 API parameter names for leaky relu.

hls4ml/converters/keras_to_hls.py
```diff
@@ -297,26 +297,18 @@ def parse_keras_model(model_arch, reader):
         layer_list.append(layer)
         if 'activation' in layer and layer['class_name'] not in activation_layers + recurrent_layers:  # + qkeras_layers:
             act_layer = {}
+            act_details = layer['activation']
             # Workaround for QKeras activations passed as an argument
-            if isinstance(layer['activation'], dict):
-                act_details = layer['activation']
+            if isinstance(act_details, dict):
```
jmitrevs, line 302: The logic here is unchanged from before, but is it really valid to say that if the activation is a dictionary, then it is QActivation? Yes, it seems to be used there more often, but can't I pass a leaky relu with a parameter somehow as a dictionary?

jmitrevs:
I want to understand the need for the assert change, so putting it to draft for now.

JanFSchulte (approved):

hls4ml/converters/keras_to_hls.py
jmitrevs: I was wondering if it was possible to explicitly pass the activation functions (and not text), and if that made a dictionary in this case. But this logic is unchanged from before in this PR. I can try to briefly test it.

jmitrevs: Actually, it does seem to need to be text:  https://github.com/keras-team/tf-keras/blob/v2.17.0/tf_keras/layers/core/dense.py#L125

bo3z (approved):

vloncar (approved):

Issue #1074, opened 2024-10-07
Compiling PyTorch LSTM with Hidden State Input


vishal-chandra:
#### Summary
Getting a C++ redefinition error on hidden state input when compiling a pytorch model with an LSTM error

#### Details
I have a model that contains an LSTM layer with the hidden state to that layer passed into `model.forward` some of the time (default None, same as the torch LSTM layer definition). When I remove this parameter of `model.forward`, the model compiles fine. However, with it included, the hidden state input variable is shown fine in the ModelGraph inputs, but is defined twice in the resulting C++ code, leading to a compilation error.

#### Relevant Code Snippet
<code>def forward(self, ..., hidden): 
     ...
     lstm_out, hidden_out = self.lstm(lstm_input, hidden)
     ...
</code>

#### Resulting Error

> Writing HLS project
> Done
> In file included from firmware/myproject.cpp:3:
> firmware/myproject.h:14:14: error: redefinition of parameter 'hidden'
>     input3_t hidden[N_INPUT_1_3*N_INPUT_2_3*N_INPUT_3_3], result_t layer11_out[N_LAYER_11]
>              ^
> firmware/myproject.h:13:102: note: previous declaration is here

#### Steps to Reproduce
Working on hls4ml 0.9.0.dev65+gafed23b1 and torch 2.6.0. Create a simple model with an LSTM layer and the hidden state as a parameter to the forward function and compile the model.

### Expected behavior
Hidden is listed only once in ModelGraph.inputs and is defined only once in the C++ code.

#### Actual behavior
Hidden is listed only once in ModelGraph.inputs but is defined twice in the C++ code, leading to a compilation error. I also notice that in the printed network topology, the LSTM layer is listed as having only one input.

JanFSchulte:
Hi! 

Thanks for using hls4ml and reporting this issue. 

We actually currently do not support passing of non-zero tensors for the initial hidden state in recurrent layers in Pytorch (see the description of the PR that introduced them https://github.com/fastmachinelearning/hls4ml/pull/850). Unfortunately, we have not really documented this limitation so far, that needs to be improved. Of course, this feature should also be properly supported, but so far we haven't quite managed to do it. 

[...]

Issue #1083, opened 2024-10-16
PyTorch: Unable to Compile Softmax or Sigmoid Layers Correctly


sei-rquartiano:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [X] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [X] Check that the issue hasn't already been reported, by checking the currently open issues.
- [X] If there are steps to reproduce the problem, make sure to write them down below.

# Quick summary
Unable to compile a PyTorch softmax or sigmoid layer. Attempted both module and functional calls; got different error messages described in next section

# Details
For the nn.Softmax() call I get a `KeyError: 'axis'` error which stops config generation. For nn.Sigmoid() and functional softmax or sigmoid calls it gets through compilation, however an `AssertionError` shows that the values are out of tolerance. More importantly than the tolerance, they appear to output the same incorrect values every time. 

### Steps to Reproduce
This is the code I've used to investigate errors with pytorch softmax/sigmoid calls. To switch between them comment/uncomment the appropriate lines in the model definition towards the top of the file. I also added a linear layer as an example of something that would normally precede a softmax. Doesn't seem to affect the error.

1. Clone hls4ml repository at current master branch
2. Run this test code
3. Change model definition to a different softmax/sigmoid call, repeat step 2

```
from pathlib import Path

import numpy as np
import pytest
import tensorflow as tf
from sklearn.metrics import accuracy_score
import torch.nn as nn
import torch
#from torchsummary import summary

import hls4ml
import os

from hls4ml.converters import convert_from_pytorch_model
from hls4ml.utils.config import config_from_pytorch_model

test_root_path = Path(__file__).parent

class SoftmaxModel(nn.Module):
    def __init__(self):
        super().__init__()
        #self.linear = nn.Linear(10,4)
        self.softmax = nn.Softmax(dim=-1)
        #self.softmax = nn.Sigmoid()
    def forward(self, x):
        #x = self.linear(x)
        return self.softmax(x)
        #return nn.functional.softmax(x,dim=-1)
        #return nn.functional.sigmoid(x)

if __name__ == "__main__":

    n_in = 2
    size_in = 10
    n_batch = 3

    model = SoftmaxModel()
    model.eval()
    print(model)

    X_input = np.random.rand(n_batch, n_in, size_in)
    with torch.no_grad():
        pytorch_prediction = model(torch.Tensor(X_input))

[...]

PR #1086, opened 2024-10-22
Fix softmax parsing in pytorch and add test

JanFSchulte:
Fixes issues with softmax activation in pytorch parser reported in https://github.com/fastmachinelearning/hls4ml/issues/1083 and adds softmax activation to the pytest. 

I also added a one-line change to fix https://github.com/fastmachinelearning/hls4ml/issues/1074 while I'm messing with the pytorch parser. 

## Type of change

## Tests

Added softmax to the pytest, where it works now. 

**Test Configuration**:

## Checklist


hls4ml/converters/pytorch/core.py
```diff
@@ -62,9 +62,13 @@ def parse_activation_layer(operation, layer_name, input_names, input_shapes, nod
             layer['activation'] = 'ThresholdedReLU'
             if layer['activ_param'] < 0:
                 raise Exception('negative threshold values not supported')
-
-        if hasattr(node, 'dim'):
+        if hasattr(class_object, 'dim'):
             layer['axis'] = class_object.dim
+            if layer['class_name'] == 'Softmax' and layer['axis'] is None:
+                layer['axis'] = -1
+            if 'IOType' in config:
```
vloncar, line 69: We have done these types of checks before, but now I think we should parse everything into the IR, and then backbends should check if the feature is supported. We can have optimizers for that. Check out feature_check.py in Vitis backend. We could add similar checks for all backends and move away from the parser (in a future PR)

jmitrevs (approved):

PR #1093, opened 2024-10-26
rm np_config.enable_numpy_behavior()

### LINKED ISSUE #1092: hls4ml optimization/dsp_aware_pruning changing tensorflow global behavior
A couple of the current tests fails if `hls4ml.optimization.dsp_aware_pruning` is imported prior to those tests. The code causing the issue appears to be
https://github.com/fastmachinelearning/hls4ml/blob/ab457084daf2956b5beae806b20729fd9103d8ad/hls4ml/optimization/dsp_aware_pruning/keras/__init__.py#L18
Is there anyway we can avoid using this?

calad0i:
# Description

Fixes #1092, while trying to keep the original printing behavior.

### DIFF
--- a/hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
+++ b/hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
@@ -4,9 +4,6 @@ import time
 import numpy as np
 import tensorflow as tf
 
-# Enables printing of loss tensors during custom training loop
-from tensorflow.python.ops.numpy_ops import np_config
-
 import hls4ml.optimization.dsp_aware_pruning.keras.utils as utils
 from hls4ml.optimization.dsp_aware_pruning.config import SUPPORTED_STRUCTURES
 from hls4ml.optimization.dsp_aware_pruning.keras.builder import build_optimizable_model, remove_custom_regularizers
@@ -15,7 +12,6 @@ from hls4ml.optimization.dsp_aware_pruning.keras.masking import get_model_masks
 from hls4ml.optimization.dsp_aware_pruning.keras.reduction import reduce_model
 from hls4ml.optimization.dsp_aware_pruning.scheduler import OptimizationScheduler
 
-np_config.enable_numpy_behavior()
 default_regularization_range = np.logspace(-6, -2, num=16).tolist()
 
 
@@ -121,7 +117,7 @@ def optimize_model(
     model.compile(optimizer, loss_fn, metrics=[validation_metric])
     baseline_performance = model.evaluate(validation_dataset, verbose=0, return_dict=False)[-1]
     if verbose:
-        print(f'Baseline performance on validation set: {baseline_performance}')
+        tf.print(f'Baseline performance on validation set: {baseline_performance}')
 
     # Save best weights
     # Always save weights to a file, to reduce memory utilization
@@ -222,7 +218,7 @@ def optimize_model(
 
         # Train model with weight freezing [pruning]
         if verbose:
-            print(f'Pruning with a target sparsity of {target_sparsity * 100.0}% [relative to objective]')
+            tf.print(f'Pruning with a target sparsity of {target_sparsity * 100.0}% [relative to objective]')
         for epoch in range(epochs - rewinding_epochs):
             start_time = time.time()
             epoch_loss_avg = tf.keras.metrics.Mean()
@@ -237,14 +233,14 @@ def optimize_model(
                 val_res = optimizable_model.evaluate(validation_dataset, verbose=0, return_dict=False)
                 t = time.time() - start_time
                 avg_loss = round(epoch_loss_avg.result(), 3)
-                print(f'Epoch: {epoch + 1} - Time: {t}s - Average training loss: {avg_loss}')
-                print(f'Epoch: {epoch + 1} - learning_rate: {optimizable_model.optimizer.learning_rate.numpy()}')
-                print(f'Epoch: {epoch + 1} - Validation loss: {val_res[0]} - Performance on validation set: {val_res[1]}')
+                tf.print(f'Epoch: {epoch + 1} - Time: {t}s - Average training loss: {avg_loss}')
+                tf.print(f'Epoch: {epoch + 1} - learning_rate: {optimizable_model.optimizer.learning_rate.numpy()}')
+                tf.print(f'Epoch: {epoch + 1} - Validation loss: {val_res[0]} - Performance on validation set: {val_res[1]}')
 
         # Check if model works after pruning
         pruned_performance = optimizable_model.evaluate(validation_dataset, verbose=0, return_dict=False)[-1]
         if verbose:
-            print(f'Optimized model performance on validation set, after fine-tuning: {pruned_performance}')
+            tf.print(f'Optimized model performance on validation set, after fine-tuning: {pruned_performance}')
 
         if __compare__(pruned_performance, rtol * baseline_performance, not increasing):
             bad_trials = 0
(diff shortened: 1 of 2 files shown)


hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
```diff
@@ -15,7 +12,6 @@
 from hls4ml.optimization.dsp_aware_pruning.keras.reduction import reduce_model
 from hls4ml.optimization.dsp_aware_pruning.scheduler import OptimizationScheduler
 
-np_config.enable_numpy_behavior()
```
vloncar, line 18: Can you provide more details as to why this breaks the flow on some setups?

hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
```diff
@@ -121,7 +117,7 @@ def optimize_model(
     model.compile(optimizer, loss_fn, metrics=[validation_metric])
     baseline_performance = model.evaluate(validation_dataset, verbose=0, return_dict=False)[-1]
     if verbose:
-        print(f'Baseline performance on validation set: {baseline_performance}')
+        tf.print(f'Baseline performance on validation set: {baseline_performance}')
```
vloncar, line 120: Will this actually print the statement, or will it "maybe print" depending on some external case?

calad0i: In tf execution environment (e.g., tf.function with jit), it is supposed to print, and it worked for me in other cases. Did not have the setup locally to run this part specifically, thus not tested here.

hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
calad0i: Some indexing and other behaviors are changed. e.g.: in default mode `tf.reduce_all(a>b)` is fine, but will raise if this is set. This causes some qkeras quantizers to return dicts for unknown reason, and I did not track down the direct cause.
Non-comprehensive list of failed tests:
hgq, conv1d, conv1dpointwise, qkeras po2, repack stream, qkeras mnist cnn, binary cnn

hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
bo3z: The NumPy flag is only required for printing the loss tensors during training. Everything else should work fine with normal Python prints. However, there are no tests that run the full dsp-optimization flow and train a model with sparsity, so we need to make sure that the behaviour still stays the same and that the loss tensors are printed.

hls4ml/optimization/dsp_aware_pruning/keras/__init__.py
bo3z: Where does this break occur? Is it with a newer version of NumPy or TensorFlow? Because the CI/CD on GitHub doesn't shows this, so I am just wondering under what environment this occurs.

calad0i: Gitlab CI test is not affected because this line is not executed before those (qkeras po2 and hgq) tests: only dsp aware pruning related tests will import the optimization api and trigger this line, and those don't run before the broken tests.

calad0i: If some tests are run with numpy behavior enabled. Full failure pattern: https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/pipelines/8406093

bo3z: I see. I have 2 comments on this but in general, I am okay with merging this PR as long as the loss is actually printed with `tf.print`. First, the error causing seems a bit artificial - is it ever likely that a user is going to import both `dsp_aware_pruning` and `hgq` / `po2` in the same module. Secondly, looking at the test logs this seems like this is a TensorFlow issue in some recent update? Looking at the docs for this function: https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/experimental_enable_numpy_behavior maybe the parameter `dtype_conversion_mode` can help solve this issue without changing the print statements.

calad0i: Response to your comments:
1. I don't think the errors are artificial -- if one runs the tests on one node, this is what to expect. I spend at least half an hour just to debug why some irrelevant tests breaks after some editing, and just to find out that the test execution order changed and things breaks because of this.
2. The issue is that global behavior is silently overridden by accessing a submodule of a package, and this should be avoided to the max extent possible. In this specific case, it only breaks hgq/po2 conversion hls4ml, but it could introduce undefined behavior on other places. I would suggest changing the print statement using `tf.print` to avoid affecting things outside, or pack it into a context manager to restore the original state after exiting the module.

Regarding if losses is actually printed, I don't have the setup to validate it on my side.

bo3z:
I just ran this code - removing the NumPy behaviour doesn't break prints, it breaks the line above `avg_loss = round(epoch_loss_avg.result(), 3)` in the file `dsp_aware_pruning/keras/__init__.py` (line 234). So this has to do with querying EagerTensors (which are not always evaluated until necessary). The error I get is: 

```
AttributeError: EagerTensor object has no attribute 'astype'. 
        If you are looking for numpy-related methods, please run the following:
        from tensorflow.python.ops.numpy_ops import np_config
        np_config.enable_numpy_behavior()
```

So if we remove this import we need to find a way to print the loss during training.

calad0i:
> I just ran this code - removing the NumPy behaviour doesn't break prints, it breaks the line above `avg_loss = round(epoch_loss_avg.result(), 3)` in the file `dsp_aware_pruning/keras/__init__.py` (line 234). So this has to do with querying EagerTensors (which are not always evaluated until necessary). The error I get is:
> 
> ```
> AttributeError: EagerTensor object has no attribute 'astype'. 
>         If you are looking for numpy-related methods, please run the following:
>         from tensorflow.python.ops.numpy_ops import np_config
>         np_config.enable_numpy_behavior()
> ```
> 
> So if we remove this import we need to find a way to print the loss during training.

Use `tf.cast(x)` instead of `x.astype` would solve that.

bo3z:
> > I just ran this code - removing the NumPy behaviour doesn't break prints, it breaks the line above `avg_loss = round(epoch_loss_avg.result(), 3)` in the file `dsp_aware_pruning/keras/__init__.py` (line 234). So this has to do with querying EagerTensors (which are not always evaluated until necessary). The error I get is:
> > ```
> > AttributeError: EagerTensor object has no attribute 'astype'. 
> >         If you are looking for numpy-related methods, please run the following:
> >         from tensorflow.python.ops.numpy_ops import np_config
> >         np_config.enable_numpy_behavior()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So if we remove this import we need to find a way to print the loss during training.
> 
> Use `tf.cast(x)` instead of `x.astype` would solve that.

That's a likely solution, but in the code we don't actually call `x.astype`. This is called somewhere internally in TF when trying to get the value of the loss tensor (`epoch_loss_avg.result()`), so we need to see what the alternative is to get the loss value to this function. Also do we now if this buggy behaviour is introduced by an updated in NumPy / TF / QKeras?

bo3z:
Turns out the changes in this PR have already been merged into main and DSP-aware pruning is actually broken in the main branch. My guess is that one of @calad0i previous PR accidentally included the change to remove enable_numpy_behaviour.

The proposed solution with tf_prints is on the right track, but one also needs to remove the round. I also found a minor inconsistency with the function docstring, and to avoid opening a PR to a PR, the two fixes are now in PR #1396. 

I suggest we close this PR, as the core change (removing enable_numpy_behaviour) is already in main and merge #1396.

JanFSchulte:
Closed in favor of #1396

PR #1094, opened 2024-10-26
Lazy converter imports and migrate to pyproject.toml

calad0i:
# Description

Make all front-end specific dependencies (`tf`, `(q)keras`, `torch`, `onnx`) optional and imported lazily.
All handlers are registered now, no matter if the frontend specific package is present. The frontend specific packages are only imported if the corresponding handler is used. This should remove the strong dependency of hls4ml on `qkeras` and `tensorflow` (e.g., when converting a model not in `tf.keras`, tensorflow no longer need to be installed).

## Type of change

## Tests

**Test Configuration**:


vloncar:
I agree with the spirit of the change but not the execution. Optional dependencies are the way to go but not unchecked lazy imports. Having imports randomly fail in the middle of some function and leaving the user to deal with the problem themselves is not friendly.

In the original proposal for modular hls4ml, frontends and backends would be installable extras. All of them. In practice this would mean basic install of hls4ml would only provide the core classes that can be used to save/load hls4ml models (a feature currently in development) and one backend (which has to be enforced somehow in `setup.py`). We didn't manage it before, but we could try again. For example, wiring the converters in such a way that if hls4ml was not installed with `pip install hls4ml[torch]` there would be no torch converter and the function `convert_from_pytorch_model()` would print that it is not available and how it should be installed to make it available. Same for ONNX and Keras and same for backends. So if a user wants an installation for converting Keras models to Vitis firmware the installation command would be `pip install hls4ml[keras,vitis]`. If they try to convert a pytorch model for oneAPI backend on that installation they should get a message that this is not what their current installation supports.

To achieve that we should make sure that our internals don't depend on big external libraries, in addition to some reordering of the imports that this PR does. To that end, we should focus on removing qkeras as the internal dependency from types and from profiling functions. Since we're already discussing ditching QKeras official for a drop-in replacement library, we can replace the internal dependency with chang's [reusable optimizers](https://github.com/calad0i/quantizers/), which mimic qkeras anyway. Then this new replacement library (yet to be named) can be built on top of the optimizers library. It would contain the definitions of layers (QDense, QConv1D/2D etc) that rely on our optimizers. Our keras frontend could then depend on this library, if even needed (as it seems that we only really need qkeras' optimizers for parsing, not the layer definitions).

calad0i:
> Having imports randomly fail in the middle of some function and leaving the user to deal with the problem themselves is not friendly

I was thinking a missing import is rather self-explanatory. Here are few proposals to throw an informative error for this without large restructuring for the short to midterm. Restructuring the whole project to multiple frontend and backend sounds like an overkill for this particular issue at the moment.

- For the frontend issue, I guess this is only relevant to the `onnx` and `(q)keras` frontend when the user passes the saved model in the config dict to the framework, as otherwise passing the model object to `hls4ml` implies the existence of the corresponding package (e.g., `torch`). In this case, I think we can just ward the `*_to_hls` function with a try-and-catch and map the corresponding import error to a custom error message prompting for installing `hls4ml[***]`. Possible import error can be thrown in the future save-load part though (likely only for `qkeras`, see below)
- More dirty but universal, wrap `importlib.import_module` and throw custom error on the lazy imports.

For types & quantizers, the `qkeras` dependency appears to be only interfacing with the `qkeras` models, but they persist the whole lifecycle of the hls models. In my opinion, the quantizers should end their lifecycle at the handler level for quantized models, but maybe there is some reason at the time of design.

The standalone `quantizers` library can be used in-place of `qkeras` quantizers in most cases, with two main caveats:
- the handling of `alpha` is different, `auto_po2` for example. Though no one should really be using them I guess.
- exponent types are not support in `quantizers`. Setting `m=0` in the minifloat quantizer could mimic that but probably with slightly different rounding.

calad0i:
Added hints on import failures from package metadata, except for profiler and dsp pruning, where the module is more or less isolated. Migrated to pyproject.toml for package config.

calad0i:
I'm not sure if hls4ml is supposed to expose `contrib` in the global namescope. Let's discuss this in the next meeting

vloncar (changes requested):
Discussed with Chang. Mostly ready, few cosmetics remaining.

pyproject.toml
```diff
@@ -1,10 +1,100 @@
 [build-system]
-# AVOID CHANGING REQUIRES: IT WILL BE UPDATED BY PYSCAFFOLD!
-requires = ["setuptools>=46.1.0", "setuptools_scm[toml]>=5", "wheel"]
 build-backend = "setuptools.build_meta"
 
+requires = [ "setuptools>=61", "setuptools-scm>=8" ]
+
+[project]
+name = "hls4ml"
+version = "1.0.0"
```
vloncar, line 8: TODO Check what happens if this is removed, because it should be overriden by the scm somehow.

vloncar:
This looks good to me, with just the extra test. Anyone else wants to take a look?

pyproject.toml
calad0i: Fixed

JanFSchulte (approved):

PR #1102, opened 2024-10-29
Change indexing in filling result for io_parallel convolutions, Vitis

jmitrevs:
# Description

Sometimes, though it's hard to exactly repeat the problem, one would see:

```
WARNING: [HLS 214-356] Array transformation on pointer indexing may lead to incorrect bank selection. Use array index instead of pointer. (firmware/nnet_utils/nnet_conv1d_resource.h:97:11)
ERROR: [HLS 214-384] in function 'void nnet::conv_1d_resource_cl<ap_ufixed<9, 0, (ap_q_mode)4, (ap_o_mode)0, 0>, ap_fixed<24, 9, (ap_q_mode)5, (ap_o_mode)3, 0>, config29>(ap_ufixed<9, 0, (ap_q_mode)4, (ap_o_mode)0, 0>*, ap_fixed<24, 9, (ap_q_mode)5, (ap_o_mode)3, 0>*, config29::weight_t*, config29::bias_t*) (.484)': Cannot apply array transformation pragma/directive because of full array load/store. (firmware/nnet_utils/nnet_conv1d_resource.h:97:11)
```

when using io_parallel convolutions with Vitis HLS. This change seems to fix the issue

## Type of change

## Tests

The pytests should make sure this is correct. I made sure that the keras API tests run both `Resource` and `Latency` for conv1d and conv2d to make sure this code is tested. I was also able to test synthesis on the conv2d pytest project--note, it does not have any activation. (I understand the original version was a workaround for such a case.)

## Checklist

### DIFF
--- a/hls4ml/templates/vitis/nnet_utils/nnet_conv1d_resource.h
+++ b/hls4ml/templates/vitis/nnet_utils/nnet_conv1d_resource.h
@@ -94,7 +94,7 @@ PartitionLoop:
         ResultLoop:
             for (unsigned i_res = 0; i_res < mult_n_out; i_res++) {
                 #pragma HLS UNROLL
-                res[i_part * CONFIG_T::n_partitions * CONFIG_T::n_pixels + i_pxl * CONFIG_T::n_pixels + i_res] =
+                res[i_part * CONFIG_T::n_pixels * mult_n_out + i_pxl * mult_n_out + i_res] =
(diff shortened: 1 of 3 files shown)


vloncar:
For historic reference, the original way of advancing the index was implemented that way as it was observed that normal indexing causes synthesis issues if the layer is the last layer of the model. I can't reproduce that with Vitis HLS anymore. Since that case is very rare anyway, I hope we'll never have to revisit the issue again.

vloncar (approved):

PR #1103, opened 2024-10-29
Multi output fix 2

calad0i:
# Description

Fix a couple of corner cases where model generation/compile fails. Now one can let a model output at (almost) all layers (except flatten/reshape simultaneously with its input in `io_parallel`). 

- Fix output node updating mechanism. The current parser assumes `set(outputs)-set(inputs)` are the output nodes, which is not true if some intermediate nodes are being outputted, or if any inputs is unused.
- `remove_node` is updated to correctly rewire in all cases, including leaf/intermediate+leaf nodes. Deprecate arg `rewire`.
- Add early stream flatten optimizer to pass original size to potential repack optimizers
- Added 1-to-3 repack catapult template

## Type of change

## Tests
`test/pytest/test_multiout_network.py/test_multi_output_nn_2`

## Checklist

### DIFF
--- a/hls4ml/backends/fpga/passes/clone.py
+++ b/hls4ml/backends/fpga/passes/clone.py
@@ -54,41 +54,59 @@ class CloneOutput(OptimizerPass):
         if isinstance(node, Clone):
             return False
 
-        return True
-
-    def transform(self, model, node):
-        if model.config.get_config_value('IOType') != 'io_stream':
+        # Not needed for io_parallel
+        io_type = node.model.config.get_config_value('IOType')
+        if io_type != 'io_stream':
             return False
 
+        # Check if the output is used more than once
         output_map = node.get_output_use_map()
+        in_output = node.name in node.model.outputs
+        for output in node.outputs:
+            if len(output_map[output]) + in_output > 1:
+                # model output also need a stream
+                return True
+
+        return False
+
+    def transform(self, model, node):
+
+        output_map = node.get_output_use_map()
+        in_output = node.name in node.model.outputs
 
         transformed = False
         for output in node.outputs:
-            if len(output_map[output]) > 1:
-                if len(output_map[output]) > 3:
-                    print(
-                        'WARNING: Cloning output {} of {} ({}) more than 3 times not currently supported'.format(
-                            output, node.__class__.__name__, node.name
-                        )
-                    )
-                    return False
-                out_var = node.get_output_variable(output)
-                for i, layer in enumerate(output_map[output], 1):
-                    attrs = {'size': np.prod(out_var.shape)}
-                    idx = layer.inputs.index(output)
-                    layer.inputs[idx] = output + '_cpy' + str(i)
+            n_outputs = len(output_map[output]) + in_output
+            if n_outputs == 1:
+                continue
+            if n_outputs > 3:
+                msg = f'ERROR: Cloning output {output} of {node.__class__.__name__} ({node.name}) more than 3 times not currently supported'  # noqa: E501
+                raise ValueError(msg)
 
-                clone_layer: Clone = model.make_node(
-                    Clone,
-                    'clone_' + node.name,
-                    attrs,
-                    [output],
-                    [output + '_cpy' + str(i + 1) for i in range(len(output_map[output]))],
-                )
-                for i in range(len(output_map[output])):
-                    key = output + '_cpy' + str(i + 1)
-                    clone_layer.attributes[key].type = node.attributes['result_t']
-                model.insert_node(clone_layer)
-                transformed = True
+            out_var = node.get_output_variable(output)
+            attrs = {'size': np.prod(out_var.shape)}
+
+            i0 = 1
+            if in_output:
+                # If the value is used as output, add one extra stream
+                idx = node.model.outputs.index(node.name)
+                node.model.outputs[idx] = node.name + '_cpy1'
+                i0 = 2
+            for i, layer in enumerate(output_map[output], i0):
+                idx = layer.inputs.index(output)
+                layer.inputs[idx] = output + f'_cpy{i}'
+
+            clone_layer: Clone = model.make_node(
+                Clone,
+                'clone_' + node.name,
+                attrs,
+                [output],
+                [output + '_cpy' + str(i + 1) for i in range(n_outputs)],
+            )
+            for i in range(n_outputs):
+                key = output + '_cpy' + str(i + 1)
+                clone_layer.attributes[key].type = node.attributes['result_t']
+            model.insert_node(clone_layer)
+            transformed = True
 
         return transformed
--- a/hls4ml/backends/fpga/passes/clone.py
+++ b/hls4ml/backends/fpga/passes/clone.py
@@ -1,4 +1,4 @@
-import numpy as np
+from math import prod
 
 from hls4ml.backends.template import FunctionCallTemplate
 from hls4ml.model.layers import Layer, register_layer
@@ -80,11 +80,12 @@ class CloneOutput(OptimizerPass):
             if n_outputs == 1:
                 continue
             if n_outputs > 3:
-                msg = f'ERROR: Cloning output {output} of {node.__class__.__name__} ({node.name}) more than 3 times not currently supported'  # noqa: E501
+                msg = f'ERROR: Cloning output {output} of {node.class_name}\
+                      ({node.name}) more than 3 times not currently supported'
                 raise ValueError(msg)
 
             out_var = node.get_output_variable(output)
-            attrs = {'size': np.prod(out_var.shape)}
+            attrs = {'size': prod(out_var.shape)}
 
             init_stream_idx = 1
             if in_output:
--- a/hls4ml/backends/fpga/passes/clone.py
+++ b/hls4ml/backends/fpga/passes/clone.py
@@ -86,13 +86,13 @@ class CloneOutput(OptimizerPass):
             out_var = node.get_output_variable(output)
             attrs = {'size': np.prod(out_var.shape)}
 
-            i0 = 1
+            init_stream_idx = 1
             if in_output:
                 # If the value is used as output, add one extra stream
                 idx = node.model.outputs.index(node.name)
                 node.model.outputs[idx] = node.name + '_cpy1'
-                i0 = 2
-            for i, layer in enumerate(output_map[output], i0):
+                init_stream_idx = 2
+            for i, layer in enumerate(output_map[output], init_stream_idx):
                 idx = layer.inputs.index(output)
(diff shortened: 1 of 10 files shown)


hls4ml/backends/fpga/passes/clone.py
```diff
@@ -54,41 +54,59 @@ def match(self, node):
...
+            if n_outputs == 1:
+                continue
+            if n_outputs > 3:
+                msg = f'ERROR: Cloning output {output} of {node.__class__.__name__} ({node.name}) more than 3 times not currently supported'  # noqa: E501
+                raise ValueError(msg)
+
+            out_var = node.get_output_variable(output)
+            attrs = {'size': np.prod(out_var.shape)}
+
+            i0 = 1
```
jmitrevs, line 89: This variable should have a better name.

test/pytest/test_multiout_network.py
```diff
@@ -19,25 +19,40 @@ def model():
     return model
 
 
+@pytest.fixture(scope='module')
+def model2():
+    in1 = keras.layers.Input(shape=(24, 8))
+    in2 = keras.layers.Input(shape=(16))
+    out1 = keras.layers.Conv1D(1, 3)(in1)
+    out1 = keras.layers.Flatten()(out1)
```
jmitrevs, line 27: I would prefer if the variables names weren't reused since initially it confused me, but it really is a minor point, so if am fine leaving it as is.

hls4ml/model/graph.py
```diff
@@ -506,6 +506,8 @@ def insert_node(self, node, before=None, input_idx=0):
 
         if next_node is not None:
             next_node.inputs[input_idx] = node.outputs[0]
+        else:
+            self.outputs = [node.outputs[0] if name == prev_node.outputs[0] else name for name in self.outputs]
```
jmitrevs, line 510: Is it obvious that prev_node.outputs[0] needs to be replaced with node.outputs[0] node is added? Both seem to have some rationale in my mind.

hls4ml/backends/fpga/passes/clone.py
jmitrevs: Or at least a comment saying how it's used.

jmitrevs:
Note, this PR includes #1119.

hls4ml/model/graph.py
```diff
@@ -514,47 +516,49 @@ def insert_node(self, node, before=None, input_idx=0):
...
             node (Layer): The node to remove
-            rewire (bool, optional): If `True`, connects the outputs of the previous node
-                to the inputs of the next node
+            rewire (bool, optional): Deprecated, no effect
 
         Raises:
-            Exception: If an attempt is made to rewire a leaf node or a node with multiple
-                inputs/outputs.
+            Exception: If an attempt is made to rewire a node with multiple inputs/outputs.
+        """  # noqa: E501
```
jmitrevs, line 531: Minor point:  why the E501? Why not wrap it?

jmitrevs (approved):

jmitrevs:
This looks good to me, and I would approve it, but didn't want to skip the step of looking at the above comments.

test/pytest/test_multiout_network.py
calad0i: It is actually a rather common practice with keras functional api to reuse same variable for representing the same data stream, see example at https://www.tensorflow.org/guide/keras/functional_api#a_toy_resnet_model

hls4ml/model/graph.py
calad0i: Laziness. Though, I did't find it very helpful to wrap long docstrings.
Found a plugin to do it automatically for me thus will change it.

hls4ml/model/graph.py
calad0i: The original logic assumes all outputs not consumed by any other node is the model output, and vice versa. This weakens the assumption, in which when a node is appended to an output node, it assumed to add a stage in the same pipeline but not causing other side effects. Otherwise, the appended node stays stray after operation, if not otherwise fixed by an optimizer. Tests will fail if this is not included.

hls4ml/backends/fpga/passes/clone.py
calad0i: For a short-lived variable appearing three times in 6 line, I didn't expect it to cause confusion. Change to `start_idx`

hls4ml/backends/fpga/passes/clone.py
```diff
@@ -54,41 +54,59 @@ def match(self, node):
...
-                for i in range(len(output_map[output])):
-                    key = output + '_cpy' + str(i + 1)
-                    clone_layer.attributes[key].type = node.attributes['result_t']
-                model.insert_node(clone_layer)
-                transformed = True
+            n_outputs = len(output_map[output]) + in_output
+            if n_outputs == 1:
+                continue
+            if n_outputs > 3:
+                msg = f'ERROR: Cloning output {output} of {node.__class__.__name__} ({node.name}) more than 3 times not currently supported'  # noqa: E501
```
vloncar, line 83: This will print the node class (a backend-specific instance) which may be confusing to the user. We have a `node.class_name` for this purpose.

hls4ml/backends/fpga/passes/clone.py
```diff
@@ -54,41 +54,59 @@ def match(self, node):
...
+                raise ValueError(msg)
+
+            out_var = node.get_output_variable(output)
+            attrs = {'size': np.prod(out_var.shape)}
```
vloncar, line 87: The output of `np.prod()` will be an `ndarray`, which works for now, but may trip up in the future (serialization and attribute validation will fail, since we would expect an integral value). Since we're already changing the code here, we might as well sort it out now.

hls4ml/backends/fpga/passes/inplace_parallel_reshape.py
```diff
@@ -11,14 +11,19 @@ class InplaceParallelReshape(OptimizerPass):
     """
 
     def match(self, node):
-        return isinstance(node, Reshape)
+        if not isinstance(node, Reshape):
+            return
```
vloncar, line 15: Can we rewrite this so it returns booleans or perhaps combine the two checks?

hls4ml/backends/fpga/passes/inplace_parallel_reshape.py
```diff
@@ -11,14 +11,19 @@ class InplaceParallelReshape(OptimizerPass):
...
-
         outvar = node.get_output_variable()
         invar = node.get_input_variable()
         newoutvar = InplaceTensorVariable(outvar, invar)
         node.set_attr(node.outputs[0], newoutvar)
+        if node.name in model.outputs:
+            prev_node = node.get_input_node()
+            assert (
+                prev_node.name not in model.outputs
+            ), f"Cannot output node {prev_node.name}: reshape is a no-op in io_parallel. As a result, the previous node {prev_node.name}'s output will be used as the output. However, this node is already an output."  # noqa: E501
```
vloncar, line 27: Does this trigger the bug/feature of black with formatting asserts, or is E501 just out of laziness? :smile: 

And a nitpick: We prefer single quotes wherever possible

hls4ml/backends/fpga/passes/inplace_stream_flatten.py
```diff
@@ -12,12 +12,13 @@ class InplaceStreamFlatten(OptimizerPass):
...
     def match(self, node):
         # Reshape acts as a Flatten layer when the result has 1 dimension
-        return isinstance(node, Reshape) and len(node.get_output_variable().shape) == 1
-
-    def transform(self, model, node):
-        if model.config.get_config_value('IOType') != 'io_stream':
+        if not (isinstance(node, Reshape) and len(node.get_output_variable().shape)) == 1:
+            # Reshape with multiple outputs will be kept as is, or repack cannot handle different shapes
             return False
+        io_type = node.model.config.get_config_value('IOType')
```
vloncar, line 18: why not a one-liner? :wink:

hls4ml/model/graph.py
```diff
@@ -514,47 +516,52 @@ def insert_node(self, node, before=None, input_idx=0):
...
+        By default, this function can connect the outputs of previous
+        node to the input of next one. If the removed node has multiple
+        inputs/outputs tensors, an exception is raised.
 
         Args:
-            node (Layer): The node to remove
-            rewire (bool, optional): If `True`, connects the outputs of the previous node
-                to the inputs of the next node
+            node (Layer): The node to remove rewire (bool, optional):
+            Deprecated, no effect
```
vloncar, line 529: Would marking this as deprecated according to the [docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) work? Would be nice to see

test/pytest/test_multiout_network.py
```diff
@@ -50,3 +65,24 @@ def test_multi_clone(model, data, backend: str, io_type: str):
 
     assert np.allclose(r_hls[0], r_keras[0], atol=1e-5, rtol=0)
     assert np.allclose(r_hls[1], r_keras[1], atol=1e-5, rtol=0)
+
+
+@pytest.mark.parametrize('backend', ['Vivado', 'Quartus', 'Vitis', 'Catapult', 'OneAPI'])
+@pytest.mark.parametrize('io_type', ['io_parallel', 'io_stream'])
+@pytest.mark.parametrize('strategy', ['latency', 'resource'])
+def test_multi_output_nn_2(model2, data2, backend: str, io_type: str, strategy: str):
```
vloncar, line 73: Can we be more descriptive instead of appending `2` to model, data and test?

vloncar (changes requested):
Looks very good so far, I'd just recommend minor cosmetic changes

hls4ml/model/graph.py
```diff
@@ -514,47 +516,54 @@ def insert_node(self, node, before=None, input_idx=0):
...
+        node to the input of next one. If the removed node has multiple
+        inputs/outputs tensors, an exception is raised.
 
-        Args:
-            node (Layer): The node to remove
-            rewire (bool, optional): If `True`, connects the outputs of the previous node
-                to the inputs of the next node
+        :param node: The node to remove.
+        :type node: Layer
+        :param rewire: Deprecated, no effect.
```
jmitrevs, line 529: This docstring formatting seems different than what we usually do. We should remain consistent.

hls4ml/model/graph.py
jmitrevs: I didn't quite follow. Isn't this to decide when you insert a node after a node whose output is in the outputs, whether that stays in the outputs or the new node output is in the outputs? But really I don't know which is better, so if you say this works better, I am fine with it.

PR #1104, opened 2024-10-29
put `code_gen.h` in custom namespace

calad0i:
# Description

Put `code_gen.h` also in custom namespace, if defined. Tests will be combined into with another PR.

### DIFF
--- a/hls4ml/backends/fpga/fpga_backend.py
+++ b/hls4ml/backends/fpga/fpga_backend.py
@@ -755,7 +755,7 @@ class FPGABackend(Backend):
 
         generated_code = (
             "template<class data_T, typename CONFIG_T>\n"
-            "class fill_buffer_{index} : public FillConv1DBuffer<data_T, CONFIG_T> {{\n"
+            "class fill_buffer_{index} : public nnet::FillConv1DBuffer<data_T, CONFIG_T> {{\n"
             "    public:\n"
             "    static void fill_buffer(\n"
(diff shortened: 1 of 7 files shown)


example-models
vloncar, line 1: can you separate this change from the rest?

vloncar:
ok, sounds reasonable, but may i has it caused problems already or you anticipate problems in the future?

example-models
calad0i: Committed by mistake, will revert it

calad0i:
> ok, sounds reasonable, but may i ask has it caused problems already or you anticipate problems in the future?

Should be good now. It is not causing problems for me at the moment, but didn't do a thorough test either.
Was wanting to isolate namespaces for jit compiling, but it still doesn't work as files/macro definitions are still conflicting.

vloncar:
You are JIT-ing the C++/HLS code? :smile:

calad0i:
Yes - it is like 60% time for compiling in *some* cases, and >200% for *some other*. Unfortunately, I had to create a separate process for each compilation, which is rather inefficient (need to parsing headers every time, etc).

vloncar:
The original reason for this has been abandoned (to my knowledge), but the change is still useful, as using default `nnet` namespace may cause problems while integrating the generated code into emulators. It should be triggered by existing tests anyway, let's wait for those and merge.

vloncar (approved):

PR #1112, opened 2024-11-06
Make auto default precision for pytorch parser

JanFSchulte:
Continuation of Jovan's PR https://github.com/fastmachinelearning/hls4ml/pull/1016 that made `auto` the default precision in the keras parser in `name` mode. This PR makes the same changes for the pytorch parser. 

## Type of change

## Tests

Verified that precision is set to `auto` when parsing a model in `name` mode with the `config_from_pytorch_model` function. 

## Checklist


hls4ml/utils/config.py
```diff
@@ -324,7 +328,10 @@ def config_from_pytorch_model(
     config = {}
 
     model_config = {}
-    model_config['Precision'] = default_precision
+    model_config['Precision'] = {}
+    model_config['Precision']['default'] = default_precision
+    if max_precision is not None:
+        model_config['Precision']['maximum'] = max_precision
```
jmitrevs, line 334: I realized I have to make the same change for the onnx parser. One question:  the keras parser also sets:
```
    model_config['BramFactor'] = 1_000_000_000
    model_config['TraceOutput'] = False
```
Is that something we want to be consistent on? Not setting BramFactor makes it default to np.inf, which is a better value, and TraceOutput defaults to False, so there is no change, but one could argue that it is good to put these in the config explicitly so that a user knows how to change them. I am fine either way, but I just wanted to bring it up.

JanFSchulte: Ah, you already opened a PR for ONNX. Never mind :)

jmitrevs (approved):

PR #1100, opened 2024-10-29
Update README.md for v1.0.0

bo3z:
# Description

> :memo: Fixes a minor mistake in the README.md. We can keep this PR open for any other changes required in the README before v.1.0.

## Type of change

For a new feature or function, please create an issue first to discuss it with us before submitting a pull request.

- [x] Documentation update

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [x] I have added tests that prove my fix is effective or that my feature works.

PR #1113, opened 2024-11-06
add max_precision to onnx parser

jmitrevs:
# Description

This is analogous to PR #1112 but for for ONNX. For ONNX, auto was the default, but max_precision was not set. This copies what we have for the Keras parser.

## Type of change

## Tests

We should make sure that this doesn't break anything, but truthfully, I am not sure if we have tests to test the max_precision behavior for any frontend, including Keras. Should we add such tests?

## Checklist

- [ ] I have made corresponding changes to the documentation. (But some exists in PR #1100)

### DIFF
--- a/hls4ml/utils/config.py
+++ b/hls4ml/utils/config.py
@@ -413,7 +413,7 @@ def config_from_pytorch_model(
 
 
 def config_from_onnx_model(
-    model, granularity='name', backend=None, default_precision='ap_fixed<16,6>', default_reuse_factor=1
+    model, granularity='name', backend=None, default_precision='fixed<16,6>', default_reuse_factor=1, max_precision=None
 ):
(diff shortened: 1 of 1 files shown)


hls4ml/utils/config.py
```diff
@@ -413,7 +413,7 @@ def make_layer_config(layer):
 
 
 def config_from_onnx_model(
-    model, granularity='name', backend=None, default_precision='ap_fixed<16,6>', default_reuse_factor=1
+    model, granularity='name', backend=None, default_precision='fixed<16,6>', default_reuse_factor=1, max_precision=None
```
JanFSchulte, line 416: Out of curiosity, why this change of removing the `ap_`? Should I do the same in https://github.com/fastmachinelearning/hls4ml/pull/1112?

JanFSchulte (approved):

PR #1116, opened 2024-11-08
Keras v3 Support

calad0i:
A# Description

Add keras v3 specific object-based parser and some layer handlers (no h5 or json loading supported). The current keras parser doesn't work with v3 functional models in general.

## Type of change

## Tests

`test/pytest/test_keras_v3_api.py`

**Test Configuration**:

Requires `keras>=3.0`. Skips the whole module if this requirement is not sufficed.

## Checklist


calad0i:
This PR will be rebased after its prerequisites are merged to resolve the conflicts.

docs/frontend/keras.rst
```diff
@@ -1,11 +1,19 @@
 ================
-Keras and QKeras
+Keras and its quantized variants
 ================
 
-Keras and the quantization library QKeras are well supported in ``hls4ml``. Currently, the Keras v2 (``tf.keras``) is the preferred version, and the future versions of ``hls4ml`` will expand support for Keras v3. The frontend is based on the parsing the serialized json representation of the model.
+Keras and the quantization library QKeras are well supported in ``hls4ml``. Both Keras v2 (``tf.keras``) and the new Keras v3 are supported. While the Keras v2 support is based on parsing the serialized json representation of the model, the Keras v3 support uses direct model inspection.
```
jmitrevs, line 5: I am not sure if the claim that the v2 support is based on parsing the serialized json representation is always true.

vloncar: It is. If a model is a live object we will parse the `Model.to_json()`. If the model is an .h5 file we will extract the `model_arch.json` from the archive and parse that.

hls4ml/utils/config.py
```diff
@@ -156,13 +156,21 @@ def config_from_keras_model(
...
+        import keras
 
-    reader = hls4ml.converters.KerasModelReader(model)
+        # model is keras.Model here
 
-    layer_list, _, _, _ = hls4ml.converters.parse_keras_model(model_arch, reader)
+        if keras.__version__ > '3.0':
+            layer_list, *_ = hls4ml.converters.parse_keras_v3_model(model)
+        else:
+            model_arch = json.loads(model.to_json())
```
jmitrevs, line 171: I don't know if the case was used in practice, but the old logic allows model to be a dict, and only if it's not not a dict, only then `model_arch = json.loads(model.to_json())`. Does the logic here work if model is a dict?

calad0i: `model.get_config` has slightly different behavior, where objects may be passed as-is (not serialized). v2 converter assumes a dict of fully serialized model config, so serialize+deserialize is used here.

vloncar (approved):

vloncar:
Two things to note for a future TODO:
- RNN is not supported via v3 handler for now
- We should get rid of `n_in` and `n_out` in pooling and standardize naming as with Conv

PR #1119, opened 2024-11-08
remove incorrect input from Constant nodes

jmitrevs:
# Description

The code in `graph._make_graph` adds an input to a layer if one is not specified, being the previous layer in the layer list. This follows the Sequential style in Keras. (As an aside, would it make sense in the future to move this logic to when the layer list is created instead of where it is?)

This behavior is wrong for Constant nodes, since they actually do not have an input. Most likely it causes no actual problem since it's ignored everywhere else, but this makes it consistent by changing it back to say there's no input.

## Type of change

## Tests

This is more for logical consistency, and I do not think it caused any problems, but it's better to have it correct, in case a problem comes up. Nevertheless, one should double-check as always that this doesn't break anyting.

## Checklist

### DIFF
--- a/hls4ml/model/layers.py
+++ b/hls4ml/model/layers.py
@@ -357,7 +357,7 @@ class Input(Layer):
 
 
 class Constant(Layer):
-    # one could consider making this a weight attribute, but given it's transient nature, I am not sure it helps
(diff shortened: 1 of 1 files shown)


vloncar (approved):
Makes sense

PR #1122, opened 2024-11-12
Added support for QONNX `Resize` node ingestion and tested with tiny UNet model

nghielme:
## Description 
All the necessary parts are added in order to properly ingest QONNX models containing `Resize` node
## Type of change

Extended the capability of QONNX ingestion of hls4ml

## Tests

I added two tests based on tiny UNet model in `test_qonnx.py` and one based on a simpler model taken from tiny UNet model. They have been both added in `pytest` directory.

## Checklist


hls4ml/backends/fpga/passes/clone.py
```diff
@@ -87,7 +87,7 @@ def transform(self, model, node):
                 )
                 for i in range(len(output_map[output])):
                     key = output + '_cpy' + str(i + 1)
-                    clone_layer.attributes[key].type = node.attributes['result_t']
+                    clone_layer.attributes[key].type = node.get_output_variable().type
```
jmitrevs, line 90: Isn't this identical?

hls4ml/converters/onnx_to_hls.py
```diff
@@ -63,6 +63,12 @@ def get_input_shape(graph, node):
     """
     rv = []
     for inp in node.input:
+        # this couple of lines doesn't look very nice but I don't think it would be considered as wrong.
+        # It is necessary for `Resize` node, since RoI input is empty but necessary to specify also scales
+        # array. It might be better handled in QONNX, refers to this issue for more details:
+        # https://github.com/fastmachinelearning/qonnx/issues/150
+        if inp == '':
+            continue
```
jmitrevs, line 71: Wouldn't it be better to put `()` as the shape? Then the inputs aren't shifted.

test/pytest/test_qonnx.py
```diff
@@ -309,6 +336,79 @@ def test_sep_conv(sep_conv_model, backend):
     np.testing.assert_allclose(y_qonnx.ravel(), y_hls4ml.ravel(), atol=1e-2, rtol=1)
 
 
+class EmptyFilledRoI(Transformation):
```
jmitrevs, line 339: Why is this in the test? Why do we need to remove a filled ROI? Can't we ignore it or assert that it must not be set?

jmitrevs (changes requested):

hls4ml/backends/fpga/passes/clone.py
nghielme: Here the output of the tests if you run the code with the original line:
```
    def match(self, node):
        if isinstance(node, (Pooling1D, Pooling2D, GlobalPooling1D, GlobalPooling2D)) and node.get_attr('pool_op') == 'Max':
>           return isinstance(node.get_input_variable().type.precision, XnorPrecisionType) and not isinstance(
                node.get_output_variable().type.precision, XnorPrecisionType
            )
E           AttributeError: 'FixedPrecisionType' object has no attribute 'precision'

hls4ml/backends/fpga/passes/xnor_pooling.py:14: AttributeError
```
Solved if you update it as proposed.

test/pytest/test_qonnx.py
nghielme: The RoI must be filled to run inferences in QONNX (see here: https://github.com/fastmachinelearning/qonnx/issues/150) so it has to be set while using QONNX model for inferences. If it is left filled during hls4ml ingestion it is considered as an input since in QONNX graph RoI is a node, generating garbage code that breaks the hls code.

hls4ml/converters/onnx_to_hls.py
nghielme: I pushed one commit that addresses this comment, I don't see the code updated from the GH review page but here is the [commit](https://github.com/fastmachinelearning/hls4ml/commit/5757ac62bd54885ac491ce3ba06a86a985dbc1ca)

nghielme (commented):
I think I addressed all the comments. I am noticing the CI/CD pipeline fails in 2 tests, [1](https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/45918890) and [2](https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/45918897), but apparently they are unrelated to this contribution.

jmitrevs:
Note, this now includes part (but not all) of #1130.

hls4ml/converters/onnx_to_hls.py
jmitrevs: I don't think this is necessary any more.

hls4ml/model/layers.py
```diff
@@ -1147,20 +1147,57 @@ class Resize(Layer):
...
-            if len(inp.shape) == 2:  # 1D -> width + chan
-                shape = [self.get_attr('out_width'), self.get_attr('n_chan')]
-                dims = [f'OUT_WIDTH_{self.index}', f'N_CHAN_{self.index}']
-            elif len(inp.shape) == 3:  # 2D -> height + width + chan
-                shape = [self.get_attr('out_height'), self.get_attr('out_width'), self.get_attr('n_chan')]
-                dims = [f'OUT_HEIGHT_{self.index}', f'OUT_WIDTH_{self.index}', f'N_CHAN_{self.index}']
+        if len(self.inputs) > 1:
+            # get the scales of Resize node from QONNX frontend
+            # see doc here: https://onnx.ai/onnx/operators/onnx__Resize.html
+            scales_idx = 2 if len(self.inputs) == 3 or len(self.inputs) == 4 else 1
```
jmitrevs, line 1153: We don't currently handle the case where "sizes" is passed instead of a "scales". If "sizes" is passed, I think we will just get the exception, "Resize 1D and Resize 2D are the ones supported in hls4m", which is not appropriate. If "sizes" is passed, we need to either use it appropriately or raise an exception saying that we don't currently support "sizes." (Either choices are fine.)

hls4ml/model/optimizer/passes/resize_remove_constants.py
```diff
@@ -0,0 +1,42 @@
...
+        model.remove_node(scales_node, rewire=False)
+        if len(node.inputs) >= 3 and node.inputs[1] != '':
+            # RoI is present only if more than 3 inputs are specified
+            # RoI position is always 1 when present
+            roi_node = node.get_input_node(node.inputs[1])
+            node.inputs[1] = ''
+            if not isinstance(roi_node, Constant):
+                raise RuntimeError("Non-constant RoI inputs are not supported")
+            model.remove_node(roi_node, rewire=False)
+        if len(node.inputs) == 4:
```
jmitrevs, line 33: This might be not needed if we don't support "sizes". If we do support, however, the message of the RuntimeError below needs to be made appropriate.

jmitrevs (changes requested):

hls4ml/model/optimizer/passes/resize_remove_constants.py
nghielme: Removed 👍

hls4ml/model/layers.py
nghielme: I added a check that raise an exception in case the input are 4, i.e. if sizes are passed. I think it is not needed to replicate the same check in this [optimiser](https://github.com/fastmachinelearning/hls4ml/pull/1122/files/b07e998d27de11c1334cacd0ea72b8b1a4b02efd#diff-ca5a48d22e8e2cad21ae235acb571aa79a9733cb64971063990099a43f39e996) since the code in layers.py is run first.

jmitrevs: Wait, didn't we say that self.inputs==2 was the old version ONNX when ROI was not used?

jmitrevs: Resize version 2 only has two inputs, the data input and scale. How old is it? I think it's easy to support.

hls4ml/model/optimizer/passes/resize_remove_constants.py
```diff
@@ -0,0 +1,35 @@
...
+        node.inputs[scales_idx] = ''
+        if not isinstance(scales_node, Constant):
+            raise RuntimeError("Non-constant shape inputs are not supported")
+        model.remove_node(scales_node, rewire=False)
+        if len(node.inputs) >= 2 and node.inputs[1] != '':
+            # RoI is present only if at least 2 inputs are specified
+            # RoI position is always 1 when present
+            roi_node = node.get_input_node(node.inputs[1])
+            node.inputs[1] = ''
+            if not isinstance(roi_node, Constant):
```
jmitrevs, line 30: Should we add a check on the constant value? What should it be if the ROI is not used?

nghielme: For the moment I considered  Resize v10 as unsupported

jmitrevs: Actually, the else 1 is also never true since we are not supporting v10. `scale_idx` is always 2.

nghielme: Right, RoI should be `''` or an empty vector. The first case is not supported by QONNX at the moment

nghielme (commented):
I reviewed all the comments, let me know if there's anything else to address. I see the pipeline failed but again it was due to the large time taken by some tests

jmitrevs (approved):

PR #1124, opened 2024-11-13
Add general transpose for vivado/vitis

calad0i:
# Description

Add general N-dimensional transpose for vivado/vitis backend. Resource consumption validated to be unchanged for all 2/3d io_parallel Vivado/Vitis HLS (2020.1/2023.2), and 2d io_stream Vitis HLS.

For Vivado HLS there is a resource overhead for io_sream transpose 2d. Template overload is used to use the original impl for io_stream 2d.

## Type of change

## Tests

`test/pytest/test_transpose_concat.py/test_highdim_permute`

## Checklist


hls4ml/backends/vivado/passes/reshaping_templates.py
```diff
@@ -97,16 +101,45 @@ def format(self, node):
...
-    static constexpr unsigned perm[3] = {{{perm_str}}};
-}};\n"""
 
-transpose_function_template = 'nnet::transpose_{dim}<{input_t}, {output_t}, {config}>({input}, {output});'
+transpose_include_list = ['nnet_utils/nnet_transpose.h', 'nnet_utils/nnet_transpose_stream.h']
+
+transpose_config_template = """struct {config_name} {{
+    static const unsigned dims = {dims};
+    static const unsigned N = {N};
+    static const unsigned* const from_shape;
```
jmitrevs, line 110: I assume you tried various things, and I know C++ can be kind of clunky for stuff like this, but is there an easier/less verbose way to initialize these arrays? Would std::arrays be more readable?

calad0i: This is due to that `vivado/vitis` cannot handle `constexpr` array indexing with "variables" (can be inferenced, but vivado/vitis compiler don't know how to do it.)
If we are to drop vivado hls support completely, this can be made `constexpr` and initialized with one-line, and recursive template unrolling will need to be used in the caller function to do the index mapping, which only works for vitis hls and takes forever on vivado hls. I did not try `std::array`, as I did not expect it to be more compatible than a constexpr raw array.

hls4ml/backends/vivado/passes/reshaping_templates.py
```diff
@@ -97,16 +101,45 @@ def format(self, node):
...
+const unsigned* const {config_name}::perm = {config_name}_perm;
+const unsigned* const {config_name}::perm_strides = {config_name}_perm_strides;
+"""
+
+transpose_function_template = 'nnet::transpose<{input_t}, {output_t}, {config_name}>({input}, {output});'
 
-transpose_include_list = ['nnet_utils/nnet_array.h', 'nnet_utils/nnet_stream.h']
+
+def permute_config_gen(name: str, shape: tuple[int, ...], perm: tuple[int, ...]):
+    new_shape = tuple(shape[i] for i in perm)
```
jmitrevs, line 131: A comment or docstring would be useful to understand how this works.

jmitrevs:
This PR looks to be quite useful and I think should be quite rapidly merged, since it makes pytorch streaming models more parsable (after removing the no longer needed graceful failure, with this PR).

hls4ml/backends/vivado/passes/reshaping_templates.py
calad0i: Resolved

calad0i:
> I would propose the following change with this PR since it's no longer needed:
> 
> ```
> diff --git a/hls4ml/model/optimizer/passes/convert_to_channels_last.py b/hls4ml/model/optimizer/passes/convert_to_channels_last.py
> index 606f42e5..5b1a69f1 100644
> --- a/hls4ml/model/optimizer/passes/convert_to_channels_last.py
> +++ b/hls4ml/model/optimizer/passes/convert_to_channels_last.py
> @@ -103,11 +103,6 @@ class ChannelsLastConverter(OptimizerPass):
>                  input = previous_node.name
>                  outshape = previous_node.get_output_variable().shape
>  
> -                if (model.config.config['IOType'] == 'io_stream') and len(outshape) == 3:
> -                    raise Exception(
> -                        'No 3D transpose available in io_stream, this model cannot be converted to channels-last'
> -                    )
> -
>                  if len(outshape) == 2:
>                      attributes = {'perm': [1, 0]}
>                  else:
> ```

Done

jmitrevs:
For a sanity check I am rerunning the pytests, but if they pass, I suggest we merge this PR.

jmitrevs (approved):

PR #1126, opened 2024-11-13
Multiline cpp fn def

calad0i:
# Description

Change clang-format config so that long function call/def will have each argument taking a new line, with brackets in their own lines.

The current formatter breaks only when a line is filled, causing multiple parameters sharing the same line, which creates walls of code that are not easy to read. This should mitigate the issue. 

Drawback: A few super long template from `ac_math.h` looks bad now, but as they are not that much better before I guess it's a reasonable tradeoff...

### Example:

Current:

```c++
void dense(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_out],
           typename CONFIG_T::weight_t weights[CONFIG_T::n_in * CONFIG_T::n_out],
           typename CONFIG_T::bias_t biases[CONFIG_T::n_out]) {
    #pragma HLS INLINE
    CONFIG_T::template kernel<data_T, res_T, CONFIG_T>::dense(data, res, weights, biases);
}
```

Modified:
```c++
void dense(
    data_T data[CONFIG_T::n_in],
    res_T res[CONFIG_T::n_out],
    typename CONFIG_T::weight_t weights[CONFIG_T::n_in * CONFIG_T::n_out],
    typename CONFIG_T::bias_t biases[CONFIG_T::n_out]
) {
    #pragma HLS INLINE
    CONFIG_T::template kernel<data_T, res_T, CONFIG_T>::dense(data, res, weights, biases);
}
```

## Type of change

- [x] Cosmatic

## Tests

>

**Test Configuration**:

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,7 +2,7 @@
 Language:        Cpp
 # BasedOnStyle:  LLVM
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: BlockIndent
 AlignArrayOfStructures: None
 AlignConsecutiveMacros: None
 AlignConsecutiveAssignments: None
@@ -26,8 +26,8 @@ AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: MultiLine
 AttributeMacros:
   - __capability
-BinPackArguments: true
-BinPackParameters: true
+BinPackArguments: false
+BinPackParameters: false
 BraceWrapping:
   AfterCaseLabel:  false
   AfterClass:      false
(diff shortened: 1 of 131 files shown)


vloncar:
I agree that the function declarations look nicer this way, that used to be the preferred-but-not-enforced style before we started using clang-format. However, the template declaration now looks really ugly.

Also, after hours of fiddling, I never managed to get labels to be indented and align to the code (in our case the loop), instead they are always aligned at the level of the surrounding brace. This is the biggest eyesore in the current codebase. Any suggestions as to how to fix that? Seems to be doable with astyle, but clang-format is more powerful and generally recommended, so I'd rather not switch the formatter

PR #1129, opened 2024-11-15
⚡ Add ESD Support to hls4ml

calad0i:
# Description

FS-ESDYLW-2__84620

Add support for general `EinsumDense` layer with any legit equation. Frontend only implemented for keras v3 now, but should also work with v2 if a handler is added.

Currently, only `vivado`/`vitis` backend is supported with `io_parallel` and `latency` strategy. I checked that model does synth successfully and cosim looks ok. Pragmas are not tuned; latency and resource consumption looks fine but not yet crosschecked with dense and pointwise conv.

## Prerequisite
This PR requires and includes #1116 (includes #1094), #1103
Will rebase this branch after these are merged. Only the last few (two at the time of creation) commits exclusively belongs to this PR.
 
## Type of change

## Tests

`test/pytest/test_einsum_dense.py/test_einsum_dense`

## Checklist

- [x] All except doc update


calad0i:
Close as included in #1116 now.

PR #1132, opened 2024-11-19
Fix problem with scale being a multidimensional array.

jurevreca12:
In case scale is a multidimensional array (e.g. `1 x num_channels`) `scale[0]` does not return a scalar. Instead
scale.item(0) should be used. 

I had this problem on the model added bellow. It has channel-wise scaling factors. The problem was
that the match function returned an array, which then lead to errors (truth value of an array is ambiguous).

The change should not affect the rest of the code, nor does it change behavior.
Since the change is so trivial, I didn't bother adding a test. However, if this is deemed desired, I
can also add a test.

Type of change:
- [ ] Documentation update
- [ ] Other (Specify)

## Tests
None

## Checklist

- [x] I have installed and run `pre-commit` on the files

[cnn_model.zip](https://github.com/user-attachments/files/17821675/cnn_model.zip)


jmitrevs (approved):

jmitrevs:
This looks good. Let's just double-check the pytests before merging.

PR #1139, opened 2024-12-04
Automatic type inference for `param_t` in Parametrised Activations

nghielme:
This simple PR implement a part related 

## Type of change

## Tests

I run some tests related to Parametrised Activations, already present in the pytests of hls4ml.

## Checklist

### DIFF
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -2,6 +2,7 @@ import math
 from typing import Iterable
 
 import numpy as np
+import struct
 
 from hls4ml.model.optimizer import ConfigurableOptimizerPass
 from hls4ml.model.types import (
@@ -561,15 +562,29 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
 
         return inferred_types
 
+    def _infer_const_precision(self, node, type_to_infer, attr_name):
+        inferred_types = []
+        def get_man_exp(f):
+            f = np.abs(f)
+            s = struct.pack('>f', f)
+            l = struct.unpack('>l', s)[0]
+            bits = '{:032b}'.format(l)
+            m = bits[-23:]
+            e = bits[-23-8:-23]
+            return m, e
+        alpha = node.get_attr(attr_name)
+        m, e = get_man_exp(alpha)
+        I = int(e, 2) - 127 + 1 # -127 is the bias of the exponent
+        W = m.rindex('1') + 2 # + 1 for accounting the index starting from 0, +1 for the leading 1 of the exponent
+        if alpha < 0:
+            I += 1
+            W += 1
+        node.attributes[type_to_infer].precision = FixedPrecisionType(W, I, True if alpha < 0 else False)
+        inferred_types.append(type_to_infer)
+        return inferred_types
+
     def _infer_par_act_precision(self, node, types_to_infer):
         inferred_types = []
-
-        # For threshold relu, set the parameter precision to be the input precision by default;
-        # for other parametrized activations, just allow the default precision to be used.
-        # Can override these values in the configuration by explicitly setting them.
-        if 'param_t' in inferred_types and self.get_attr('activation').lower() == 'thresholdedrelu':
-            in_type = node.get_input_variable().type.precision
-            node.attributes['param_t'].type = in_type
-            inferred_types.append('param_t')
-
+        if 'param_t' in types_to_infer:
+            inferred_types.extend(self._infer_const_precision(node, 'param_t', 'activ_param'))
         return inferred_types
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -574,14 +574,17 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
             e = bits[-23 - 8 : -23]
             return m, e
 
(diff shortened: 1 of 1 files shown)


nghielme:
I see some tests related to oneAPI fails; it's hard to me to understand why they fail, how should I proceed?

jmitrevs:
If you have a linux setup it should be pretty straightforward to install oneAPI, and then you can run the pytest. But we can wait to look at the other issues first. Maybe it will clear itself.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -1,4 +1,5 @@
 import math
+import struct
```
jmitrevs, line 2: Struct is much too low level for what we are doing here. We have a python float. We should use it, not look at the bits.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -561,15 +562,34 @@ def _infer_rnn_precision(self, node, types_to_infer):
...
-        # For threshold relu, set the parameter precision to be the input precision by default;
-        # for other parametrized activations, just allow the default precision to be used.
-        # Can override these values in the configuration by explicitly setting them.
-        if 'param_t' in inferred_types and self.get_attr('activation').lower() == 'thresholdedrelu':
-            in_type = node.get_input_variable().type.precision
-            node.attributes['param_t'].type = in_type
-            inferred_types.append('param_t')
+        def get_man_exp(f):
+            f = np.abs(f)
+            s = struct.pack('>f', f)
```
jmitrevs, line 570: Use calculations:  based on the value, you can easily determine how many bits you need. Going to structs is hard to follow.

jmitrevs (changes requested):

jmitrevs:
Note the utility we have in hls4ml already:
https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/utils/fixed_point_utils.py#L139-L141

JanFSchulte:
I wanted to try to install oneAPI myself, so I played with this PR a bit. The issue seems to be that the precision for the parameter of the leaky ReLU is reduced significantly, from `typedef ac_fixed<16,6,true> quantizer_param_t;` to a one-bit `typedef ac_fixed<1,0,false> quantizer_param_t;`. Vivado and the other backends seem to be able to handle it, but I'm not sure it makes sense in this case because we have negative slopes here and need it to be signed. The other backends seem to be able to deal with it. But for oneAPI, a signed variable is enforced to have at least two bits: 
```
signed _BitInt must have a bit size of at least 2
```
So we need to make sure to take this into account when inferring the precision for the parameters.

nghielme:
I'll take a look soon

jmitrevs:
Please check the logic of the precision setting.

jmitrevs:
I added a unit test to cover the various options, so I am more confident. It did discover an error in the max setting for unsigned FixedPrecisionType, which I fixed, and am including here, though it's logically independent.

hls4ml/model/optimizer/passes/infer_precision.py
```diff
@@ -573,9 +574,17 @@ def _infer_par_act_precision(self, node, types_to_infer):
...
+        if 'param_t' in types_to_infer:
+            if node.get_attr('activation').lower() == 'thresholdedrelu':
+                # For threshold relu, set the parameter precision to be the input precision by default;
+                in_type = node.get_input_variable().type.precision
+                node.attributes['param_t'].precision = in_type
+                inferred_types.append('param_t')
+            else:
+                # find a constant to represent the values
+                param = node.get_attr('activ_param')
+                precision = _get_precision_from_constant(param)
```
JanFSchulte, line 586: If I understand this correctly, we are basically hard-coding the bit width of the parameter to be 8 (9 if signed) and assign the fractional and integer bits based on the value. Is that correct? Because trying to find a way to infer the needed total precision has been something that has stumped me forever when working on the brevitas stuff, but it seems that here as well the only solution is to hardcode some arbitrary value.

JanFSchulte: Makes sense, thanks. I think I will then implement something similar to the non-power-of-2 cases for brevitas.

nghielme:
Looks good to me. One small note, I think the test could be rewritten in a more `pytest` way, like this:
```
@pytest.mark.parametrize(
    "val, expected_width",
    [
        (0, 1),
        (-1024, 2),
        (1024, 1),
        (0.03125, 1),
        (-0.03125, 2),
        (1.25, 3),
        (-1.25, 4),
        (1.1, 8),
        (-1.1, 9),
    ]
)
def test_precision_from_constant_unit(val, expected_width):
    """Test determining precision needed for a constant."""
    max_width = 8
    fp = _get_precision_from_constant(val, max_width)

    assert fp.min <= val <= fp.max
    assert fp.width == expected_width
    assert fp.signed == (val < 0)

    quantum = 2.0 ** -fp.fractional
    if expected_width < max_width:
        assert val % quantum == 0
```

JanFSchulte (approved):

JanFSchulte:
Tests have only the "expected" failures now, so I think this is ok. I agree with Nicolo's comment on the pytest though, so if you could integrate that before merging that would be great, Jovan.

jmitrevs:
I just put Nicolo's test in the pytests instead of the one I had (with minor pre-commit changes). I also updated to the latest main so ideally the test failures should be gone.

PR #1143, opened 2024-12-06
Fixes to parsing of pytorch models when using torch functionals

JanFSchulte:
A# Description

When torch functionals are used for pooling and flatten operations, some arguments were not correctly caught (or in the case of flatten, this type of usage was not supported at all). This PR fixes these issues. 

## Type of change

## Tests

The example model in https://github.com/pytorch/examples/blob/main/mnist/main.py would fail to parse before, and now succeeds, modulo the not supported log_softmax activiation in the end. 

## Checklist

### DIFF
--- a/hls4ml/converters/pytorch/reshape.py
+++ b/hls4ml/converters/pytorch/reshape.py
@@ -93,7 +93,7 @@ def parse_flatten_layer(operation, layer_name, input_names, input_shapes, node, 
     layer['class_name'] = 'Reshape'
     layer['name'] = layer_name
     layer['inputs'] = input_names
-    if node.op == "call_module":
+    if node.op == 'call_module':
         start_dim = class_object.start_dim
         end_dim = class_object.end_dim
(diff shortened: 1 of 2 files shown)


hls4ml/converters/pytorch/reshape.py
```diff
@@ -93,13 +93,23 @@ def parse_flatten_layer(operation, layer_name, input_names, input_shapes, node,
     layer['class_name'] = 'Reshape'
     layer['name'] = layer_name
     layer['inputs'] = input_names
-
-    start_dim = class_object.start_dim
-    end_dim = class_object.end_dim
-    if end_dim + 1 == 0 or end_dim + 1 > len(input_shapes[0]):
-        end_dim = len(input_shapes[0])
+    if node.op == "call_module":
```
jmitrevs, line 96: Vladimir suggests that this should be in single quotes.

jmitrevs (approved):

PR #1146, opened 2024-12-11
Don't overwrite already set accum_t, fix pointwise output resolution

jmitrevs:
# Description

Modify `Layer._set_accum_t()` to not overwrite an already existing `accum_t`, which can happen in an optimizer. Also, modify the pointwise init functions to do nothing, since the values are already set. Otherwise, the result_t was overwritten, potentially nullifying an explicit value set by a quantizer (a Quant node in a failing test). 

## Type of change

## Tests

The failing test that this fixes will be added soon. It is important that it doesn't break anything else.

TODO:  We should confirm that there are no other optimizers that may cause the same probelm

## Checklist

### DIFF
--- a/hls4ml/backends/fpga/fpga_layers.py
+++ b/hls4ml/backends/fpga/fpga_layers.py
@@ -73,12 +73,14 @@ class BatchNormalizationQuantizedTanh(Layer):
 class PointwiseConv1D(Conv1D):
     '''Optimized Conv1D implementation for 1x1 kernels.'''
 
-    # Nothing to do, will pick up function and config from class name
-    pass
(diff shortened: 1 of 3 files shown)


jmitrevs:
This needs https://github.com/fastmachinelearning/example-models/pull/19 to be merged.

jmitrevs:
With the updated example-models, and existing test (`test_qonnx.py::test_sep_conv[Vitis]`) fails in main but passes with this change. For the TODO, I moved the optimizer to merge the linear activation to later, so I do not think there are any other layers that will fail in this way.

JanFSchulte (approved):

PR #1149, opened 2024-12-13
oneAPI 2025.0 include changes

laurilaatu:
# Removing include for 2025.0

## Type of change

oneAPI 2025.0 removes `interfaces.hpp` and the relevant functions are available from `fpga_extensions.hpp`

## Checklist

### DIFF
--- a/hls4ml/templates/oneapi/myproject_test.cpp
+++ b/hls4ml/templates/oneapi/myproject_test.cpp
@@ -11,6 +11,10 @@
 
 #include <sycl/ext/intel/fpga_extensions.hpp>
 
+#if (__INTEL_CLANG_COMPILER < 20250000)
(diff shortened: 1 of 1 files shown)


jmitrevs:
Changing this to draft till we update our test environment.

bo3z (changes requested):
We can merge this but I would like to confirm this indeed works with oneAPI < 2025.0. If not, there should be an #ifdef checking for the version.

JanFSchulte:
Pytests have passed, so I guess this change is save an can be merged now?

jmitrevs:
The pytest wouldn't catch the failure since it's in the _test file. Lauri will add ifdefs, he said.

JanFSchulte:
Ah, I misunderstood. Great news about ifdefs.

PR #1152, opened 2024-12-13
remove test skip since problem fixed in qonnx

jmitrevs:
# Description

There currently is one test that is skipped because the qonnx channels-last conversion fails with it. Wit the current head of qonnx, this is no longer the case, so we no longer need to skip the test. However, we need to update our test environment first, so this is created as a draft at the moment

## Type of change
 [x] Other (Specify)  Stop skipping a test

## Tests

The pytests should all pass after we create a new testing container.

## Checklist


bo3z (commented):
This looks good to me, can we create a new Docker image and run the tests, so this can be merged?

jmitrevs:
This depends on #1200. After that goes in, this should be able to go in (though it should be confirmed).

JanFSchulte (approved):

PR #1155, opened 2024-12-17
added check for conv implementation

jicampos:
# Description

This change adds a check in the Vivado backend to skip instruction calculations for CONV1D and CONV2D when using linebuffer implementations. Since linebuffers do not require calculating **instructions**, **min_height**, or **min_width**, performing these calculations is redundant.

## Tests

Should not break current tests

## Checklist


jmitrevs:
I will just add that the reason for this change was because in some cases the instructions became huge and made compilation more difficult, even though they were then unused.

jmitrevs:
It does seem to cause some test failures that should be investigated.

hls4ml/backends/vivado/passes/conv_stream.py
```diff
@@ -18,7 +18,7 @@ def transform(self, model, node):
             raise Exception(f'Cannot generate instructions for node {node.name} ({node_class})')
 
     def _generate_1d_instructions(self, node):
-        if node.model.config.get_config_value('IOType') == 'io_stream':
+        if node.model.config.get_config_value('IOType') == 'io_stream' and node.get_attr('conv_implementation') == 'Encoded':
```
jmitrevs, line 21: One should maybe make this check case-insensitive.

jmitrevs:
I think "Instructions" implies encoded to me, but maybe not to everyone. But I do like the idea of moving this to the match. There may need to be some changes in the parameter creation to make sure unused variables aren't implicitly required, but that should be fairly easy to fix.

jicampos:
> Would it make sense to add this check in the match() method rather than the transform() method? That way the optimizer will never even be executed. Also, since this only applies to encoded implementation, should we rename this optimizer to reflect that?


We would still need to run the optimizer to add dummy weights in the case of LineBuffer implementations but we can skip generating instructions with the added condition in the if statement.

jmitrevs:
I think the suggestion from @bo3z is the better approach. Can we change it so that the check is on the match? There may be some downstream minor fixes needed with that approach, though, mainly allowing for undefined instructions and min widths/heights.

jicampos:
> I think the suggestion from @bo3z is the better approach. Can we change it so that the check is on the match? There may be some downstream minor fixes needed with that approach, though, mainly allowing for undefined instructions and min widths/heights.

I've updated the optimizer match() to check the IOType and implementation. The LineBuffer default values are now added to Conv2DConfigTemplate and Conv1DConfigTemplate.

hls4ml/backends/vivado/passes/convolution_templates.py
```diff
@@ -108,6 +108,11 @@ def format(self, node):
             else:
                 params['conv_fn'] = 'Conv1DResource'
 
+        if node.get_attr('implementation').casefold() == 'linebuffer':
```
jmitrevs, line 111: Would this fail if for some reason I set the implementation to encoded but then used io_parallel? I know it's not meaningful, but I wonder if it would be best to just set them if not set regardless of the implementation.

hls4ml/backends/vivado/passes/convolution_templates.py
```diff
@@ -239,6 +244,12 @@ def format(self, node):
         else:
             params['fill_fn'] = 'FillConv2DBuffer'
 
+        if node.get_attr('implementation').casefold() == 'linebuffer':
```
jmitrevs, line 247: Same issue here as above

jmitrevs (changes requested):

hls4ml/backends/vivado/passes/convolution_templates.py
jicampos: After updating the tests to include this case, it fails since the optimizer pass now checks for io_stream. The latest commit (a3ec21022042f627fe6ddaccebae04db0a19cf3d) resolves this issue. The test case has also been added in pytest.

jmitrevs (approved):

jmitrevs:
We should wait to make sure all the pytests pass first before merging. (Maybe got ahead of myself approving.)

jicampos:
> We should wait to make sure all the pytests pass first before merging. (Maybe got ahead of myself approving.)

It seems the tests are failing because of a connection issue in the qkeras tests, and is unrelated to the PR. 

Before rerunning any tests, should we update the catapult backend with the same changes?

jmitrevs:
I agree that the test failure is unrelated. If you can add the same to Catpult, that would be great. (No other backend has encoded.)

jicampos:
> I agree that the test failure is unrelated. If you can add the same to Catpult, that would be great. (No other backend has encoded.)

That sounds good, the catapult backend has been updated.

hls4ml/backends/catapult/passes/convolution_templates.py
```diff
@@ -210,6 +218,15 @@ def format(self, node):
         else:
             params['fill_fn'] = 'FillConv2DBuffer'
 
+        if (
```
jmitrevs, line 221: Is it possible to just always set default values?

hls4ml/backends/catapult/passes/conv_stream.py
```diff
@@ -6,7 +6,12 @@ class GenerateConvStreamingInstructions(OptimizerPass):
     '''Generates the instructions for streaming implementation of CNNs'''
 
     def match(self, node):
-        return isinstance(node, (Conv1D, SeparableConv1D, Conv2D, SeparableConv2D))
+        is_match = (
+            isinstance(node, (Conv1D, SeparableConv1D, Conv2D, SeparableConv2D))
+            and node.model.config.get_config_value('IOType').casefold() == 'io_stream'
```
jmitrevs, line 11: We haven't used casefold in other places. People has suggested to use 'lower' for consistency with other parts of the code.

jmitrevs (changes requested):

hls4ml/backends/catapult/passes/convolution_templates.py
```diff
@@ -94,6 +94,9 @@ def format(self, node):
         else:
             params['fill_fn'] = 'FillConv1DBuffer'
 
+        params.setdefault('min_width', node.get_attr('in_width'))
```
jmitrevs, line 97: I think these can be replaced with just `params['min_witdth'] = node.get_attr('in_width')` and similar. Note that params is created on line 82, so `params['min_witdth']` is not set. The setdefaults has the same effect as `params['min_witdth'] = node.get_attr('in_width')`, so it's better to just use the more straightforward form.

jmitrevs (changes requested):

jmitrevs (approved):

PR #1158, opened 2024-12-22
Infrastructure for saving/loading hls4ml models

vloncar:
# Description

Adds the ability to save and load hls4ml models (serialize/deserialize them). Given a `ModelGraph`, this will serialize it in a a single file which can be loaded at a later stage. The saved model doesn't depend on the original Keras/PyTorch/ONNX model in any way. 

The feature is in part inspired by Keras' model saving feature. The main format used for serialization is JSON, all objects save their state in dictionaries which are serialized into JSON. Assuming disk space is not a problem, generated JSON is nicely formatted during writing to file. No objects are pickled, as this is way too unsafe. The numpy arrays (weights) are saved in npz format. We save model graph (list of layers), the model information and config into separate files. This (along with some versioning information is packaged into a `.fml` file, which is just a `.tar.gz` with a different name. 

Internally, this works by adding a few methods to types, quantizers, layers and model graph itself. The interface is defined by the `Serializable` class. Classes would typically implement `serialize_state()` method, which should return a dictionary of current state of the object. Additionally, there's also a `serialize_class_name()` which is needed to know what instance are we saving, but most classes won't need to deal with this. Deserialization is done with a class method `deserialize()`. To support this feature some restructuring had to be done. `ModelGraph` has been intended to be created only with a layer list from a converter, which is not compatible with (de)serialization, so it was split into initialization of empty ModelGraph and conversion of layer list from converters to `Layer` objects. Furthermore, `Layer`'s initialization has to be skipped, as we're basically restoring a state post-initialization. Types and quantizers are more straightforward to save/load. Loaded model should be indistinguishable from the original, but there may be some corner cases of some hacks of internal state of layers (or partially optimized models) not working on loaded models, we can catch these over time. But for "final" models (one you're happy enough with to call `write()`/`compile()`/`build()` on) saving/loading should always work.

One somewhat ugly part in the current implementation is that due to the creation of dynamic wrapper classes, we cannot directly deserialize to them, instead we create the original types, and have to run `<backend>:specific_types` optimizer to truly get an object that is identical to the original one. Running that optimizer for a given backend looks a bit hacky, but is ok for now since all backends have an optimizer by that name.

## Type of change

## Tests

Included is a test in `test_serialization.py` that tests saving/loading QKeras and QONNX models. These cover serialization of most types and quantizers that can appear in a model, but obviously not all possible layers. Maybe a more thorough test would be to extend most existing tests to save and load a model and then continue working with a loaded model. But I'll leave that to a future PR.

## Checklist

I've done all the usual checks prior to opening this PR.


jmitrevs:
Even though the running time exceeded the limits, there were failures in test_serialization beforehand.

JanFSchulte:
> Even though the running time exceeded the limits, there were failures in test_serialization beforehand.

Indeed. I am just rerunning the tests to see which part is taking so long, and the QONNX test runs really fast, so it is the serialization test itself that is very slow.

test/pytest/test_serialization.py
```diff
@@ -0,0 +1,105 @@
...
+    serialize_model(hls_model, out_dir / 'qonnx_model.fml')
+    hls_model_clone = deserialize_model(out_dir / 'qonnx_model.fml')
+    hls_model_clone.config.config['OutputDir'] = str(out_dir / 'clone')
+    hls_model_clone.compile()
+    y_clone = hls_model_clone.predict(X)
+
+    np.testing.assert_equal(y_original, y_clone)
+
+
+@pytest.mark.parametrize('backend', ['Vitis'])  # Disabling OneAPI for now excessive run time
```
bo3z, line 69: OneAPI is enabled in the above QKeras test. Is it faster with QKeras than with QKeras or should the above one be also commented out?

hls4ml/model/graph.py
```diff
@@ -307,8 +308,90 @@ def _parse_hls_config(self):
...
+    def serialize(self):
+        state = {}
 
-class ModelGraph:
+        config = self.config.copy()
+        config.pop('KerasModel', None)
+        config.pop('OnnxModel', None)
+        config.pop('PytorchModel', None)
+
+        # Much of this may not be needed and is already in 'config' dict but is kept here to be sure
```
bo3z, line 319: This seems a bit error-prone, because if there's a new attribute added to HLS config and it's not reflected in these two functions, it could fail later on. I guess there are three ways to handle this:
- As the comment suggests, assume most (all?) of the variables are reflected in config, so the others are simply fail-safe mechanisms... do we know which of this variables aren't directly stored in config?
- Find a way to implicitly iterate through all the internal member variables of HLSConfig and store them to state. This seems like something that's doable in Python
- Ignore for now, as it doesn't seem the HLS config will change a lot in the (near) future.

hls4ml/model/graph.py
```diff
@@ -318,46 +401,57 @@ class ModelGraph:
...
+        self._applied_flows = []  # keep track of the applied flows
+        self.index = 0
+        self.output_vars = {}
+        self._top_function_lib = None
 
-        # keep track of the applied flows
-        self._applied_flows = []
+    @classmethod
+    def from_layer_list(cls, config_dict, layer_list, inputs=None, outputs=None):
+        def _find_output_variable_names(layer_list, layer_names):
```
bo3z, line 416: Just a personal preference - but I am not the biggest fan of functions inside of functions unless absolutely necessary. Is there a reason this function got moved?

hls4ml/model/quantizers.py
```diff
@@ -76,6 +90,10 @@ def __call__(self, data):
         ones = np.ones_like(data)
         return np.where(data > 0.5, ones, np.where(data <= -0.5, -ones, zeros))
 
+    def serialize_state(self):
+        state = {}
```
bo3z, line 94: Minor, but not sure if I am missing something obvious here: the BinaryQuantizer has a state but the TernaryQuantizer doesn't? In my understanding of the serialization, neither should have the state? Or if yes, both?

hls4ml/utils/serialization.py
```diff
@@ -0,0 +1,225 @@
...
+
+import numpy as np
+
+from hls4ml.model.graph import HLSConfig, ModelGraph
+from hls4ml.model.optimizer import optimize_model
+
+from .._version import version
+
+
+def serialize_model(model, file_path):
```
bo3z, line 16: Could we add a short docstring to this function?  Could be a copy of the docs, but still nice to have since it's a user-facing function.

bo3z (changes requested):
First round of reviews of the saving/loading infrastructure of hls4ml models. Generally, looks very good - most of the comments are minor with some questions for my better understanding. Two high-level questions / comments, for here as well:
1. Would it be better to name the methods save_model and load_model? This is closer to what Keras does and maybe a bit clearer to most users?
2. At what stage can a model be saved? After conversion but before calling .compile(...) or also after compilation? Most of the examples/tests I've seen in this PR are before compilation. Generally, it would be useful to be able to save/load the model after compilation (e.g. I converted the model on my local machine, compiled it to verify accuracy with predict(...) and then saved it, to synthesize it on a remote node). This is trickier though and can open up non-clear issues....e.g, if the CPU ISA is different between nodes the comiled hls4ml library is useless. But also useful, since compilation can sometimes take long (especially with lots of code generation).

test/pytest/test_serialization.py
vloncar: This is actually due to a bug in oneAPI that causes very long running time at 100% CPU usage due to context switches between threads. Reported to Intel.

test/pytest/test_serialization.py
```diff
@@ -0,0 +1,105 @@
...
+from tensorflow.keras.layers import Flatten, Input
+from tensorflow.keras.models import Sequential
+
+import hls4ml
+from hls4ml.utils.serialization import deserialize_model, serialize_model
+
+test_root_path = Path(__file__).parent
+example_model_path = (test_root_path / '../../example-models').resolve()
+
+
```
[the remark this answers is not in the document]
vloncar: If you wish, but will be the same result. Ultimately we're serializing the ModelGraph, where it originated from should not matter much. I added QONNX just to be sure more advanced models are working

hls4ml/model/graph.py
vloncar: Another one of those cases of "if only we had better config infrastructure" :smile: 

The problem we have is that there's the initial config dict, and the current config class. Then there may be code that uses one or the other, so we need to save both. Your second solution is error prone too, as it would assume we know how to handle all possible values of internal members so we can save them. 

In another project we're investigating Pydantic for for config schemas, maybe we can apply that to hls4ml in the future. For now I feel we could maybe make it clear to future developers that if they add new internal members they need to also follow up with serialization functions.

hls4ml/model/graph.py
vloncar: No other function uses it and it doesn't depend on `self`. If it was separate it would also have to be a `classmethod`, but it felt weird to have a class method intended only for internal use. I actually considered moving it to a separate utility file, but that also felt like too much given that it really is used only once.

hls4ml/model/graph.py
```diff
@@ -890,3 +986,25 @@ def build(self, **kwargs):
...
+            'index': self.index,
+            'applied_flows': applied_flows,
+        }
+
+        return state
+
+    @classmethod
+    def deserialize(cls, state):
+        raise Exception(
+            '{cls.__name__} is not intended to be deserialized directly. Use {cls.__name__}.from_saved_state instead.'
```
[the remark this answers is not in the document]
vloncar: indeed it does

hls4ml/model/quantizers.py
vloncar: This is a weird one... The "state" here is simply the choice of the number of bits used (1 or 2). For ternary we will always use 2 bits so there's no state to save. State is something you will need to pass to `__init__(...)` to recreate the object.

hls4ml/model/types.py
```diff
@@ -588,5 +731,16 @@ def __init__(self, code):
     def __str__(self):
         return str(self.code)
 
+    def serialize_class_name(self):
+        cls = self.__class__
```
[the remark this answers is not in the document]
vloncar: It's different. The wrangling of the name is needed because types are created dynamically (belonging to a backend). `Source` isn't one of them

hls4ml/utils/serialization.py
vloncar: Sure I'll do that

vloncar:
For the general comments:
1. Can be done. I actually tried not to copy Keras naming convention, but if that's what people prefer, I'm fine with renaming the function
2. You can call it afterwards. We save the IR, not what the IR ultimately generates (writes as a project). But we ensure that what is written is the same whether the model was serialized or not. `write()` doesn't change the internal state, so you're safe to call it many times at whichever point you want. `compile()` will call `write()`, then compile to create the `.so` and link it. This `.so` and the subsequent link to python runtime cannot be saved as they are not portable at all. Even on the same CPU type, but different OS family. For example on a desktop/laptop people would use Ubuntu or something else modern, and then move it to server for synthesis which is likely to run RHEL or clone. The compiled `.so` won't work and will need recompilation. Luckily you can be sure that when you call `write()`, `compile()` or `build()` you that first stage of writing the project to disk will result in the the same thing as the original model.

vloncar:
Now also includes an option to "convert" from an existing project written on disk which we can argue is a form of model loading. Adds a function `convert_from_existing_project()` that returns a special type of `ModelGraph`  where only the `compile()`, `predict()` and `build()` can run. This is useful if you want to use python integration after the project already exists. I used this "feature" for a while for development, as it allows one to generate an HLS project, manually make changes to it, and then go back to python to test stuff with `predict()` without the use of C++ testbench (which let's be honest, is a cumbersome way to test out the project). The person that asked for this to be an official feature uses it for testing in a separate environment that doesn't have ML frameworks

JanFSchulte:
That looks pretty cool! I'm just wondering if `convert_from_existing_project()` is the best name for the function, as no actual conversion is happening IIUC this is not a way to modify the project, just to be able to re-run these functions. Maybe `load_existing_project()` would be a bit more clear on the scope.

vloncar:
I didn't know where to put it, in converters or in utilities. If it is in converter, then it should be a convert_from. Also not sure where to put the `FilesystemModelGraph` class definition either. Feels like a very weird appendage currently in `graph.py`. We need to converge on the naming of functions in this PR anyway, so let's come up with a solution to all of this. Otherwise the PR is ready.

jmitrevs:
Should we maybe rename `serialize_model` to `save_model` and `convert_from_existing_project` `load_model`? Or something similar? One could also potentially have `hls_model.save(filename)` too.

vloncar:
Currently there is `serialize()`/`deserialize()` that work with a specific format that contains all information about a model in a `.fml` file. Then there is `convert_from_existing_project()` which creates a limited graph from existing project created with `write()` (full reconstruction is not possible). Calling that feature `load_model()` doesn't feel right. What would we do with `deserialize()` in that case?

I thought about renaming the main functions as `export()`/`import()` but it still feels that `load_model` for existing project is out of place as there's no equivalent `save()` or `save_model()`.

JanFSchulte:
I like `export()` and `import()` for the proper saving/loading of models. For the other case, I'd still go with something like `load_project()` to make sure that it allows to interact with an HLS project and not an hls4ml model. This specific functionality, including the `FilesystemModelGraph` could live in a `hls4ml/utils/projects.py`, or similar.

vloncar:
I simplified the API a bit: Saving is done with `ModelGraph.save()` (as suggested by Jovan), loading with `hls4ml.converters.load_saved_model()`. The existing project can be linked with `hls4ml.converters.link_existing_project()`. That way all user-facing model-creation calls are in the same place and all model manipulation calls are on the model object, no need for callinfg functions from other parts of the API.

JanFSchulte (approved):

bo3z (approved):

PR #1161, opened 2025-01-07
Bug fixes for channel-last conversions in pytorch

JanFSchulte:
## Description

Tensors were not correctly transposed back before reshape layers in case the `internal` option was used for the channels last conversion in pytorch, which is fixed in this PR. 

Also, 3D transposes would be inserted into the model, which in the case of `io_stream` leads to failures to compile since we don't have an implementation for this case. This is now caught and we exit with a meaningful error message. 

Finally, one default setting in the pytorch config is brought in line with the documentation. 

## Type of change

## Tests

Tested with a MNIST config Jovan got from Siemens, verified that these changes fix the issues observed there. 

## Checklist


jmitrevs (approved):
Assuming all the tests finish successfully we should merge this.

Issue #1166, opened 2025-01-20
Datatype does not match in HLS converted quantized lstm from QKeras


JiaMingLin:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x] If there are steps to reproduce the problem, make sure to write them down below.
- [ ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
I try to convert quantized LSTM from QKeras via HLS4ML, but the `hls_model.compile()` would fail.

## Details
* Tensorflow version: 2.14.1
* QKeras version: 0.9.0
* HLS4ML version: updated from the main, installed by `pip install -e .`

The error message is as following
```
2025-01-20 20:25:43.718528: I tensorflow/compiler/xla/stream_executor/cuda/cuda_gpu_executor.cc:894] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero. See more at https://github.com/torvalds/linux/blob/v6.0/Documentation/ABI/testing/sysfs-bus-pci#L344-L355
2025-01-20 20:25:43.718666: I tensorflow/compiler/xla/stream_executor/cuda/cuda_gpu_executor.cc:894] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero. See more at https://github.com/torvalds/linux/blob/v6.0/Documentation/ABI/testing/sysfs-bus-pci#L344-L355
2025-01-20 20:25:43.734997: W tensorflow/core/common_runtime/gpu/gpu_device.cc:2211] Cannot dlopen some GPU libraries. Please make sure the missing libraries mentioned above are installed properly if you would like to use GPU. Follow the guide at https://www.tensorflow.org/install/gpu for how to download and setup the required libraries for your platform.
Skipping registering GPU devices...
/home/robin/miniconda3/envs/tf214/lib/python3.10/site-packages/keras/src/constraints.py:365: UserWarning: The `keras.constraints.serialize()` API should only be used for objects of type `keras.constraints.Constraint`. Found an instance of type <class 'qkeras.quantizers.quantized_bits'>, which may lead to improper serialization.
  warnings.warn(
Interpreting Sequential
Topology:
[...]

PR #1171, opened 2025-01-22
Fixes for quantised RNNs in data type inconsistencies

bo3z:
# Description

> :memo: Various fixes for data type inconsistencies for quantised LSTM / GRU (see below for detailed explanation)
> First discovered by issue #1166; further showed much more fundamental issues with the RNN layer implementations with Vivado / Vitis backend

## Type of change

## Tests
TODO: Add additional tests; so far it has only been confirmed that this fixes issue #1166. However, our current tests don't have different data-types between weights and recurrent weights

## Checklist

## Description
1. There was a mismatch between the parsed quantizer from QKeras and the weight type & name for recurrent weights, causing the recurrent weights to always default to the global default precision. This has been fixed in: hls4ml/converters/keras/qkeras.py. This was the original error discovered by #1166 
2. However, this error has further shown that the HLS implementation assumes that weight and reccurent weight (as well as bias and reccurent bias) have the same datatype, which is not necessarily true. QKeras let's us have different quantisers for weight and reccurent weights (although recurrent bias doesn't have a custom quantiser). For this the HLS template has been modified and the corresponding Vivado / Vitis backend pass.
3. The final issue discovered by this is the inconsistency in activation data types. Namely, a lot of the activation calls had incorrect templates (defaulting to `data_T` or `weight_T`), when in fact they should be `accum_T`. I traced all the activations and they should now match the variables that are being passed by them. However, this would be worth double checking by someone else.

**Additional TODOs:**
- Verify whether these issues are present for oneAPI and Catapult. I had some mysterious issues when compiling the code, I guess my environment is not up to date to use these backends. If someone who has a bit more expertise with oneAPI and Catapult could check this PR and the associated issue out, it would be great. 
- Add a test that confirms the fix works in "heterogeneous" quantisation.


JanFSchulte:
pre-commit.ci autofix

JanFSchulte:
Looks good to me. Not sure we want to wait for someone to check oneAPI and Catapult before merging this, so with the added test it might be good to go already.

vloncar:
Looks good to me, we can follow up with Intel/Catapult PR. Let's see what the tests say before merging.

JanFSchulte (approved):

PR #1205, opened 2025-02-20
Adjust model output if last node is removed

JanFSchulte:
If an optimizer removes the last node in a model, currently the model output names are not adjusted but the output variable is removed from the list of output vars nevertheless, resulting in crashes. 

Fixed here by checking if we are removing the output node and if so, setting the model output to be the previous node. 

## Type of change

## Tests

Parsing a small brevitas model with a `QuantReLu` as the last layer via QONNX currently fails but parses correctly with this fix. 

## Checklist

### DIFF
--- a/hls4ml/backends/fpga/fpga_backend.py
+++ b/hls4ml/backends/fpga/fpga_backend.py
@@ -755,7 +755,7 @@ class FPGABackend(Backend):
 
         generated_code = (
             "template<class data_T, typename CONFIG_T>\n"
-            "class fill_buffer_{index} : public FillConv1DBuffer<data_T, CONFIG_T> {{\n"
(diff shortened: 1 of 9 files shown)


calad0i:
Line 550 was meant to cover this issue, could you check why that line is not triggered? Does the node has more than more inputs, or its name is not in outputs?

JanFSchulte:
In this case, the name is not in the outputs, the output has a dedicated name `global_out` that is different from the node name. This is a model coming from brevitas through QONNX and things then get modified quite excessively inside hlsm4l as well, so overall we end up with this case that is not caught otherwise.

calad0i:
As far as I understand, hls4ml `ModelGraph` references layer output by name, and we assume each layer to have exactly one output for now. As I cannot find the `global_out` in the main branch right now, could you point me to the relavent.

jmitrevs (approved):

PR #1240, opened 2025-03-21
☢️ Libero backend for rad-hard PolarFire line of FPGAs

vloncar:
# Description

Initial (still a draft) implementation of a backend for MicroChip PolarFire line of FPGAs. The suite of tools is called "Libero" so that's used for the name of the backend too. The HLS compiler itself is "SmartHLS", based on the previously open-source LegUp HLS. The syntax is similar to Vivado/Vitis HLS, allowing for simple porting, however synthesis results are different and will need some fine-tuning to get working well.

The backend is currently not feature complete. Only Dense layers are supported. Given the nature of applications we likely won't aim to have full feature set of other backends. The initial goal to have both strategies in io_parallel, we'll explore streaming implementation if required.

## Type of change

## Tests

Currently no tests. Will be added prior to marking the PR ready for review.

## Checklist

(still some left)

### DIFF
--- a/hls4ml/backends/__init__.py
+++ b/hls4ml/backends/__init__.py
@@ -1,5 +1,6 @@
 from hls4ml.backends.backend import Backend, get_available_backends, get_backend, register_backend  # noqa: F401
 from hls4ml.backends.fpga.fpga_backend import FPGABackend  # noqa: F401
+from hls4ml.backends.libero.libero_backend import LiberoBackend
 from hls4ml.backends.oneapi.oneapi_backend import OneAPIBackend
 from hls4ml.backends.quartus.quartus_backend import QuartusBackend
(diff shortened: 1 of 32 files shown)


vloncar:
This is now ready for review/merge.

The latest feature is the support for `compile()` but this had to be implemented in a hacky way. The `ap_fixpt` types require the SmartHLS compiler (based on gcc 5) so we have to use the `shls` compiler for this. Problem is this compiler is controlled via a Makefile that isn't really menat for building shared libraries, rather for making binaries. Therefore we produce two makefiles, and switch the active one depending on if we do `build()` or `compile()`. Furthermore, the types are not header-only (i.e., stand-alone) and require linking with the `libhls_x86` library that is shipped as part of the installation of Libero, so we need a path to it.

PR #1019, opened 2024-06-07
Add support for parsing simple brevitas layers as part of pytorch models

JanFSchulte:
This PR adds support for the parsing of all brevitas layers supported by hls4ml to the pytorch parser. To this end, this PR adds a new quantizer which only propagates the desired precision to the hsl4ml model, as brevitas already provides the quantized tensors which we pick up directly. For models with supported layers, this now hooks into the bit-exact workflow as well to give most precise results. Accuracy is good for the other cases as well. 

If we ever decide to merge this, needs a pass to de-claudify comments and the documentation. 

## Type of change

For a new feature or function, please create an issue first to discuss it
with us before submitting a pull request.

Note: Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Tests

Tested locally with simple model and added pytests. 

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [x] I have added tests that prove my fix is effective or that my feature works.

PR #1247, opened 2025-03-26
Simple PyTorch extension API

vloncar:
# Description

The equivalent of Keras extension API for PyTorch. The least hacky way of doing it, but only limited to custom modules. The custom modules should extend `hls4ml.utils.torch.HLS4MLModule` (itself a `torch.nn.Module`). I borrowed the `CustomFXTracer` change from #1019 and moved it to a different location, since the file contains `HLS4MLModule` class intended to be used by the user.

## Type of change

## Tests

Test is included in `test_extensions_pytorch.py`

## Checklist

### DIFF
--- a/test/pytest/test_extensions_pytorch.py
+++ b/test/pytest/test_extensions_pytorch.py
@@ -43,8 +43,8 @@ class RemoveDuplicateReverse(hls4ml.model.optimizer.OptimizerPass):
         first = node.get_input_node()
         second = node
 
-        model.remove_node(first, rewire=True)
-        model.remove_node(second, rewire=True)
+        model.remove_node(first)
+        model.remove_node(second)
         return True
 
 
--- a/test/pytest/test_extensions_pytorch.py
+++ b/test/pytest/test_extensions_pytorch.py
@@ -1,0 +1,192 @@
+from pathlib import Path
+
+import numpy as np
+import pytest
+import torch
+
+import hls4ml
+import hls4ml.utils.torch
+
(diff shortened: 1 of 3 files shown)


test/pytest/test_extensions_pytorch.py
```diff
@@ -0,0 +1,192 @@
...
+    '''OptimizerPass to remove consecutive HReverse layers.'''
+
+    def match(self, node):
+        return isinstance(node, HReverse) and isinstance(node.get_input_node(), HReverse)
+
+    def transform(self, model, node):
+        first = node.get_input_node()
+        second = node
+
+        model.remove_node(first, rewire=True)
```
JanFSchulte, line 46: I think the `rewire` parameter doesn't do anything anymore, might be good to just not include it here if people end up using this as an example.

JanFSchulte (changes requested):

test/pytest/test_extensions_pytorch.py
vloncar: I have probably copy/pasted that line from one of the other optimizers. Should I replace all instances of this? I see there's a few around

JanFSchulte: Might be good to clean this up in general. I'm not sure why we have this around at all anymore, since the parameter is really just a dummy according to https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/graph.py#L520-L537

JanFSchulte (approved):

PR #1257, opened 2025-04-02
pytest based synthesis tests

marco66colombo:
# Description

This PR introduces **synthesis tests** into the `pytest` framework for `hls4ml`.

The goal of this change is to automate the validation of HLS synthesis reports against predefined baselines to ensure stability and correctness over time, using pytest.

### Main Changes

- **New submodule** in `tests/pytests` called `baselines`, which is a Git submodule pointing to an external repository containing synthesis baselines.  
  - The structure of that repository follows the format:  
    `backend_tool/version/test_artifacts.json`
  - Note: Currently, this is a temporary personal repo and **must be replaced** by a proper one under the organization before merging.

- **Baseline management**:
  - Baselines are organized by backend (e.g., Vivado, Vitis), tool version (e.g., 2020.1, 2023.1), and test case.
  - A **manual and recurrent update policy** for these baselines should be defined and enforced.

- **New `conftest.py`** file to manage shared pytest fixtures:
  - Introduces the `synthesis_config` fixture, which provides:
    - Whether to run synthesis tests (`RUN_SYNTHESIS` environment variable).
    - Tool versions (via env vars, e.g., `VIVADO_VERSION`, defaulting to `2023.1`).
    - Build arguments (backend-dependent) for `model.build(...)`.

- **Synthesis logic** centralized in `synthesis_helpers.py`:
  - Main method: `test_synthesis`.
  - If `RUN_SYNTHESIS=true`, runs synthesis and compares the report to baseline.
  - Skips test if:
    - Tool is not available,
    - Baseline is missing
    - Currently, in the above cases, the whole test is skipped, so it might has to change in future. It is also true that this method is called after any other assert in the test case, so it would still fail in case of a wrong assert in the code before this method is called.

- **Tolerances** for report values can be specified using `get_tolerance`.
  - Currently, the tolerance values are placeholders and need to be properly defined.

- **CI integration**:
  - Synthesis reports are saved as JSON and collected as artifacts via GitLab CI (`ci-template.yml` updated accordingly).

- **Current test coverage**:
  - Only `test_keras_api.py` includes synthesis tests at the moment.
  - Easily extendable to other tests:
    - Add `baseline_file_name` to the test,
    - Call `test_synthesis()`.

- **OneAPI backend is currently skipped** due to missing synthesis reports support.

---

## Type of change

- [x] Other: improvment of testing infrastrucutre using pytest

---

## Tests

Synthesis tests are run conditionally and compared against versioned baselines if `RUN_SYNTHESIS=true` is set in the environment.

## Checklist

- [ ] I have added tests that prove my feature works.


jmitrevs:
Can we test this to actually run SYNTHESIS tests?

marco66colombo:
yes, but if the env var 'RUN_SYNTHESIS' is not set, the default is False.
I can add a line in the ci-template.yml and export RUN_SYNTHESIS=True

test/pytest/ci-template.yml
```diff
@@ -9,8 +9,15 @@
     - source /opt/intel/oneapi/setvars.sh --force
     - git config --global --add safe.directory /builds/fastmachinelearning/hls4ml
     - git submodule update --init --recursive hls4ml/templates/catapult/
+    - git submodule update --init --recursive test/pytest/
     - if [ $EXAMPLEMODEL == 1 ]; then git submodule update --init example-models; fi
     - pip install .[testing,sr,optimization]
+    - sudo yum install -y libidn
```
vloncar, line 15: What causes this requirement?

marco66colombo: I was trying to fix an error that is present only while running the code in the GitLab runner and not if I run the docker image locally.
It happens with oneAPI backend.
It was a test, but it didn't fix the issue and I'll remove it in the next commit. 
Failed gitlab job: https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/54542573

marco66colombo:
```$ which cmake
/opt/Xilinx/Vitis/2020.1/tps/lnx64/cmake-3.3.2/bin/cmake
$ cmake --version
cmake: error while loading shared libraries: libidn.so.11: cannot open shared object file: No such file or directory
```

I found that the difference is that the `cmake` comes from the Vitis installation with Vitis2020.1. 
This is what causes the error.

marco66colombo:
Do you all agree with the approach of storing all the baselines in a dedicated GitHub repo, and add it as a module under `test/pytest/baselines` in `hls4ml`?
If yes, we need to create the repository within Fastmachinelearning and then I can upload as baselines the artifacts generated by the synthesis tests.

jmitrevs:
I generally agree about storing the baselines in a dedicated repo. Is CERN gitlab or regular github preferred? We create the testing containers in gitlab. I don't really have a preference.

JanFSchulte:
Thanks, this looks very nice to me! From the test logs, it seems to have triggered the synthesis tests successfully and the code looks good to me, so I think this is basically ready for merge. I just have a few questions for the way forward:

- What are the plans to extend this to more tests? How much can we realistically do with the resources we have available? 
- As you said, the thresholds are preliminary, how are these going to be tuned? 
- What's preventing us from testing Quartus right now?

marco66colombo:
@JanFSchulte thanks for the review and the follow-up questions.

- Extending coverage: I’d like to bring synthesis to additional pytest files, but the pace depends on the CERN GitLab runner capacity. I’ll check the available concurrency/quotas and follow up with a better answer for this question. If possible, it would be great to hear which tests the community considers highest priority for synthesis so we can focus on those first.

- Tuning thresholds: the plan would be to sync with the contributors who regularly use the Vivado/Vitis/oneAPI flows so we can calibrate them against real reports. I’m happy to incorporate suggestions as soon as we have them, since I’m not confident about the best values for every metric, so any feedback is very appreciated.

- Balancing batches:  The longest jobs come both from parametrized tests that explode into many slow combinations and from individual cases that are expensive to synthesize. To tackle that, I’m trimming the heaviest branches directly in the tests with small if guards around run_synthesis_test, and on the CI side this PR introduces SPLIT_BY_TEST_CASE so we can shard the slowest files. I had been experimenting with more advanced batching policies, but I intentionally shipped this simpler version first so we can gather experience and iterate based on the feedbacks. I would like to discuss more on this.

PR #1273, opened 2025-04-15
Add support for einsum operation to pytorch parser (requires 1116)

JanFSchulte:
Builds on Chang's keras v3 PR #1116 and exposes the new einsum implementation through the pytorch parser. pytorch doesn't have an an equivalent to `EinsumDense` but allowing the use of einsum operations in some custom model would still be useful. 

## Type of change

## Tests

Added 2 use cases (outer product and batch matrix multiplication to the pytests, works without issues. 

## Checklist

### DIFF
--- a/hls4ml/converters/pytorch/core.py
+++ b/hls4ml/converters/pytorch/core.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 from hls4ml.converters.pytorch_to_hls import pytorch_handler
+from hls4ml.utils.einsum_utils import _validate_einsum_expr
 
 
 @pytorch_handler('Constant')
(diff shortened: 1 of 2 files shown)


hls4ml/converters/pytorch/core.py
```diff
@@ -157,3 +157,35 @@ def parse_batchnorm_layer(operation, layer_name, input_names, input_shapes, node
...
+    input_shapes_tmp[1][0] = 1
+    layer['inp0_shape'] = tuple(input_shapes_tmp[0])
+    layer['inp1_shape'] = tuple(input_shapes_tmp[1])
+
+    # Run einsum to infer output shape
+    import torch
+
+    a = torch.randn(input_shapes_tmp[0])
+    b = torch.randn(input_shapes_tmp[1])
+    layer['out_shape'] = tuple(torch.einsum(layer['equation'], a, b).shape)
```
calad0i, line 189: It is preferred to use `hls4ml.utils.einsum_utils._validate_einsum_expr` here, something like:

```python
inp_shape0 = input_shapes_tmp[0][1:]
inp_shape1 = input_shapes_tmp[1][1:]

layer['equation'], layer['out_shape'] = _validate_einsum_expr(node.args[0], inp_shape0, inp_shape1
```

In this way wildcard dimensions (`...`) will be fixed to indices and explicit torch call will not be required.
Otherwise all looks good.

PR #1284, opened 2025-04-29
Round inputs for dense unrolled RNN tests to make pytests more stable

JanFSchulte:
We are still having problems with numerical stability of RNNs in the pytests for dense unrolled. This PR aims to fix that by implementing the rounding @jmitrevs added in #1215 for these tests as well. As I haven't been able to reproduce the failure locally, I couldn't check if this actually fixes it

## Type of change

## Checklist

### DIFF
--- a/test/pytest/test_dense_unrolled.py
+++ b/test/pytest/test_dense_unrolled.py
@@ -107,6 +107,7 @@ def test_resource_unrolled_rnn(rnn_layer, backend, io_type, static, reuse_factor
     # Subtract 0.5 to include negative values
     input_shape = (12, 8)
     X = np.random.rand(50, *input_shape) - 0.5
+    X = np.round(X * 2**16) * 2**-16  # make it exact ap_fixed<32,16>
 
     layer_name = rnn_layer.__name__.lower()
     keras_model = Sequential()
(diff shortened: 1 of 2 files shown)


test/pytest/test_dense_unrolled.py
```diff
@@ -107,6 +107,7 @@ def test_resource_unrolled_rnn(rnn_layer, backend, io_type, static, reuse_factor
     # Subtract 0.5 to include negative values
     input_shape = (12, 8)
     X = np.random.rand(50, *input_shape) - 0.5
+    X = np.round(X * 2**16) * 2**-16  # make it exact ap_fixed<32,16>
```
calad0i, line 110: This is indeed `ap_fixed<17, 0>`. Though, as you mentioned that the issue is more prevalent with unrolled dense, is there anything compromising bit-exactness between implementations?

JanFSchulte: I think it might be the RNN, or maybe just LSTM, in general. Jovan did this change to get the tests to behave better in the pytorch parser case. It seemed to have worked there, so I'd be in favor of merging this now so get more meaningful test results, and look into why this is such an issue later.

JanFSchulte:
This issue with the RNN precision is still causing pytests to fail in other PRs, btw. So I would still be in favor of merging this so that we don't get as many spurious test failures.

PR #1291, opened 2025-05-05
more general build_lib script

calad0i:
# Description

`$OSTYPE` is not a reliable source for determining the platform. On OpenSUSE 20250503, bash returns `"linux"` instead of `"linux-gnu"` and the script failed without `-fPIC`.

The updated script checks if  `-fno-gnu-unique` is supported, and appends it if so.

## Type of change

## Tests

Triggers on specific platforms.

### DIFF
--- a/hls4ml/templates/catapult/build_lib.sh
+++ b/hls4ml/templates/catapult/build_lib.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 CC=g++
-if [[ "$OSTYPE" == "linux-gnu" ]]; then
-    CFLAGS="-O3 -fPIC -std=c++11 -fno-gnu-unique"
-elif [[ "$OSTYPE" == "linux"* ]]; then
-    CFLAGS="-O3 -fPIC -std=c++11 -fno-gnu-unique -Wno-pragmas"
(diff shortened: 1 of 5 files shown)


vloncar:
Instead of replacing `"$OSTYPE" == "linux-gnu"` with `"$OSTYPE" == "linux"*` (a few characters change) the suggestion is to invoke the compiler to do nothing just to see if it will fail and if it doesn't then add a cflag... bit of an overkill, no? :smile:

jmitrevs:
As an aside, should we move up from C++11? Vitis HLS I think uses C++14, Intel uses C++17. Probably makes little difference, though.

vloncar:
@jmitrevs Only if we split the build script over vivado/vitis (vivado compiler will say c++14 doesn't exist)

calad0i:
Using this way as `$OSTYPE` seems to unstable depending on the shell versison, and test if the compiler handles `-fno-gnu-unique` would always work as long as a compiler exists (and doesn't really take time).

jmitrevs:
In general I think this fix is good and would be in favor of merging it. I just put please test as a sanity check to make sure it doesn't break things.

Issue #1287, opened 2025-05-01
SeLU activation range?


jmitrevs:
## Prerequisites
Please make sure to check off these prerequisites before submitting a bug report.
- [x] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
- [x] Check that the issue hasn't already been reported, by checking the currently open issues.
- [x] If there are steps to reproduce the problem, make sure to write them down below.
- [x] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary

The SeLU activation uses `(typename data_T::value_type)1.0507009873554804934193349852946`.  This is dangerous because there are no guarantees that the data type has a range covering 1.05. What if the data range is -0.25 to 0.25 or so? The dimensions should be decoupled. One possibility is to just hardcode, say `ap_ufixed<16, 1, AP_RND>` or something.

PR #1298, opened 2025-05-12
Fix SeLU lambda precision (Follow-up to #1287)

valerioedu:
- Replaced literal cast 1.050700987… → res_T with a static const ap_fixed<16,1> lambda, preserving range and ~1.5 × 10⁻² LSB precision regardless of user-chosen res_T.

- Removed redundant datareg scope; now a single per-element branch: if (x ≥ 0) y = λ · x
else y = selu_table[idx] (with index clamped to [0,N-1]).

- Guard against negative-index underflow; behaviour for <0 inputs unchanged.


vloncar:
Can you add a test case that demonstrates that this is indeed working as expected and improves upon the current implementation? Should be straightforward

vloncar:
Look under `tests/pytest`

jmitrevs:
There is a selu test in test_activations, so if the standard pytests pass (which I just started), I would suggest we merge this.

PR #1299, opened 2025-05-14
Support cloning up to 7 times

vloncar:
# Description

@makoeppel from ETH needs cloning to support up to ~6~ actually 7 clones for the Inception model for some project with PSI people.

## Type of change

## Tests

Extended the test in `test_stream_clone.py` to test all new cases.

## Checklist

- [x] I have checked all the items in the checklist


calad0i:
Should we just go with codegen at this point 😂...?

vloncar:
Maybe if we actually had code-gen and not just a string template manipulation... :wink:

calad0i (approved):
:joy:

PR #1302, opened 2025-05-28
skip oneapi test if icpx doesn't exist

calad0i:
# Description

skip oneapi test if icpx doesn't exist

# Type of change

- [x] Other (Specify)

## Tests

Not needed

## Checklist

- [ ] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).

### DIFF
--- a/hls4ml/backends/oneapi/oneapi_backend.py
+++ b/hls4ml/backends/oneapi/oneapi_backend.py
@@ -175,6 +175,12 @@ class OneAPIBackend(FPGABackend):
         try:
             subprocess.run('which icpx', shell=True, cwd=builddir, check=True)
         except subprocess.CalledProcessError:
+            try:
+                import pytest
+
+                pytest.skip('icpx not present')
+            except ImportError:
+                pass
             raise RuntimeError('Could not find icpx. Please configure oneAPI appropriately')
         subprocess.run('cmake ..', shell=True, cwd=builddir, check=True)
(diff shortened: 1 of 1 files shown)


jmitrevs:
Although I am in general supportive of this, I do worry about silent failures. We should make sure that the CI makes sure the tests aren't accidentally skipped. The CI should fail if they are skipped.

calad0i:
Most likely cmake directs the compiler to icpx in this case. Actually the skips are added to the main codebase instead of tests, so every time .compile() is called with oneapi backend, it will try to skip. Outside test, this skip is treated as an exception.
Will need to look closer if hooking into a fixture can work this way. A project scope fixture is only called once, and this will need to determine for each test if `'oneapi'` is used for any parameter... (and if one test targets oneapi specifically and doesn't use it as a parameter it will be missed).

hls4ml/backends/oneapi/oneapi_backend.py
```diff
@@ -175,6 +175,15 @@ def compile(self, model):
         try:
             subprocess.run('which icpx', shell=True, cwd=builddir, check=True)
         except subprocess.CalledProcessError:
+            try:
```
JanFSchulte, line 178: Do we still need this try - except block if the `pytest` import only happens when it's available?

PR #1307, opened 2025-06-08
Support for floating point types

vloncar:
# Description

Current hls4ml supports only arbitrary precision integer and fixed-point types. This PR adds support for floating point types. Floating point types are defined in `ap_float.h` and `ac_float.h` of the respective libraries and cover two distinct cases: IEEE floating-point standard (basically C/C++ types) and general floating-point implementation (any combination of mantissa and exponent). AC types library is more broad and offers more flexibility than AP types one. The PR covers both by introducing `FloatPrecisionType` for the general case (covered by the `ac_float`) and `StandardFloatPrecisionType` (for `ap_float<W,E>` and `ac_std_float<W,E>`). In principle one could cram everything in a single type but that makes it complicated to track what is the actual intended use, especially because of the 1-bit difference between AC and AP types.

To use, user can specify `float`, `double`, `half` and `bfloat16` as type and this will result in `StandardFloatTypePrecision` objects to be used and those C++ types used in the generated code. Note that `half` and `bfloat16` aren't supported out of the box and require the user to tweak the code to make it compile, as it is dependent on the compiler how these are exposed. If the user specifies `std_float<W,E>` the `ap_float<W,E>` and `ac_std_float<W,E>` will be generated using the same object. Finally, for full control of AC type, user can use `ac_float<W,I,E,Q>` which will use the `FloatPrecisionType` class and emit the corresponding type in code.

The PR is somewhat incomplete as there are numerous nuances of full support of the general case and the `half`/`bfloat16` but is a good starting point and is self-contained. The problem remains with AP types that don't have a public version of `ap_float.h` (we'll ask AMD about open-sourcing it), so if user uses a general floating-point type the local compilation with `compile()` won't work. In the future we can tackle the include issue (also for `half` and `bfloat16` on host compilers), as well as look into optimizations of algorithms (for example, using the accumulator type for the CMVM). The intention right now is not to make this a first-class supported feature, rather an exotic option for users who know what they want and are aware of the caveats and rough edges, but more crucially it allows us to avoid silly test failures due to bitwidth issues. This could be advertised as an experimental feature, but I see we're completely lacking any documentation on type setting, so that may come as a separate PR.

## Type of change

## Tests

Tests for parsing the new types has been added to `test_types.h`.

## Checklist

Yeah, yeah, I did all the things in the checklist.


vloncar:
I'll add ap_float.h and mapping to C++23 types. Converting to draft until I make the change

hls4ml/templates/vivado/firmware/defines.h
```diff
@@ -6,6 +6,9 @@
 #include "nnet_utils/nnet_types.h"
 #include <cstddef>
 #include <cstdio>
+#if __cplusplus > 202002L
```
vloncar, line 9: This check should be `#if __cplusplus >= 202302L` but since GCC doesn't fully support C++23 the constant is not set yet to that value, hence we check for "newer than C++20". Alternatively, we could check `__STDCPP_­BFLOAT16_­T__` and `__STDCPP_­FLOAT16_­T__` but these may also not be available even if the implementation is there...

JanFSchulte:
I can now synthesize a model in Vitis 2024.1 when I use `float` as the data type. For an `ap_float`, there are still some issues, this is the full error stack I'm getting. 

```
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:24:3: warning: use of this statement in a constexpr function is a C++14 extension [-Wc++14-extensions]
  if (val <= 0)
  ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:29:3: warning: multiple return statements in constexpr function is a C++14 extension [-Wc++14-extensions]
  return log2_ceil((val + 1) / 2) + 1;
  ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:25:5: note: previous return statement is here
    return std::numeric_limits<int>::lowest();
    ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:27:5: note: previous return statement is here
    return 0;
    ^
In file included from ../../../../myproject_test.cpp:1:
In file included from /home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/algorithm:61:
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:324:18: error: no viable overloaded '='
              *__result = *__first;
              ~~~~~~~~~ ^ ~~~~~~~~
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:386:22: note: in instantiation of function template specialization 'std::__copy_move<false, false, std::random_access_iterator_tag>::__copy_m<const float *, ap_float<32, 8> *>' requested here
                              _Category>::__copy_m(__first, __last, __result);
                                          ^
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:422:23: note: in instantiation of function template specialization 'std::__copy_move_a<false, const float *, ap_float<32, 8> *>' requested here
      return _OI(std::__copy_move_a<_IsMove>(std::__niter_base(__first),
                      ^
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:454:20: note: in instantiation of function template specialization 'std::__copy_move_a2<false, __gnu_cxx::__normal_iterator<const float *, std::vector<float>>, ap_float<32, 8> *>' requested here
      return (std::__copy_move_a2<__is_move_iterator<_II>::__value>
                   ^
../../../../firmware/nnet_utils/nnet_helpers.h:255:10: note: in instantiation of function template specialization 'std::copy<__gnu_cxx::__normal_iterator<const float *, std::vector<float>>, ap_float<32, 8> *>' requested here
    std::copy(in_begin, in_end, dst);
         ^
../../../../myproject_test.cpp:63:13: note: in instantiation of function template specialization 'nnet::copy_data<float, ap_float<32, 8>, 0UL, 300UL>' requested here
      nnet::copy_data<float, input_t, 0, N_INPUT_1_1*N_INPUT_2_1*N_INPUT_3_1>(in, x);
            ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:162:13: note: candidate function not viable: no known conversion from 'const float' to 'ap_float<32, 8>' for 1st argument
  ap_float& operator=(ap_float &&) = default;
            ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:163:13: note: candidate function not viable: no known conversion from 'const float' to 'const ap_float<32, 8>' for 1st argument
  ap_float& operator=(const ap_float &) = default;
            ^
In file included from ../../../../myproject_test.cpp:1:
In file included from /home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/algorithm:61:
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:754:11: error: no viable overloaded '='
        *__first = __tmp;
        ~~~~~~~~ ^ ~~~~~
/home/tools/Xilinx/Vitis_HLS/2024.1/tps/lnx64/gcc-8.3.0/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/bits/stl_algobase.h:789:23: note: in instantiation of function template specialization 'std::__fill_n_a<ap_float<32, 8> *, unsigned long, double>' requested here
      return _OI(std::__fill_n_a(std::__niter_base(__first), __n, __value));
                      ^
../../../../firmware/nnet_utils/nnet_helpers.h:304:79: note: in instantiation of function template specialization 'std::fill_n<ap_float<32, 8> *, unsigned long, double>' requested here
template <class data_T, size_t SIZE> void fill_zero(data_T data[SIZE]) { std::fill_n(data, SIZE, 0.); }
                                                                              ^
../../../../myproject_test.cpp:95:19: note: in instantiation of function template specialization 'nnet::fill_zero<ap_float<32, 8>, 300UL>' requested here
            nnet::fill_zero<input_t, N_INPUT_1_1*N_INPUT_2_1*N_INPUT_3_1>(x);
                  ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:162:13: note: candidate function not viable: no known conversion from 'const double' to 'ap_float<32, 8>' for 1st argument
  ap_float& operator=(ap_float &&) = default;
            ^
/home/tools/Xilinx/Vitis_HLS/2024.1/include/ap_float.h:163:13: note: candidate function not viable: no known conversion from 'const double' to 'const ap_float<32, 8>' for 1st argument
  ap_float& operator=(const ap_float &) = default;
            ^
2 warnings and 2 errors generated.
make: *** [csim.mk:87: obj/myproject_test.o] Error 1
```

test/pytest/test_types.py
```diff
@@ -80,8 +82,105 @@ def test_precision_type_creation(capsys):
...
+
+    evalprec = FPGABackend.convert_precision_string(prec_str)
+    assert isinstance(evalprec, StandardFloatPrecisionType)
+    assert evalprec.use_cpp_type == is_cpp and prec_str in str(evalprec)
+
+
+@pytest.mark.parametrize(
+    'prec_tuple',
+    [
+        # Should result in ap_float
```
jmitrevs, line 170: Is the comment correct?

PR #1317, opened 2025-06-24
fix merge templates

calad0i:
# Description
Allow merge template to take inputs with different precision and cast both to output precision; let 1/2 always be `ap_ufixed<1,0>/ac_fixed<1,0,false>`.


vloncar:
This looks good, but can you also provide an example of when it is needed? Like include a test case that would trigger it.

calad0i:
Added test for inp1_t != inp2_t. The test case for 0.5 type issue is not trivial to add (needs very low fractional bits, not easy to add for the current approx equal results test), but will be triggered in hgq2 tests anyway.

PR #1321, opened 2025-07-02
Purge dim name

calad0i:
# Description

Get rid of dim names in the whole repository, based on the DA PR.

## Type of Change

- [x] Cleaning up

## Tests

N/A

## Checklist

- [x] all


vloncar:
Now you're adding unrelated commits to this PR :smile:

calad0i:
> Now you're adding unrelated commits to this PR 😄

Let me move the distutil one to separate PR? It is deprecated and breaks on python 312 and 313.

JanFSchulte:
I think that was about the change to the bit-exact algorithm which probably needs to go the DA PR :-D

calad0i:
> The pytest error is fixed now, but there seems to be an error in the synthesis test now:
> 
> ```
> ../../../../myproject_test.cpp:15:20: error: 'N_OUT000' was not declared in this scope
> ```
> 
> I suspect the name gets malformed somehow, but from reading the code changes it's not obvious to me what's going wrong.
> 
> Otherwise I think this is ready for merge.

Caused by global search & replace at https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/writer/vivado_accelerator_writer.py#L314. I have no idea why it is written like this.

calad0i:
> I noticed that vivado_accelerator_writer has some lines like `line.replace(out.size_cpp(), 'N_OUT').replace(out.name, 'outputs').replace(out.type.name, 'output_axi_t')` and `newline = line.replace(out.size_cpp(), 'N_OUT').replace(out.type.name, 'output_axi_t')`. Could that be the reason? Thought that should only affect the accelerator backend.

`INFO: [HLS 200-10] Creating and opening solution '/var/jenkins_home/workspace/hls4ml_pr_1321/test/hls_prj/KERAS_3layer-backendVivadoAccelerator-boardpynq_z2-xc7z020clg400_1-c5-io_parallel-rf1-ap_fixed_16_6_-Resource-Vivado-2020.1/myproject_prj/solution1'.` The failure point is at vivado accelerator, so it is.

hls4ml/backends/catapult/passes/recurrent_templates.py
```diff
@@ -80,17 +80,19 @@ def __init__(self):
 
     def format(self, node):
         params = self._default_config_params(node)
+        in_0, in_1 = map(str, node.get_input_variable().shape[:2])
 
-        params['n_in'] = node.get_input_variable().dim_names[1]
-        params['n_sequence'] = node.get_input_variable().dim_names[0]
+        params['n_in'] = in_1
```
vloncar, line 85: instead of the style of calling these `in_0, in_1, out_0, out_1` etc, why not give them meaningful names like they will be used later (`n_in, n_sequence, n_state` etc). Across all `*_template.py` files.

hls4ml/backends/catapult/passes/recurrent_templates.py
```diff
@@ -80,17 +80,19 @@ def __init__(self):
 
     def format(self, node):
         params = self._default_config_params(node)
+        in_0, in_1 = map(str, node.get_input_variable().shape[:2])
```
vloncar, line 83: Do they really need to be strings at this point? They get passed to `format()` so that should convert them to strings if they are not already

calad0i: `out_*` are being concatenated with string here and there, and I am not sure if `in_*` are not used somewhere else in some weird way as there was an explicit `str()` op.

jmitrevs:
There are related pytest errors:
```
FAILED test_layernorm.py::test_layernorm_parsing[Vitis] - AttributeError: 'TensorVariable' object has no attribute 'dim_names'
```

vloncar:
The other failures will be addressed in different PRs

PR #1325, opened 2025-07-08
Report full path when file not found

llewitt:
This is maybe personal preference but I found it much easier to debug why my weights files weren't being found after making this change.


vloncar:
Indeed. Can you make the change to the nnet_helpers.h in catapult backend so we keep consistent?

llewitt:
@vloncar I think this is good to go?

vloncar (approved):

PR #1327, opened 2025-07-08
Add compatibility with new Vitis command-line compilation flow

vloncar:
# Description

From 2024.2 onward there's no more `vitis_hls` command, and `vitis` command launches the new GUI. The command `vitis-run` can be used instead to invoke the compilation from the command line. However, in their infinite wisdom, AMD didn't provide an option to pass arguments to the .tcl script. Therefore we have to tweak how we control the build arguments. The writer will now write a new file called `build_opt.tcl` which contains the build arguments that we used to override from the command line and source it just like `project.tcl` is included. The `build()` command will rewrite the default contents of the file with the arguments provided by the user.

## Type of change

- [x] New feature (non-breaking change which adds functionality) - sort of

## Tests

Affects synthesis, so no tests

## Checklist

- [x] Yes

### DIFF
--- a/hls4ml/backends/vitis/vitis_backend.py
+++ b/hls4ml/backends/vitis/vitis_backend.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import subprocess
 import sys
+from pathlib import Path
 
 from hls4ml.backends import VivadoBackend
 from hls4ml.model.flow import get_flow, register_flow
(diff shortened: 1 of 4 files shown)


bo3z:
These changes look good to me. Though I am wondering whether we should keep vitis_hls for backward compatibility --- is vitis_run available in for e.g., 2023.2, 2024.1?

vloncar:
Did either of you actually try to run it? The first commit had a bug :wink:

JanFSchulte:
Don't have a new enough Vitis installation around to test the first commit :-D But I did test the first tutorial notebook with Vitis 2024.1 now that it's also affected with this PR and it runs fine.

PR #1338, opened 2025-07-21
bit exact extension

calad0i:
# Description

Extends full model bit propagation infra to other frontends; `config_from_whatever_model()` may still be needed.

## Type of change

## Checklist

- [x] all


calad0i:
> For the `quantized_bits` quantizer, this seems to result in a duplication of the inputs, for example:
> 
> ```
> void myproject(
>     result_t quantizer_input[1],
>     result_t quantizer_input[1]
> ) { ...
> ```
> 
> This doesn't compile due to redefinition of the variable. Haven't dug deep enough to actually find the cause of the problem.

This test converts an orphaned quantizer. The fuse quantizer flow merged it with the input layer by overridding the precision, and there is 0 layer left in the model and hence `inp=out`, the duplicated name.

Made fuse only happen is bit-exact is enabled, though in this case I think removing this test would also make sense.

PR #1340, opened 2025-07-23
oneapi backend less copy paste

calad0i:
# Description

Less code copy paste

## Type of change

[x] House keeping

## Tests

-

## Checklist

- [x] all

### DIFF
--- a/hls4ml/backends/oneapi/oneapi_backend.py
+++ b/hls4ml/backends/oneapi/oneapi_backend.py
@@ -200,11 +200,8 @@ class OneAPIBackend(FPGABackend):
         subprocess.run(f'make {build_type}', shell=True, cwd=builddir, check=True)
 
         if run:
-            assert build_type in (
-                'fpga_emu',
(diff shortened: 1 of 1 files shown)


hls4ml/backends/oneapi/oneapi_backend.py
```diff
@@ -207,7 +199,12 @@ def build(self, model, build_type='fpga_emu', run=False):
         subprocess.run('cmake ..', shell=True, cwd=builddir, check=True)
         subprocess.run(f'make {build_type}', shell=True, cwd=builddir, check=True)
 
-        if run and build_type in ('fpga_emu', 'fpga_sim', 'fpga'):
+        if run:
+            assert build_type in (
```
vloncar, line 203: asserts should be sanity checks from internal flows. exceptions should be used for user-provided calls. since we expect user to provide the `build_type` and `run` parameters, the incorrect combination should be an exception, not an assert.

PR #1347, opened 2025-07-28
Coyote accelerator backend

bo3z:
# Description

> :memo: This PR introduces a new accelerator backend, CoyoteAccelerator, which leverages the open-source Coyote shell for deploying models on a PCI-attached FPGA. 

Generally, Coyote offers several advantages, when compared to some other shells, including:
* Networking support, so the backend can easily be extended to support distributed inference. Also interesting for in-network ML.
* GPU - FPGA integration, so models can be executed on a combination of hardware
* Dynamic reconfiguration, which could allow run-time reconfiguration of models
* Multi-tenancy, so multiple models could be deployed concurrently.

The backend is briefly described in Section 9.7 of the paper: https://arxiv.org/pdf/2504.21538.

## Type of change

## Tests
This backend was compared agains a modified* version of the VivadoAccelerator backend: the backend was modified to run HLS synthesis with Vitis instead of Vivado (also using Vitis templates and optimizers), while the rest of the backend infrastructure (drivers, data movers remained the same since they also work in newer version of Vivado). Results are attached below - clearly indicating an advantage in Coyote, for two reasons (1) optimised data movement, bypassing card memory and (2) optimised host-side library (Python, C++).

In principle, the correct test would be to compare against VitisAccelerator (#991), but only after the io_parallel issues are resolved. However, the expectation is that the result will stay mostly the same, sine the underlying platform requires a data copy between host and card memory. 

Will add some more results, also for io-stream CNN, and comparisons to VitisAccelerator.

Screenshot 2025-07-28 at 12 11 24

Figure above: comparison of CoyoteAccelerator with modified Vivado Accelerator for the UNSW-NB15 dataset in io_parallel. 

## Checklist


hls4ml/templates/coyote_accelerator/host_libs.cpp
```diff
@@ -39,9 +39,7 @@ void CoyoteInference::predict() {
     }
 
     // Poll on completion; each batch increments the counter by one
-    while (coyote_thread.checkCompleted(coyote::CoyoteOper::LOCAL_TRANSFER) != batch_size) {
-        std::this_thread::sleep_for(std::chrono::nanoseconds(50));
-    }
+    while (coyote_thread.checkCompleted(coyote::CoyoteOper::LOCAL_TRANSFER) != batch_size) {}
```
vloncar, line 42: Wouldn't this cause 100% CPU usage while the program is polling?

bo3z: On one of the cores, yes. 

But sleep for less than 50us is not well-defined on most Linux platforms. Hence, measured latency can go from ~4us to >50us even though the "true" execution latency is still 4us.

docs/backend/accelerator.rst
```diff
@@ -75,3 +75,70 @@ The ``predict`` method will send the input data to the PL and return the output
...
+================================
+
+Additionally, the Coyote backend generates and compiles a C++ program that can be used to run inference on the FPGA.
+The binary can be found in ``<hls4ml-output-dir>/build/<project-name>_cyt_sw/bin/test`` and when launched, it will
+run inference using the inputs from ``tb_data``. Similar to the Python overlay, the bitstream and driver must be loaded before running the inference.
+
+Example
+======================
+
+Similar to the ``VivadoAccelerator``backend, we first generate a bitstream from a Keras model ``model`` and a config.
```
JanFSchulte: Documentation should mention that hls4ml needs to be cloned with the submodules checked out to get Coyote, and that a Vitis installation is needed to be present.

JanFSchulte (commented):
A few misc comments based on trying to run the CoyoteAccelerator for a dummy model. Right now, i am stuck with a python import error:
image

which is puzzling because I do have `jinja2` installed in my environment and the same import works fine in an interactive python session.

Also, can you fix the pre-commit issues?

vloncar (commented):
Looks good. I have mostly questions for better understanding and minor nitpicks that I don't feel are crucial, rather optional.

hls4ml/backends/coyote_accelerator/coyote_accelerator_overlay.py
```diff
@@ -0,0 +1,104 @@
...
+
+                for j in range(batch_size):
+                    tmp = self.coyote_lib.get_inference_predictions(model, j)
+                    y[total_batches * batch_size + j] = np.ctypeslib.as_array(tmp, shape=y_shape)
+
+                cnt = 0
+                total_batches += 1
+
+        self.coyote_lib.free_model_inference(model)
+        print(f'Batch size: {batch_size}; batches processed: {total_batches}')
```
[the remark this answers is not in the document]
bo3z: I'll add a function argument, verbose.

hls4ml/backends/coyote_accelerator/coyote_accelerator_backend.py
```diff
@@ -0,0 +1,150 @@
...
+            f'-DEN_HLS_COSIM={int(cosim)} '
+            f'-DEN_HLS_VALIDATION={int(validation)} '
+            f'-DHLS_CLOCK_PERIOD={hls_clock_period} '
+            f'-DHLS_CLOCK_UNCERTAINTY="{str(hls_clock_uncertainty)}%"'
+        )
+
+        if not os.path.exists(f'{model.config.get_output_dir()}/build/{model.config.get_project_name()}_cyt_hw'):
+            os.mkdir(f'{model.config.get_output_dir()}/build/{model.config.get_project_name()}_cyt_hw')
+        os.chdir(f'{model.config.get_output_dir()}/build/{model.config.get_project_name()}_cyt_hw')
+        os.system(cmake_cmd)
```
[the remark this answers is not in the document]
bo3z: Will change.

bo3z:
Should be mostly ready to merge now. Think all the comments were addressed.

JanFSchulte:
Still need to fix the pre-commit checks ;) The "line too long" ones can't be automatically fixed.

JanFSchulte:
pre-commit.ci autofix

JanFSchulte:
The pre-commit manifest test now complains:

```
  lists of files in version control and sdist match
  you have hls4ml/contrib/Coyote/hw/services/network/hls/ip_handler/testgen/memlib.pyc in source control!
  that's a bad idea: auto-generated files should not be versioned
```

So I think you just need to git remove this file and we can move on with the PR tests.

bo3z:
Should be fixed now.

hls4ml/templates/coyote/host_libs.cpp
```diff
@@ -0,0 +1,73 @@
...
+        dst_sgs.emplace_back(dst_sg);
+    }
+}
+
+CoyoteInference::~CoyoteInference() {}
+
+void CoyoteInference::flush() {
+    // Reset output tensors to zero
+    for (unsigned int i = 0; i < batch_size; i++) {
+        memset(dst_mems[i], 0, out_size);
```
JanFSchulte: Should this be `out_size * sizeof(float)` like above?

hls4ml/backends/coyote/coyote_overlay.py
```diff
@@ -0,0 +1,106 @@
...
+        model = self.coyote_lib.init_model_inference(batch_size, int(np.prod(X[0].shape)), int(np.prod(y_shape)))
+
+        cnt = 0
+        avg_latency = 0
+        avg_throughput = 0
+        total_batches = 0
+        for x in X:
+            self.coyote_lib.set_inference_data(model, x, cnt)
+            cnt += 1
+            if cnt == batch_size:
```
JanFSchulte, line 86: Do we need a protection here for cases where `len(X)` isn't a multiple of `batch_size` and we end with some leftover input that's silently dropped from the inference?

hls4ml/backends/coyote/coyote_overlay.py
```diff
@@ -0,0 +1,106 @@
...
+                    tmp = self.coyote_lib.get_inference_predictions(model, j)
+                    y[total_batches * batch_size + j] = np.ctypeslib.as_array(tmp, shape=y_shape)
+
+                cnt = 0
+                total_batches += 1
+
+        self.coyote_lib.free_model_inference(model)
+        if verbose:
+            print(f'Batch size: {batch_size}; batches processed: {total_batches}')
+            print(f'Mean latency: {round(avg_latency / total_batches, 3)}us (inference only)')
```
JanFSchulte: I think we can have a division by zero here if `len(X) < batch_size`. Not sure how likely that is, but might be worth putting some protection.

JanFSchulte:
pre-commit.ci autofix

JanFSchulte:
pre-commit.ci autofix

JanFSchulte (approved):

PR #1362, opened 2025-08-14
PQuant🌶️ integration

enlupi:
# Description

This draft PR is intended to start a discussion on how to best integrate the [PQuant🌶️](https://github.com/nroope/PQuant) library with hls4ml.
The current state of the PR supports layers pruned and quantized using fixed quantizers. Support for high granularity quantization is still in development.

⚠️⚠️ In order for the PQuant layers to be parsed correctly, the code expects the user to run some custom functions that add the quantization parameters as layer attributes (see the test file for examples). In the future these functions should be added to PQuant🌶️.  ⚠️⚠️ 

## Type of change

## Tests

Unit test `test/pytest/test_pquant.py` was added to test correct parsing of the pruned/quantized layers and that the precision is set correctly.
⚠️⚠️ THE TEST IS CURRENTLY BROKEN! The file needs to be split into two, to test PyTorch and Keras frontend separately ⚠️⚠️ 

## Checklist


calad0i (commented):
Maybe also checkout #1338 so that the bit-exact pass also applies for pquant models - In short, if the quantizers are converted to `FixedPointQuantzier` object, or linear layers with quantization precision set properly with `trusted=True`, full model precision propagation ([this file](https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/optimizer/passes/bit_exact.py)) can be used with little changes and all precision inference should work (will need to add rules for the multiplier relu function but likely nothing else).
If you don't want to use that pass, maybe consider deciding if users should use `config_from_keras_model` or not. If not, maybe leave it untouched and do enforcement in a dedicated pass.

PR #1369, opened 2025-08-21
Bump checkout version

JanFSchulte:
https://github.com/fastmachinelearning/hls4ml/pull/1367 seems to break the mirroring of the repo to gitlab. The bump to the version of the checkout action are still necessary, so this PR is meant to replace the broken one. 

## Type of change

- [x] Other (Specify)

## Checklist

- [x] N/A

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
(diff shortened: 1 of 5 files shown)


jmitrevs:
Do you understand the test failure?

JanFSchulte:
I have set up fresh tokens for the repo that we need to keep the actions working after this update. It works nicely for the `test-sphinx/build` one, but the update of the branch on PR from fork is not able to be tested in this PR and might still be problematic. I think it should work, but we probably just need to bite the bullet, merge this, and then see if everything works.

PR #1375, opened 2025-09-02
use int16 instead

calad0i:
# Description

Fix weird corner cases where bitwidth overflows in `bit_exact` flow. 

The issue can only be trigger by an ill-defined model. However, as the overflows masks the correct error message, users may be confused.

## Type of change

## Checklist

- [x] all


calad0i:
> In what corner case does this happen?

...when a model has multiple unquantized components. By default, the bitwdith produced is 126 if nothing is specified. However, if there is some non-trivial operation happening in the middle (e.g., matmul), the output can be >128 bits, and boom, overflow.

Since int16 has upto 65535, we will need around... (65535-128)/23 ~ 3000 consecutive unquantized layers (max weight bw is capped at 23), so I suppose it is fine :joy:

PR #1377, opened 2025-09-04
Fix pytorch related pytest failures

JanFSchulte:
There currently are a couple of pytorch-related failures in the pytests. 

The main one is a bug with recurrent layers where the weights and hidden weights need to be transposed to mach the keras convention that is used in our implementation. Since this is not strictly a channels-last conversion I have added this directly to the parser so that the channels-last conversion feature can still be used as before. This issue was masked before because of a bug in the converter that ignored the "off" setting. After I fixed this recently, this issue was revealed, but missed at the time. 

I also increased the bit width in one of the einsum tests to avoid occasional failures because of limited precision. There is nothing wrong otherwise, so this seems like the best option. 

## Type of change

## Tests

Failing pytests pass now

## Checklist

### DIFF
--- a/hls4ml/converters/pytorch/recurrent.py
+++ b/hls4ml/converters/pytorch/recurrent.py
@@ -51,10 +51,9 @@ def parse_rnn_layer(operation, layer_name, input_names, input_shapes, node, clas
     if class_object.dropout > 0:
         raise Exception('hls4ml does not support RNNs with dropout')
 
+    # transpose weight and recurrent weight to match keras order used in the HLS code
     layer['weight_data'] = class_object.weight_ih_l0.data.numpy().transpose()
-    print('weight shape: ', layer['weight_data'].shape)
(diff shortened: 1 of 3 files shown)


calad0i:
Can you push an empty commit and rerun the tests?

jmitrevs:
I retried the HGQ + DA tests. Try to replicate locally with the same random seed. Otherwise, could there be version differences between the test environment and some libraries that would make the difference?

JanFSchulte:
It's certainly reproduced in the test environment here, see for example the oneAPI accelerator PR tests: https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/60904849 So I guess it's just a matter of luck reproducing it online. I just didn't want to run all 386 tests offline to reproduce on failure and haven't found a seed yet that gives me a failure for a single test.

JanFSchulte:
Guess it passed now. A little weird.

calad0i (approved):

PR #1383, opened 2025-09-09
rm fxpmath dependency

calad0i:
# Description

Remove dependency of #1139 on fxpmath, which generates a warning when pip installing for using outdated `setup.cfg`.

## Type of change

- [x] Housekeeping

## Tests

Inherits from #1139

## Checklist

- [x] all

### DIFF
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -2,9 +2,9 @@ import math
 from collections.abc import Iterable
 
 import numpy as np
-from fxpmath import Fxp
 
 from hls4ml.model.optimizer import ConfigurableOptimizerPass
+from hls4ml.model.optimizer.passes.bit_exact import minimal_kif
 from hls4ml.model.types import (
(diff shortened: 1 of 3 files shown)


calad0i:
By definition of the fixed point numbers, there will always be one more value on the negative side, and `ac_fixed` has the same behavior here. The problem is likely oneapi specific. Also, for the current oneapi to work, unsigned numbers also needs >=2 bits iirc.

jmitrevs:
I had assumed it was just signed that required two bits, but I might be wrong. Best to confirm with @laurilaatu .

laurilaatu:
@jmitrevs @calad0i 
Currently 2 bits is required for both signed and unsigned cases.

calad0i:
For `ac_fixed` it is already the case (see https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/fpga/fpga_types.py#L99). For `ap_fixed` there is no need to do this.

JanFSchulte (approved):

PR #1391, opened 2025-09-28
Avoid layer with name `out_something` crashing conversion

calad0i:
# Description

Currently, if a layer is named `out_something`, the generated firmware will be unsynthesizable due to a naming conflict. The program checks for the prefix `out_` for special variables, which was unprotected and may come from user defined layer names and let the writer writing the same definition twice (`out_something` and `something` will be two keys in attrs, but `VariableMapping` strips off the `out_` prefix and the writer iterates twice on it) 

## Type of change

## Checklist

- [x] all


JanFSchulte:
I'm pretty sure the test failure here is unrelated to the changes in this PR, so I'm gonna go ahead and merge it.

PR #1400, opened 2025-11-19
Integration of PQuantML with hls4ml

enlupi:
# Description

This PR introduces the possibility of directly parsing PQuantML layers into hls4ml, both using the PyTorch and Keras V3 frontend. It uses HGQ2 conventions and relies on the *bit_exact* optimization pass to ensure correct precision is enforced and bit exactness.

## Type of change

## Tests

Tests are present both for PyTorch and Keras V3 frontend: `test_pquant_pytoch.py` and `test_pquant_keras.py`. 

## Checklist


JanFSchulte:
Thanks! I think you need to catch up to change in the keras3 parser where the `register` decorator has been removed.

docs/advanced/pquant.rst
```diff
@@ -0,0 +1,89 @@
...
+  - Joint pruning + quantization
+  - Layer-wise precision configuration
+  - Flexible training pipelines
+  - PyTorch and Keras V3 implementations
+  - Integration with hardware-friendly toolchains (e.g., hls4ml)
+
+PQuantML enables efficient deployment of compact neural networks on resource-constrained hardware such as FPGAs and embedded accelerators.
+
+
+Key Features
```
JanFSchulte: Please come up with a new name for this section to avoid the duplicate label

JanFSchulte (changes requested):

docs/advanced/pquant.rst
enlupi: I am not sure I understand the issue.
You need a different name from Key Features?

JanFSchulte: The sphinx test has a warning about the hgq.rst having a section with the same label, so I assume that it throws off the linking within the docs if there are duplications. So yeah, slightly changing it from "Key Features" should solve it

hls4ml/converters/pytorch/pquant.py
```diff
@@ -0,0 +1,174 @@
...
+parse_pqpool1d_layer = make_pquant_handler(parse_pooling_layer, 'PQAvgPool1d', 'AvgPool1d')
+parse_pqpool2d_layer = make_pquant_handler(parse_pooling_layer, 'PQAvgPool2d', 'AvgPool2d')
+
+
+def parse_quant_activation_layer(operation, layer_name, input_names, input_shapes, node, class_object, data_reader, config):
+    layer = {}
+
+    layer['activation'] = class_object.activation_name
+
+    print(f'Parsing activation: {layer["activation"]}')
```
JanFSchulte: We usually don't have printout like this, please remove.

hls4ml/converters/pytorch/pquant.py
```diff
@@ -0,0 +1,174 @@
...
+    if layer['activation'] == 'hard_tanh':
+        layer['class_name'] = 'HardActivation'
+        layer['slope'] = 0.5
+        layer['shift'] = 0.5
+        layer['slope_prec'] = FixedPrecisionType(width=2, integer=0, signed=False)
+        layer['shift_prec'] = FixedPrecisionType(width=2, integer=0, signed=False)
+        warn(f'Hard Tanh activation {layer_name} is currently not supported for bit-exactness.')
+
+    elif layer['activation'] == 'relu' and class_object.use_multiplier:
+        raise Exception('hls4ml does not currently support activations with multiplier')
```
JanFSchulte: We usually raise a more specific `NotImplementedError` in these cases.

hls4ml/converters/pytorch/pquant.py
```diff
@@ -0,0 +1,174 @@
...
+                    None,
+                    None,
+                )[0]
+            ]
+            oq_shapes = [output_shape]
+    else:
+        oq_confs = []
+        oq_shapes = []
+
+    out_shapes = []
```
JanFSchulte: Looks like this `out_shapes` list is not used for anything?

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -274,20 +274,41 @@ def resolve_getitem_source(node_name, visited=None):
...
+                            'Layer name: {}, layer type: {}, input shape: {}'.format(
+                                lay['name'],
+                                lay['class_name'],
+                                input_shapes,
+                            )
+                        )
+                    layer_list.append(lay)
+
+                    if idx < len(layer) - 1:
+                        inputs_map[lay['name']] = inputs_map.get(layer[idx + 1]['name'], layer[idx + 1]['name'])
```
JanFSchulte: I don't quite get the logic in this mapping, can you please explain it?

docs/advanced/pquant.rst
```diff
@@ -0,0 +1,89 @@
...
+      input_shape=input_shape,
+      )
+   hls_model = convert_from_pytorch_model(PQmodel, ...)
+   # Model-wise precision propagation is done automatically for PQuantML models for bit-exactness
+   # Do NOT pass precision config if you don't know what you are doing
+
+   hls_model.compile()
+
+.. note::
+   Do not pass any precision configuration from ``hls4ml.converters.convert_from_<frontend>_model`` in general. PQuantML-defined models will invoke model-wise precision propagation automatically to ensure bit-exactness between the PQuantML model and the generated HLS code (See `here <./precision.html>`__ for more details).
```
JanFSchulte: Can we document also what happens if this note is ignored? Will things break, will results be silently incorrect?

JanFSchulte (changes requested):

hls4ml/converters/pytorch/pquant.py
enlupi: left there by mistake, I removed it.

hls4ml/converters/pytorch/pquant.py
enlupi: Yes, I now return the output directly without creating the out_shapes object.

hls4ml/converters/pytorch/pquant.py
enlupi: I modified it to use the `NotImplementedError`.

docs/advanced/pquant.rst
enlupi: Understood, changed it to "Key Characteristics"

hls4ml/converters/pytorch_to_hls.py
enlupi: Let's assume we have two consecutive layers A, who has an input and output quantizer, and B.
Layer A will be parsed as 3 layers: FixedPointQuantizer A_iq, A proper, and FixedPointQuantizer A_oq. The input argument for the node B, though, will still be A, as A_oq did not exist at the moment of tracing the model. This means that the correct flow of data in the model will not be reconstructed, and both A_oq _and_ B will be seen as outputs of A (A_oq will also be one of the outputs of the global model, as it is not used as input by anyone).

What these lines do is ensuring that if a layer sees A as input, it gets redirected to the next parsed layer, i.e. A_oq, and the correct flow can be reconstructed. 

Admittedly, this has the useless consequence of redirecting the layers with A_iq as node argument to A, which should never happen as A_iq did not exist at tracing time.
Perhaps it could be reworked to just work on the A equivalent with something along the lines of:
 
`  if lay['name'] == layer_name and idx != len(layer) - 1:`
 `      inputs_map[lay['name']] = inputs_map.get(layer[-1]['name'], layer[-1]['name'])`
 
 Alternatively, we could do an explicit check for PQuantML layers and set the input map before all of this, as it is done in line 209 for skip layers.

JanFSchulte: Thanks for the explanation. I think the suggestion with the explicit check to only work on A might be a good idea to not have unintended consequences some time in the future, even if the current implementation is save now.

enlupi: Got it.
I changed it so that I explicitly check if it is a PQuantML layer and if it has an output quantizer; only in this case, the inputs_map is changed.

docs/advanced/pquant.rst
enlupi: I was following HGQ2's note, as the underlying bit exact pass that is invoked is the same. From my understanding, explicit layer precision is simply ignored (as per the [Model-wise Precision Inference doc](https://fastmachinelearning.org/hls4ml/advanced/precision.html)), while if you have both auto precision in the config and invoke the bit exact pass it will lead to undefined behavior.

JanFSchulte:
Looks good to me now, thanks! Lets run the test one last time and merge once they are done.

JanFSchulte (approved):

PR #1409, opened 2025-12-12
Backend predict hook

dimdano:
Move the existing C-simulation logic into an internal `_predict()` method and update the public `predict()` API to optionally dispatch to a backend-provided `predict()` implementation. For example, this might be useful for external backends (e.g., AIE) to implement accelerator-specific execution while preserving the standard hls4ml prediction interface.

## Type of change

## Checklist

- yes all

### DIFF
--- a/hls4ml/model/graph.py
+++ b/hls4ml/model/graph.py
@@ -913,7 +913,7 @@ class ModelGraph(Serializable):
             return backend.predict(self, x, *args, **kwargs)
 
         return self._predict(x)
-        
+
     def trace(self, x):
         print(f'Recompiling {self.config.get_project_name()} with tracing')
         self.config.trace_output = True


vloncar:
Sorry, but this is not really the best way of handling this functionality and a more thorough review should have been done. It should be possible to develop a backend by implementing a `Backend` class. Now you're making it such that a Backend developer should know a secret way to override a pretty basic functionality that is not documented anywhere, and can only be figured out by digging deep into the code. Furthermore it doesn't really address the requirements of passing extra (non-input) parameters to the `predict()` method, as requested by the acc backend developers.

If the backends require different way of handling `predict()`, expose it through `Backend.predict(*args, **kwargs)`, then make all current ones use `FPGABackend.predict()`, then others that need more can override. That way all functionality is in a common place for all backends, not scattered across two separate boundaries (IR vs backend).

jmitrevs:
I would advocate for the implementation Vladimir mentioned. All the other similar backend-specific things are done with inheritance, not hooks.

PR #1418, opened 2026-01-17
Add LHC trigger use case context to README

siddardhadesu:
What:
- Added one sentence to the README introduction mentioning ultra-low-latency applications and LHC trigger use cases.

Why:
- Helps readers quickly understand the motivation and common real-world usage of hls4ml.

Note:
- Documentation-only change; no code behavior affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@
 
 A package for machine learning inference in FPGAs. We create firmware implementations of machine learning algorithms using high level synthesis language (HLS). We translate traditional open-source machine learning package models into HLS that can be configured for your use-case!
 
-hls4ml is designed for ultra-low-latency inference and has been used in real-time applications such as trigger systems in high-energy physics experiments (including at the Large Hadron Collider).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,8 @@
 
 A package for machine learning inference in FPGAs. We create firmware implementations of machine learning algorithms using high level synthesis language (HLS). We translate traditional open-source machine learning package models into HLS that can be configured for your use-case!
 
+hls4ml is designed for ultra-low-latency inference and has been used in real-time applications such as trigger systems in high-energy physics experiments (including at the Large Hadron Collider).
```
JanFSchulte: There are some concerns with this formulation. Specifically we should avoid the impression that hls4ml is HEP specific and limited to these kind of applications. So this should be rephrased to make clear that HEP trigger systems are just one example. Other applications to mention could be picked from section 10 of https://arxiv.org/pdf/2512.01463 

Also, we should probably avoid the term `real-time` as it means different things to different people.

PR #1421, opened 2026-01-20
Fix Keras v3 model conversion in numerical profiling

Abubakar-rashid:
# Description

## Summary
Fixes numerical profiling failure when using Keras v3 models with `hls4ml.model.profiling.numerical()`.

## Problem
When calling `numerical(model=kerasmodel, hls_model=hlsmodel, X=X_test)` with a Keras v3 model, the function fails because `convert_from_config()` in `hls4ml/converters/__init__.py` was always using the Keras v2 converter (`keras_v2_to_hls()`), even for Keras v3 models.

The bug occurs at line 125 where the code routes to `keras_v2_to_hls()` for all Keras models without checking the version. This causes incompatibility issues when the `numerical()` profiling function internally calls `get_unoptimized_hlsmodel()`, which uses `convert_from_config()` to recreate the model.

## Solution
Added Keras version detection in `convert_from_config()` to properly route:
- Keras v3 models → `keras_v3_to_hls()`
- Keras v2 models → `keras_v2_to_hls()`

This mirrors the same logic already present in `convert_from_keras_model()` (lines 231-237), ensuring consistency across the codebase.

## Changes Made
- **hls4ml/converters/__init__.py**: Added `elif 'KerasModel' in yamlConfig:` block with version check
- **test/pytest/test_keras_v3_profiling.py**: New comprehensive test file with 5 unit tests

## Type of change

- [ ] Documentation update
- [ ] Other (Specify)

### Test File: `test/pytest/test_keras_v3_profiling.py`

Created 5 unit tests that verify numerical profiling works correctly with Keras v3 models:

1. **test_keras_v3_numerical_profiling_simple_model** - Tests basic profiling with Dense layers
2. **test_keras_v3_numerical_profiling_with_activations** - Tests profiling with test data (X parameter)
3. **test_keras_v3_numerical_profiling_conv_model** - Tests profiling with Conv2D layers
4. **test_keras_v3_numerical_profiling_with_hls_model** - Tests integration with HLS models
5. **test_keras_v3_numerical_profiling_batch_norm** - Tests with BatchNormalization layers

### Test Behavior
- **Master branch**: Tests FAIL  (demonstrates the bug exists)
- **This branch**: Tests PASS  (demonstrates the fix works)

### How to Reproduce/Run Tests
```bash
# Tests will be skipped if Keras < 3.0 is installed
pytest test/pytest/test_keras_v3_profiling.py -v

# To run in an environment with Keras 3.0+:
pip install keras>=3.0
pytest test/pytest/test_keras_v3_profiling.py -v

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ exclude: (^hls4ml\/templates\/(vivado|quartus)\/(ap_types|ac_types)\/|^test/pyte
 
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.11
+  rev: v0.14.13
   hooks:
     - id: ruff
       args: [--fix]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ exclude: (^hls4ml\/templates\/(vivado|quartus)\/(ap_types|ac_types)\/|^test/pyte
 
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.13
+  rev: v0.14.14
   hooks:
     - id: ruff
       args: [--fix]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     - id: ruff-format
 
 - repo: https://github.com/tox-dev/pyproject-fmt
-  rev: v2.11.1
+  rev: v2.12.1
   hooks:
     - id: pyproject-fmt
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Consult [GitHub Help](https://help.github.com/articles/about-pull-requests/) for
 1. In the pull request description, clearly document all changes made and the expected behavior.
 1. If you are introducing new functionality, add at least one unit test under the `test` folder and make sure it passes before you submit the pull request.
 1. Similarly, if you are fixing a bug, add at least one unit test under the `test` folder such that the master branch fails the test and your branch passes the test.
-1. Install and run `pre-commit` on the files that you have edited. We are adiabatically turning on linting for the full repository (see [#678](https://github.com/fastmachinelearning/hls4ml/pull/678)).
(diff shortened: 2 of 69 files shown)

### CODE REVIEW

**Review by JanFSchulte — CHANGES_REQUESTED** [review]

**Review by bo3z — CHANGES_REQUESTED** [review]

**Comment by calad0i** [comment]
hgq2 is never intended to be used together or coexist with any version of qkeras, though. Not a bug, won't fix.

**Comment by calad0i** [comment]
> some broken version of qkeras to be installed

Isn't this the compatibility layer provided by hgq2 with conflicting the namespace? hgq2 depends on `(quantizers -> numpy), keras, tqdm`, and nothing else.

**Comment by JanFSchulte** [comment]
Sounds like it is this compatibility layer then. If that means we can't remove it, we need to better protect against it here https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/profiling.py#L29-L40. @Abubakar-rashid can you change https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/profiling.py#L30 to be 

```
from qkeras import QActivation
```
?
if that works, you can add back the hgq tests to the test list that Benjamin was asking about.

bo3z (approved):

JanFSchulte (approved):

rayansingh:
Hi @Abubakar-rashid @vloncar,

I'm new to the project and looking to make a first contribution. I noticed that `test_keras_v3_numerical_profiling_with_hls_model` in `test/pytest/test_keras_v3_profiling.py` is still skipped, with a note that `convert_from_config` needs to be updated to be compatible with Keras v3 model serialization.

The skip was introduced in this PR (#1421).

Is this something you're planning to follow up on, or is it open for someone else to pick up? If the latter, I'd like to take a shot at it and would appreciate any pointers.

Thanks,
Rayan

vloncar:
To my knowledge, this is not being followed up, and is free to take on. It's possible that there are underlying issues with these libraries being unable to coexist in the same environment. Perhaps we'll have to split the profiling into multiple modules to resolve the imports. And improve the testing of each function. In other words, it may be more work than what it seems from the discussion.

rayansingh:
Thanks, I'll take a look. Appreciate the pointers!

PR #1422, opened 2026-01-23
Allow AP_SAT, AP_RND for 'maximum' precision in HLS Config

morunner:
# Description

During the development of a gravnet model for hls4ml, I found that setting rounding and saturation for the maximum allowed precision can be beneficial for increasing accuracy. Below you can see histograms and mean difference with one standard deviation when the maximum precision is `ap_fixed<16,8,AP_RND,AP_SAT,0>` and when rounding and saturation are not enabled (`ap_fixed<16,8>`). 

In the current upstream main branch of hls4ml,  rounding and saturation modes set through the 'maximum' field  in the HLS config are ignored during precision inference (see e.g. [here](https://github.com/fastmachinelearning/hls4ml/compare/main...morunner:hls4ml:allow-sat-rnd-in-max-precision#diff-f72b0c39349380f86b420d32544fafed8e2e06e27c042d11abb4ca11a9203350L190)). I thus propose a single function `_apply_max_precision_constraints` to be applied where necessary in the `infer_precision.py` module, which adheres the following rules:

- Width/Integer: Always constrained to the minimum of inferred vs max.
- Rounding/Saturation: Inherited from max_precision ONLY if they differ from the defaults 
   (meaning the user likely set them explicitly).
- Signedness: max_precision signed arg is always preferred.

We can of course discuss, what the preferred ruleset should be here.

No additional dependencies are required for this change.

## Type of change

- [x] Other (Accuracy Improvement)

### Pytest

Added a new pytest module, `test_max_precision.py`, which tests the newly added `_apply_max_precision_constraints` function isolated and within the `_infer_precision` function, using mocks.

### Conversion to HLS

Ran the full jupyter notebook for gravnet keras conversion to hls at hls4ml-gravnet ([Link](https://github.com/morunner/hls4ml-gravnet/blob/dev/notebooks/compare-keras-hls4ml-model.ipynb)) to generate the below listed plots, with the proposed change enabled and disabled. The profiling section was ran with [this](https://github.com/morunner/hls4ml/commit/a626ea1bd198bdfcc98c5038011b5bda9b5756ec) fix applied. We currently do not provide the fully trained model open-source, since it is not finalized. @bo3z please contact me directly, also regarding the dataset, if needed. 

## Checklist

- [ ] I have made corresponding changes to the documentation. -> tbd depending on the ruleset for RND, SAT

## GravNet plots showing accuracies and bias across layers

With rounding and saturation enabled for the maximum precision.

gravnet_accuracy_with_rnd_sat_dense_params
gravnet_accuracy_without_rnd_sat_dense_params
gravnet_bias_with_rnd_sat_dense_params
gravnet_bias_without_rnd_sat_dense_params

### DIFF
--- a/hls4ml/model/optimizer/passes/infer_precision.py
+++ b/hls4ml/model/optimizer/passes/infer_precision.py
@@ -113,6 +113,54 @@ class InferPrecisionTypes(ConfigurableOptimizerPass):
                 return False
         return True
 
+    def _apply_max_precision_constraints(self, node, precision):
+        """
+        Clamps the precision to the node's max_precision constraints.
+
+        Logic:
+        1. Width/Integer: Always constrained to the minimum of inferred vs max.
+        2. Rounding/Saturation: Inherited from max_precision ONLY if they differ from the defaults
+           (meaning the user likely set them explicitly).
+        3. Signedness: max_precision signed arg is always preferred.
+        """
+        max_precision = self._get_maximum_precision(node)
+
+        if max_precision is None:
+            return precision
+
+        new_width = min(precision.width, max_precision.width)
+        new_integer = min(precision.integer, max_precision.integer)
+
+        # Default modes defined in FixedPrecisionType
+        default_type = FixedPrecisionType()
+        DEFAULT_RND = default_type.rounding_mode
+        DEFAULT_SAT = default_type.saturation_mode
+        DEFAULT_SAT_BITS = default_type.saturation_bits
+
+        if max_precision.rounding_mode != DEFAULT_RND:
+            new_rounding_mode = max_precision.rounding_mode
+        else:
+            new_rounding_mode = precision.rounding_mode
+
+        if max_precision.saturation_mode != DEFAULT_SAT:
+            new_saturation_mode = max_precision.saturation_mode
+        else:
+            new_saturation_mode = precision.saturation_mode
+
+        if max_precision.saturation_bits != DEFAULT_SAT_BITS:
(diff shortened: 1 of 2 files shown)


jmitrevs:
Generally setting the maximum precision is not something we recommend using much, I don't think. It's better to either quantize the values in the training, or if doing PTQ, explicitly set certain widths to more reasonable values in the configuration. The maximum width is not granular enough for that. One can see what width one gets without the maxumum setting and modify the configuration till it is satisfactory.

Also, rounding and saturation for the accumulator often make the accumulation much slower. It is better to keep it wider and if needed, use saturation and rounding in the activation step right after it, where its cost is insignificant. This more fine-grained way of doing things is recommended instead of using the maximum precision.

bo3z:
@morunner do you have some results (in terms of resource usage) before and after this change. @jmitrevs and @calad0i mentioned in our last dev meeting that AP_SAT may not be the most resource-friendly for accumulators (due to the underlying implementation of the saturation operation). Most times, the recommended way is to simply increase the bit width of the variable.

If that's the case from your results we should keep this PR open as a reference (in case someone is interested in using similar functions), but not merge it.

calad0i:
> slightly higher latency

You can always increase the bitwidth to match/get better accuracy, since accum is very cheap in comparison to the other operations happening there. In most cases, one should be able to have no real performance difference between the python model and the hls model bit allocating sufficient bitwidths in the accumulator to avoid overflow/underflows. Unless there is very specific hardware restriction (usually not), I would suggest against the use of maximum precision.

PR #1426, opened 2026-01-26
vivado/vitis support sample broadcasting merge ops

calad0i:
# Description

Allow add/subtract/mul/avg/... between tensors of shape `(a, b, c)` and `(1, c)` or `(c,)` or `(1, b, c)` (must match the last dimensions) in vivado/vitis backends. Illegal operations still fail silently like what they do now.

## Type of change

## Checklist

- [x] all

### DIFF
--- a/hls4ml/backends/vivado/passes/merge_templates.py
+++ b/hls4ml/backends/vivado/passes/merge_templates.py
@@ -5,6 +5,7 @@ from hls4ml.model.layers import Concatenate, Dot, Merge
 # Merge templates
 
 merge_config_template = """struct config{index} : nnet::merge_config {{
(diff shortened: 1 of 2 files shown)


JanFSchulte:
Looks like there are issues with the streaming implementation of merge, all the test failures are caused by compiler errors like 

```
firmware/nnet_utils/nnet_merge_stream.h:15:35: error: ‘n_elem’ is not a member of ‘config3’
   15 |     for (int i = 0; i < CONFIG_T::n_elem / input1_T::size; i++) {
```

calad0i:
Streaming supported is not added, so it is still broken as what it was.

JanFSchulte (approved):

PR #1429, opened 2026-01-30
Da custom layer

calad0i:
# Description

Allows the use of any subclass of `keras.layers.Layer` object that is supported in da4ml to be used in the hls4ml project, such as LUT-based layers (such as `hgq.layers.DenseT/ConvT12D` or` keras.layers.Conv3D`). Due to parser limitation, `keras.operation.Operation` still cannot be supported (yet), so arbitrary tensor slicing or indexing operations will not work for now.

Now, the keras v3 parser will try da parser fallback before trying v2 parser. If it falls by not finding the corresponding layer or da4ml is not present, it will proceed with the current behavior.

## Type of change

## Tests

Still no keras v3 test infra there.

## Checklist

- [ ] All.

### DIFF
--- a/hls4ml/converters/__init__.py
+++ b/hls4ml/converters/__init__.py
@@ -167,6 +167,8 @@ def convert_from_keras_model(
     backend='Vivado',
     hls_config=None,
     bit_exact=None,
+    allow_da_fallback=True,
+    allow_v2_fallback=True,
     **kwargs,
 ):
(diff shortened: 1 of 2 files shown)


JanFSchulte:
Yes, the point is more that they are not explained in the doc for that function. So the user has no pointers on how to configure this, except by reading the code. I guess the specific request would be to add this to the function doc string for `convert_from_keras_model`.

PR #1442, opened 2026-02-17
Issue 1438 write tar consistency

### LINKED ISSUE #1438: CI artifacts missing in some jobs
## Prerequisites
- [ ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
Some CI jobs upload no artifacts even though the CI template expects tarballs and/or synthesis reports, which results in artifact upload warnings. *The jobs do not fail, but this is noisy*.

### Steps to Reproduce
1. Look at a CI job that only executes tests without synthesis and without explicit tarball generation, e.g.
   - `test_recurrent_pytorch.py::test_gru`, `test_lstm`, `test_rnn`
   - https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/69760179
2. Observe the artifact upload step.

### Expected behavior
Artifacts should be produced.

### Actual behavior
Artifacts are produced only for certain backends/tests (e.g., Catapult where tarballs are always generated). Other jobs produce none and the upload step logs:
```
Uploading artifacts...
WARNING: test/pytest/*.tar.gz: no matching files...
WARNING: test/pytest/synthesis_report_*.json: no matching files...
ERROR: No files to upload
```

### Possible fix
- Add a `WriteTar` configuration to `CatapultWriter` for consistency with other backends.
- If artifacts are desired, update tests to set `write_tar=True`.

### Additional context
The root cause is inconsistent artifact production: Catapult always generates `.tar.gz` archives (unconditional `write_tar` in `CatapultWriter.write_hls()`), while other backends only generate tarballs when `WriterConfig.WriteTar` is enabled (default False).

Evidence
- `hls4ml/writer/catapult_writer.py`: `write_hls()` always calls `self.write_tar(model)` (no config gate).
- `hls4ml/writer/vivado_writer.py`, `oneapi_writer.py`, `quartus_writer.py`: tarball creation is gated by `WriteTar`.

marco66colombo:
# Description

This PR fixes #1438 by making `WriteTar` handling consistent and explicit across writers.

## Summary

- Move `WriteTar` if condition one layer higher: check `WriterConfig.WriteTar` in `write_hls()` before calling `write_tar()`.
- Keep `write_tar()` as a pure “write tarball” method (no internal config check).
- Add `write_tar` support to Catapult backend config (`create_initial_config(..., write_tar=False)` + `WriterConfig['WriteTar']`), so Catapult follows the same config path as other backends.
- Extend `test_writer_config::test_write_tar` coverage to include `Catapult`.
- Add `write_tar` support to `VivadoAcceleratorBackend.create_initial_config(...)` and include `VivadoAccelerator` in `test_writer_config::test_write_tar` coverage.

## Motivation / Context

Issue #1438 highlights inconsistent artifact behavior and writer semantics.

A key point from prior maintainer discussion is that keeping tar generation disabled by default is better for users, to avoid generating large numbers of extra files during local development and debugging.

This PR keeps that default (`WriteTar=False`) while making behavior consistent across backends.

## Notable behavior note

For Catapult specifically, when `WriteTar=True` and tar creation is triggered:
- previous behavior: keep existing tar and print “already exists”
- new behavior: overwrite existing tar (consistent with other backends)

This is intentional for consistency across backends.

For VivadoAccelerator, with `WriteTar=True`, tar can be generated/regenerated twice (once via the Vivado writer path and once via `write_new_tar` after accelerator-specific files are added). This matches past behavior and is not introduced in this PR.

## Follow-up (in progress)

I still need to update tests/CI so artifact generation is always enabled where CI expects tar artifacts.

Possible approaches under consideration:
- Add a central switch in `test/pytest/conftest.py` (for example wrapping conversion/config creation to force `write_tar=True` in CI).
- Update each test case explicitly to pass `write_tar=True` (more explicit, but many more edits required).

I am working on this as a follow-up step.

## Type of change

- [ ] Documentation update
- [ ] Other (Specify)

## Tests

- I ran `test/pytest/test_writer_config.py` in a local environment.

**Test Configuration**:

- Local development environment using same image used in CI.

## Checklist


marco66colombo:
As mentioned in the **Follow-up** section, we still need to decide how to (and if) ensure tar artifact generation in CI.

Should we enable tar artifact generation for every CI test case?

If yes, I see two practical approaches:

1. **Central CI toggle in `test/pytest/conftest.py`**
   - Implementation: force `write_tar=True` for conversions only when a CI toggle is enabled (env var like `HLS4ML_FORCE_WRITE_TAR=1` or pytest flag like `--force-write-tar`).
   - Pros: single-point change, no per-test edits, easy CI on/off.
   - Cons: more implicit behavior (wrapper/monkeypatch), may miss uncommon paths if they bypass wrapped entrypoints.

2. **Explicit per-test updates**
   - Implementation: update tests to pass `write_tar=True` in each conversion call.
   - Pros: explicit behavior in each test, finer-grained control.
   - Cons: many edits across the suite, higher maintenance burden.

Happy to proceed with either approach if we want to activate artifact generation.

JanFSchulte:
Dusting this off, and answering the last question, I think the approach in the Vivado backend seems better to me. Not propagating the other settings seems destined to introduce unexpected behavior at some point.

PR #1466, opened 2026-04-16
HGQ2 rename of attention layers

laurilaatu:
# Attn layers in HGQ

## Type of change

Fix breaking conversion of attention layers caused by HGQ2 restructuring.


JanFSchulte:
Thanks! This probably should go with an adjustment of the required version of HGQ2 so that things stay consistent? @calad0i what do you think?

calad0i:
Should be `HGQ2>=0.1.7` for this change.

calad0i:
One unrelated test failure due to ci issue, all passed otherwise.

calad0i (approved):

PR #1470, opened 2026-05-05
Basic SNN functionality in hls4ml (mapped from snntorch)

bmdillon:
# Description

This PR adds initial PyTorch/SNN support to hls4ml.

It includes conversion support for IF/LIF neuron layers, SNN readout handling, Vivado templates, documentation, tests, and a small example notebook.

Dependencies:
- `snntorch`
- `torch`

## Type of change

- [x] Documentation update

## Tests

Added pytest coverage for PyTorch SNN parsing, IF/LIF layer handling, scalar/vector beta and threshold values, SNN readout options, and custom PyTorch SNN extension parsing.


bmdillon:
pre-commit.ci autofix

docs/advanced/snn.rst
```diff
@@ -0,0 +1,184 @@
+========================================
+Spiking Neural Networks (PyTorch/SNN)
+========================================
+
+This page describes the initial SNN support in the PyTorch frontend.
```
vloncar, line 5: Somewhere in this document it should also be mentioned that the support is added in Vitis backend, not others.

docs/advanced/snn.rst
```diff
@@ -0,0 +1,184 @@
...
+
+   pip install hls4ml[snn]
+
+Execution model
+===============
+
+Current hls4ml SNN implementations are synchronous (clock-driven). Neuron state
+updates and layer computations run in standard HLS pipelines/streams each cycle
+according to interface handshakes. The generated design is not a native
+asynchronous/event-routed neuromorphic architecture (yet!).
```
vloncar, line 24: I would also add information that the implementation doesn't feature the control of unrolling/latency via `reuse_factor` as most of the other networks in Vitis/Vivado backends do.

docs/advanced/snn.rst
```diff
@@ -0,0 +1,184 @@
...
+
+   mem[i] = beta * mem[i] + input[i];
+
+No threshold or reset-on-spike is applied in membrane mode. The supported
+membrane decision policies are:
+
+* ``argmax_membrane``
+* ``binary_logit`` (emits ``mem(class_1) - mem(class_0)`` for binary classifiers)
+
+The example in the Jupyter Notebook follows this approach.
```
vloncar: This should link to the tutorial, which is where the notebook will be moved

hls4ml/converters/pytorch/snn.py
```diff
@@ -0,0 +1,146 @@
+import numpy as np
```
vloncar, line 1: (not a comment on the file, but overall on the PR)

Have you considered parsing via [NIR](https://neuroir.org/docs)? This could expose support for more frameworks, but with more checks for supported features.

hls4ml/converters/pytorch/snn.py
```diff
@@ -0,0 +1,146 @@
...
+        raise Exception(f'Only scalar "beta" is supported for SNNReadout membrane mode, got shape {arr.shape}')
+    return float(arr[0])
+
+
+@pytorch_handler('Leaky')
+def parse_lif_layer(operation, layer_name, input_names, input_shapes, node, class_object, data_reader, config):
+    assert operation == 'Leaky'
+
+    n_out = input_shapes[0][-1]
+    beta = _parse_scalar_or_vector(getattr(class_object, 'beta', None), 'beta', n_out)
```
vloncar: Couldn't you move the `getattr` inside the `_parse_...` function to match the style of the others?

hls4ml/converters/pytorch_to_hls.py
```diff
@@ -422,6 +422,17 @@ def resolve_getitem_source(node_name, visited=None):
     if len(input_layers) == 0:
         input_layers = None
 
+    fixed_window_readouts = [
```
vloncar: This parsing post-processing to complete layer information is better done as an optimizer pass later on, rather than in the parser itself. We try to avoid doing these things in the parser

hls4ml/model/layers.py
```diff
@@ -1037,6 +1037,115 @@ def initialize(self):
...
+                var_name='be{index}',
+                type_name=beta_t.name,
+                precision=beta_t.precision,
+                data=beta,
+            )
+
+
+class SNNReadout(Layer):
+    _expected_attributes = [
+        Attribute('n_classes', configurable=False),
```
vloncar: `configurable=False` is the default, no need to be explicit about it

hls4ml/model/layers.py
```diff
@@ -1037,6 +1037,115 @@ def initialize(self):
...
+        Attribute('window_size', value_type=int, default=1, configurable=False),
+        Attribute('class_threshold', value_type=int, default=1, configurable=False),
+        Attribute('beta', value_type=float, default=1.0),
+        ChoiceAttribute('output_mode', choices=['spike', 'membrane'], default='spike', configurable=False),
+        ChoiceAttribute(
+            'state_reset_policy',
+            choices=['fixed_window', 'tlast', 'host_pulse', 'never'],
+            default='fixed_window',
+            configurable=True,
```
vloncar: Is it actually intended to be configurable? In this context, it means the user may override it after they made the model already

hls4ml/templates/vivado/nnet_utils/nnet_if_neuron.h
```diff
@@ -0,0 +1,112 @@
...
+    static constexpr float threshold = 1.0;
+    static const snn_reset_mode reset_mode = snn_reset_mode::subtract;
+    typedef float threshold_t;
+    typedef float membrane_t;
+};
+
+template <class data_T, class res_T, typename CONFIG_T>
+void if_neuron(data_T data[CONFIG_T::n_in], res_T res[CONFIG_T::n_out],
+               const typename CONFIG_T::threshold_t threshold_vec[CONFIG_T::n_out]) {
+    #pragma HLS PIPELINE II=1
```
vloncar: Is there a way to add support for tuning the pipeline parallelism via `reuse_factor`?

hls4ml/templates/vivado/nnet_utils/nnet_if_neuron.h
```diff
@@ -0,0 +1,112 @@
...
+            for (unsigned i = 0; i < CONFIG_T::n_out; i++) {
+                #pragma HLS UNROLL
+                mem[i] = 0;
+            }
+        }
+    }
+}
+
+template <class data_T, class res_T, typename CONFIG_T>
+void if_neuron(hls::stream<data_T> &data_stream, hls::stream<res_T> &res_stream,
```
vloncar: To be in line with (most) of the other backend templates, streaming implementations could be placed in `nnet_snn_stream.h`, so you only have two files (3 if you keep the `nnet_snn_common.h` separate)

hls4ml/templates/vivado/nnet_utils/nnet_snn_readout.h
```diff
@@ -0,0 +1,323 @@
...
+            ts = 0;
+            for (unsigned i = 0; i < CONFIG_T::n_classes; i++) {
+                #pragma HLS UNROLL
+                mem[i] = 0;
+            }
+        }
+        return;
+    }
+
+    if (CONFIG_T::decision_rule == snn_decision_rule::binary_logit) {
```
vloncar: Could the `decision_rule` be implemented as a function pointer? That way you avoid the excessive if statements (and the valid combination has already been checked in the earlier stages of conversion)

hls4ml/templates/vivado/build_prj.tcl
```diff
@@ -161,7 +161,10 @@ if {$opt(reset)} {
 } else {
     open_solution "solution1"
 }
-catch {config_array_partition -maximum_size $maximum_size}
+if {[catch {config_array_partition -maximum_size $maximum_size}]} {
+    # Vitis HLS 2024.1 removed -maximum_size; use complete threshold when available.
+    catch {config_array_partition -complete_threshold $maximum_size}
```
vloncar: This is not really the same. `maximum_size` was saying "don't error out when asked to partition an array up to this size", while `complete_threshold` says "automatically do complete partitioning of anything below this size". This will have consequences in implementations that don't do complete partitioning (basically the `resource` strategy). We shouldn't do it. Eventually we should remove the command completely, once the legacy Vivado backend is deprecated

hls4ml/contrib/snntorch.py
```diff
@@ -0,0 +1,63 @@
+import torch
```
vloncar, line 1: I tend to think this shouldn't be a part of the `utils` submodule, rather a separate module where we keep the training-related. In the past we placed this under `contrib`.

test/pytest/test_extensions_pytorch_snn.py
```diff
@@ -0,0 +1,76 @@
...
+                n_classes=4, window_size=10, decision_rule='threshold_then_argmax', class_threshold=3
+            )
+
+        def forward(self, x):
+            x = self.fc(x)
+            return self.readout(x)
+
+    pmodel = PyTorchModel()
+    config = hls4ml.utils.config_from_pytorch_model(
+        pmodel, (4,), default_precision='ap_fixed<16,6>', granularity='name', backend='Vivado'
```
vloncar: Don't you mean Vitis here? (and in all other tests too)

hls4ml-snn-example.ipynb
vloncar, line 1: This should be made as a PR to the hls4ml-tutorials repo, not placed here.

vloncar (changes requested):
Overall looks good, needs minor changes to keep it in line with the rest of the codebase

bmdillon:
pre-commit.ci autofix

docs/advanced/snn.rst
bmdillon: Yeah good point, thanks, added to docs

docs/advanced/snn.rst
bmdillon: This is also now added to the docs

hls4ml/converters/pytorch/snn.py
bmdillon: I did consider this yeah, but the snntorch parsing mirrors the standard pytorch parsing pretty closely so this seemed like the smoothest way to implement SNNs.  With snntorch the whole workflow also seems closer to the pytorch workflow.  It might be possible to support more frameworks through converting from NIR to snntorch, then to hls4ml.  But perhaps in the future it will make sense to parse via NIR.

hls4ml/converters/pytorch/snn.py
bmdillon: Yeah, done

hls4ml/converters/pytorch_to_hls.py
bmdillon: Moved this to an optimizer pass, added model/optimizer/passes/snn.py
Does this work?

hls4ml/model/layers.py
bmdillon: Removed this

hls4ml/model/layers.py
bmdillon: No actually, good point, updated

hls4ml/templates/vivado/nnet_utils/nnet_if_neuron.h
bmdillon: Possibly, the spiking neurons are stateful across timesteps, so I need to think a bit about it.  But it’s something to look at in the near-future for sure.  For now, the dense layers in the SNN are standard pytorch layers and so have RF support, and that’s the main bottleneck in reducing LUTs.

bmdillon: Can this be left for a future PR?

hls4ml/templates/vivado/nnet_utils/nnet_if_neuron.h
bmdillon: This has now been refactored into nnet_snn.h and nnet_snn_stream.h

hls4ml/templates/vivado/build_prj.tcl
bmdillon: Yes you’re right, I should’ve returned to this.  I’ve reverted this change.

hls4ml/contrib/snntorch.py
bmdillon: I’ve moved this to contrib for now.  I agree, utils is not the correct place for it.  But this class is not training related, it’s central to the readout of the SNN over the entire window.

test/pytest/test_extensions_pytorch_snn.py
bmdillon: Yes, updated

hls4ml-snn-example.ipynb
bmdillon: Removed the file, will put the tutorial in the hls4ml-tutorials repo

hls4ml/templates/vivado/nnet_utils/nnet_snn_readout.h
bmdillon: Could you explain a bit more, where should the functions be?  For now I’ve refactored the code in nnet_snnreadout.h to include helper functions, if we need more I can edit further

docs/advanced/snn.rst
bmdillon: Updated

JanFSchulte:
The PR tests show several failures for this new feature. Can you have a look and make sure things work correctly?

image

bmdillon:
> The PR tests show several failures for this new feature. Can you have a look and make sure things work correctly?
> 
> image

These should be fixed now, all run locally.  Just a note, I updated utils/config.py, it was setting some layer attributes to default values if they weren't TypeAttribute or RF.  If this is the intended behaviour, I can find an alternative fix.

hls4ml/converters/pytorch/snn.py
vloncar: ok, let's see if there's interest for NIR in the future

hls4ml/templates/vivado/nnet_utils/nnet_if_neuron.h
vloncar: It can, if you put mention the limitation in the documentation, which you did, so ok

hls4ml/templates/vivado/nnet_utils/nnet_snn_readout.h
vloncar: I was thinking if there's a way to have something like `CONFIG_T::decision_rule()` that would be a function pointer to the helper functions that you would set in the config templates. We used a similar approach for matrix-vector multiplication functions in the past. It's not essential for this PR though, it's purely cosmetics

hls4ml/model/optimizer/passes/snn.py
```diff
@@ -0,0 +1,31 @@
...
+        ]
+        if len(readouts) == 0:
+            return False
+
+        window_size = readouts[0].get_attr('window_size', 0)
+        changed = False
+        for node in model.graph.values():
+            if isinstance(node, (IFNeuron, LIFNeuron)) and node.get_attr('window_size') != window_size:
+                node.set_attr('window_size', window_size)
+                changed = True
```
vloncar, line 29: I don't think this is needed. It is required when changes are made to the execution graph (nodes added/removed/replaced), or if change in an attribute of one node affects some other. But the latter case should be avoided. I don't think that we need to return `True` in this pass. Can you check if it works without it?

hls4ml/utils/config.py
```diff
@@ -406,7 +406,9 @@ def make_layer_config(layer):
             elif attr.name == 'reuse_factor':
                 layer_config[attr.config_name] = default_reuse_factor
             else:
-                if attr.default is not None:
+                if attr.name in layer:
+                    layer_config[attr.config_name] = layer[attr.name]
```
vloncar, line 412: So the idea here is that there may be attributes of the layer that the user configured during model definition, which would then be part of the model when parsed, but that the user can still override during conversion and deployment? Up to this point we considered the semantics of configurable attributes as something purely in hls4ml that controls the behavior of the model in a way that cannot be specified during model definition (in e.g., pytorch, snntorch or kersas), such as the reuse factor or type definitions for non-quantized models. One can argue this breaks the concept of what it means for something to be configurable, but I don't have strong opinions on enforcing it. We kept it hidden (not part of the dict this method produces, but still doable if you really know what you're doing). Thoughts @JanFSchulte @jmitrevs @calad0i @bo3z ?

vloncar:
To me this looks ready, assuming tests pass. Let's see if other main devs have issue with the slight change in semantics of configurable attributes.

hls4ml/utils/config.py
calad0i: This looks good to me. Maybe apply the same change to the other frontends also for consistency.

I would argue that the qkeras autoconfig generator is already doing similar things with the quantizers, i.e., quantizer params defined through config and overridable by the user, though not the same `Attribute` knobs, I would argue they are similar.

If the user overrides anything inside the default config, they shall know what they are doing. Also, adding the 
"default" config here don't change the actual downstream behavior (if this doesn't exist, one can still insert it manually), so also no new footgun.

hls4ml/model/optimizer/passes/snn.py
bmdillon: This is needed to update the window_size in the spiking neurons.  Everything is mapped from snntorch, and in snntorch the Leaky neurons do not have a window_size parameter, so hls4ml defaults it to zero, and then updates it later to match the window_size in the readout layer.  Without it, the window_size parameter in the spiking neurons stays at 0 and would not reset properly.  I ran a test locally to confirm this is needed.

hls4ml/utils/config.py
bmdillon: Same change now applied in keras and onnx frontends in config.py

hls4ml/templates/vivado/nnet_utils/nnet_snn_readout.h
bmdillon: Ok I see, it's a good point, I'll keep it in mind for a future PR

vloncar (approved):
Looks great.

PR #1473, opened 2026-05-13
Add support for qkeras v3

makoeppel:
These are the Handlers needed for qkerasV3 (https://pypi.org/project/qkeras-v3/). Updated tutorials: https://github.com/bo3z/hls4ml-tutorial/tree/tutorial-revamp-2026 were tested with the new qkerasV3.


bo3z:
Thank you for this!

We should update pyproject.toml to add an optional dependency qkeras-v3 which installs qkeras-v3 and a recent version of tensorflow (one matching keras-v3) from pip (the same way qkeras is now installed).

In my understanding this doesn't yet cover all the layers, which is fine, but we should probably add QConv1D/QConv2D layers, as these are used in the tutorials.

Once the above two are added, the PyTests could also migrate to QKeras v3 (should be as easy as updating the pyproject.toml for the testing dependencies)

makoeppel:
I did all the requested changes. For the qconv part the fallback to keras-v2 did the job. I added only a new test case in the test_qkeras.py to test for qconv layers. I also had to convert the 3layer model to "*.keras": https://github.com/fastmachinelearning/example-models/pull/24.

Most tests will actually fail if only keras3 is used. Since the qkeras part is now mostly converted to keras-v3 I was wondering if there should not be also an update to keras3 for the rest of the project. Certainly not in this PR.

docs/intro/setup.rst
```diff
@@ -53,7 +53,7 @@ The following Python packages are all optional and are only required if you inte
 * `PyTorch <https://pytorch.org/get-started>`_ is required by the PyTorch converter.
 
 * Quantization support
-   * `QKeras <https://github.com/fastmachinelearning/qkeras>`_: based on Keras v2. See `frontend/keras <../frontend/keras.html>`_ for more details
+   * `QKeras <https://github.com/fastmachinelearning/qkeras-v3>`_: based on Keras v3. See `frontend/keras <../frontend/keras.html>`_ for more details
```
bo3z: We should try to keep both QKeras and QKeras v3 (the same way we support HGQ and HGQ2 as well as Keras v2 and v3).

hls4ml/converters/keras_v3/_base.py
```diff
@@ -136,6 +136,8 @@ def maybe_get_activation_config(self, layer, out_tensors):
         activation = getattr(layer, 'activation', None)
         name = layer.name
         if activation not in (keras.activations.linear, None):
+            if 'qkeras' in str(type(activation)):
```
bo3z: What's the purpose of this check?

hls4ml/converters/keras/qkeras.py
```diff
@@ -118,27 +120,42 @@ def get_activation_quantizer(keras_layer, input_names, activation_name='activati
...
 
     activation_config = keras_layer['config'][activation_name]
-    quantizer_obj = get_quantizer(activation_config)
-    activation_config = {}
-    # some activations are classes
-    if hasattr(quantizer_obj, 'get_config'):
-        activation_config['class_name'] = quantizer_obj.__class__.__name__
-        if activation_config['class_name'] == 'ternary' or activation_config['class_name'] == 'binary':
-            activation_config['class_name'] += '_tanh'
-        activation_config['config'] = quantizer_obj.get_config()
```
bo3z, line 128: These changes seem pretty much the same as the original but catching some corner cases. Is there anything specific to QKeras v3? And do these changes affect QKeras v2 complatibility?

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,49 @@
...
+    handles = ('qkeras.qlayers.QActivation', 'QActivation')
+
+    def handle(
+        self,
+        layer,  # qkeras.qlayers.QActivation
+        in_tensors,
+        out_tensors,
+    ) -> tuple[dict[str, Any], ...]:
+
+        # --- v2 handler plumbing (same pattern as your dispatcher.v2_call) ---
```
bo3z: Can we get rid off these left-over comments (if they don't explain much; which seems to be the case)? Same for the above IMPORTANT: comment

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,49 @@
...
+        layer_dict = {'config': config, 'class_name': layer.__class__.__name__}
+
+        class IsolatedLayerReader:
+            def get_weights_data(self, layer_name, var_name):
+                assert layer_name == layer.name, f'Processing {layer.name}, but handler tried to read {layer_name}'
+                for w in layer.weights:
+                    if var_name in w.name:
+                        return np.array(w)
+                return None
+
```
bo3z: This class seems repeated in several files...can we have its definition once in some common/util file.

hls4ml/converters/keras_v3/qkeras/batchnorm.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+from ..core import KerasV3LayerHandler
+
+
+class QKerasQConv2DBatchnormHandler(KerasV3LayerHandler):
+    handles = ('qkeras.qlayers.QConv2DBatchnorm', 'QConv2DBatchnorm')
```
bo3z: Is it possible to also handle just a quantized conv layer (QConv2D) layer, without BatchNorm. We should also consider Conv1D case.

hls4ml/converters/keras_v3_to_hls.py
```diff
@@ -262,21 +262,40 @@ def get_weights_data(self, layer_name, var_name):
...
-                'name': f'{layer.name}_{act_cls_name}',
-                'input_keras_tensor_names': (intermediate_tensor_name,),
-                'output_keras_tensor_names': output_names,
-            }
+            if 'qkeras' in str(type(activation)):
+                from hls4ml.converters.keras.qkeras import get_activation_quantizer
+
+                act_config = get_activation_quantizer(layer_dict, input_names)
+                act_config.update(
+                    {
```
bo3z: We're not defining class_name, activation in this branch? Are we sure this will not cause downstream issues?

And, will this approach still work for QKeras v2?

hls4ml/model/optimizer/passes/qkeras.py
```diff
@@ -112,15 +112,18 @@ def match(self, node):
     def transform(self, model, node):
         # The quantizer has to be applied to set the scale attribute
         # This must be applied to the _unquantized_ weights to obtain the correct scale
-        import tensorflow as tf
 
         quantizer = node.weights['weight'].quantizer.quantizer_fn  # get QKeras quantizer
         weights = node.weights['weight'].data_unquantized  # get weights
-        qweights = quantizer(tf.convert_to_tensor(weights))
+        qweights = quantizer(weights)
```
bo3z: Do we not need the tf.convert_to_tensor? Will it still work with QKeras v2 if we remove the case?

pyproject.toml
```diff
@@ -44,11 +44,7 @@ optional-dependencies.optimization = [
   "packaging",
 ]
 optional-dependencies.profiling = [ "matplotlib", "pandas", "seaborn" ]
-optional-dependencies.qkeras = [
-  "qkeras",
-  "tensorflow>=2.8,<=2.14.1",
-  "tensorflow-model-optimization<=0.7.5",
-]
+optional-dependencies.qkeras = [ "qkeras-v3" ]
```
bo3z: We should support both qkeras-v2 and v3, so we should have seperate dependencies for qkeras and qkeras-v3

pyproject.toml
```diff
@@ -70,6 +66,7 @@ optional-dependencies.testing-keras3 = [
   "da4ml",
   "hgq2>=0.1.7",
   "keras>=3.10",
+  "qkeras-v3",
```
bo3z, line 81: here, we should follow the existing approach:

testing with keras-v2 and keras-v3, separately (which we already do)

test/pytest/test_qkeras.py
```diff
@@ -563,10 +653,10 @@ def test_qsimplernn(test_case_id, backend):
     X = np.stack([X, X], axis=1).reshape(1, 5, 2)
 
     model = Sequential()
+    model.add(Input(shape=(5, 2)))
```
bo3z: Is this a requirement of QKeras v3? That models must have an input layer? Or is this a consequence of how we parse the model?

hls4ml/model/quantizers.py
```diff
@@ -126,8 +129,29 @@ def __init__(self, config):
...
-        return self.quantizer_fn(data).numpy()
-        # return self.quantizer_fn(data)
+        if self.quantizer_fn is not None:
+            try:
+                return self.quantizer_fn(data).numpy()
+            except TypeError:
+                pass
+        if self.qkeras_config['class_name'] != 'quantized_bits':
+            raise RuntimeError(f'Cannot evaluate QKeras quantizer {self.qkeras_config["class_name"]}')
+        return self._quantize_bits(data, self.qkeras_config)
```
bo3z: What's the purpose of this check? Will we ever encounter this condition (asking since we didn't cover this case in qkeras-v2)?

test/pytest/test_keras_v3_api.py
```diff
@@ -29,6 +31,92 @@
 test_root_path = Path(__file__).parent
 
 
+def test_qkeras_qdense_v3_handler_chains_quantized_activation(monkeypatch):
```
bo3z: What's the purpose of these two tests? These don't fit our usual style of tests?

bo3z (changes requested):
As stated in some of the comments, the biggest change is aiming to keep support for both QKeras v2 and QKeras v3 (the same way we support Keras v2 and v3). The rest of the comments are for consistency with the rest of the code base and better understanding 

@calad0i since you worked on the Keras v3 parser, could you also provide some feedback on this PR when you get the chance?

docs/intro/status.rst
```diff
@@ -59,6 +60,8 @@ A summary of the on-going status of the ``hls4ml`` tool is in the table below.
 +-----------------------+-----+-----+--------------+--------+--------+-----+
 | QKeras                | ✅  | ✅  | ✅           | ✅     | N/A    | N/A |
 +-----------------------+-----+-----+--------------+--------+--------+-----+
+| QKeras-v3             | ✅  | ✅  | ✅           | ✅      | ✅    | N/A |
```
calad0i, line 63: Isn't garnet a specific example implemented in some version of qkeras v2?

hls4ml/converters/keras/qkeras.py
calad0i: Actually, I would suggest keeping all keras v2 logic kept as-is. If anything differs from the current behavior, use the v3 parser. The v2 parser was supposed to be a temporary workaround, not something to be used long termed in production.

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,49 @@
+from typing import Any
+
+import numpy as np
+
+from hls4ml.converters.keras_v3.core import KerasV3LayerHandler  # adjust import to your tree
```
calad0i: Please ask your agent to clean up left over comments like these...

hls4ml/converters/keras_v3_to_hls.py
JanFSchulte: Since this is in the keras v3 converter,  this code will never be used to parse a QKeras v2 model. So we don't need to worry about maintaining compatibility here (even though I agree that this is a concern in general).

hls4ml/converters/keras_v3/_base.py
calad0i: Base class shall not be modified tailored to any specific backend. Return in tuple form to skip for automatic activation appending, see example in relu handler or unary lut for HGQ.

hls4ml/converters/keras_v3_to_hls.py
```diff
@@ -262,21 +262,40 @@ def get_weights_data(self, layer_name, var_name):
...
             ret[0]['output_keras_tensor_names'] = (intermediate_tensor_name,)
-            act_cls_name = activation.__name__
-            act_config = {
-                'class_name': 'Activation',
-                'activation': act_cls_name,
-                'name': f'{layer.name}_{act_cls_name}',
-                'input_keras_tensor_names': (intermediate_tensor_name,),
-                'output_keras_tensor_names': output_names,
-            }
+            if 'qkeras' in str(type(activation)):
```
calad0i: Same, modifying the base class for one backend is not allowed. Use a base class for qkeras backend for that, see example in https://github.com/calad0i/hls4ml/blob/main/hls4ml/converters/keras_v3/hgq2/_base.py#L70

test/pytest/test_keras_v3_api.py
```diff
@@ -29,6 +31,92 @@
...
+    assert activation_config['activation'] == 'ternary_tanh'
+    assert activation_config['threshold'] == 0.33
+    assert activation_config['activation_quantizer']['class_name'] == 'ternary_tanh'
+
+
+def test_config_from_functional_model_with_false_built_flag():
+    inputs = keras.Input(shape=(3,), name='input')
+    outputs = Dense(2, name='dense')(inputs)
+    model = keras.Model(inputs, outputs)
+    model.built = False
```
JanFSchulte: What is this trying to test? I thought we explicitly require the model to be built?

hls4ml/converters/keras_v3_to_hls.py
```diff
@@ -303,13 +322,17 @@ def parse_keras_v3_model(model: 'keras.Model', allow_da_fallback=True, allow_v2_
         ValueError: If a circular dependency is detected.
     """
 
-    assert model.built, 'Model must be built before parsing'
```
calad0i, line 306: Why replacing the built check if inputs and outputs exists?

test/pytest/test_qkeras.py
```diff
@@ -1,14 +1,15 @@
 import warnings
 from pathlib import Path
```
JanFSchulte, line 2: Echoing previous comments by Benjamin, we should have a separate test for QKeras v3 and keep testing QKeras v2.

hls4ml/model/optimizer/passes/qkeras.py
calad0i: Likely will break the v2 behavior here... maybe a helpder fn that like `to_numpy` that safely makes any object into `np.ndarray`. The .numpy() behavior works for tf but if you use jax frontend it will break.

hls4ml/model/types.py
```diff
@@ -837,7 +837,10 @@ def _format(self):
 
     def __iter__(self):
         data = self._format()
-        self._iterator = iter(data.reshape((np.product(data.shape[:-1]), 2)))
+        if hasattr(np, 'product'):
```
calad0i: Let's use math.prod here?

pyproject.toml
calad0i: hgq2 and qkeras-v3 both exports `qkeras` module, likely will cause some issues.

hls4ml/converters/keras_v3/qkeras/batchnorm.py
calad0i: Matching condition without module path, like `QConv2DBatchnorm` here, shall be used with great care and only when it is strictly needed (e.g., general fallback handler). Doesn't seem to be the case here.

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,49 @@
...
+        v2_handler = v2_layer_handlers.get(layer.__class__.__name__)
+        if v2_handler is None:
+            raise ValueError(f'No v2 handler found for {layer.__class__.__name__}')
+
+        hls_conf, _ = v2_handler(layer_dict, input_names, input_shapes, reader)
+
+        hls_conf['input_keras_tensor_names'] = list(input_names)
+        hls_conf['output_keras_tensor_names'] = list(output_names)
+        hls_conf.setdefault('name', layer.name)
+        hls_conf.setdefault('class_name', 'QActivation')
```
calad0i: There is the `self.default_config` config you can use to fill in the trivial attrs.

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,49 @@
...
+
+        return (hls_conf,)
```
calad0i, line 35: Already tuple style return, the special case handling for activation in base class is redundant

hls4ml/converters/keras_v3/qkeras/batchnorm.py
```diff
@@ -0,0 +1,57 @@
...
+        from hls4ml.converters.keras_v2_to_hls import layer_handlers as v2_layer_handlers
+
+        v2_handler = v2_layer_handlers.get(layer.__class__.__name__)
+        if v2_handler is None:
+            raise ValueError(f'No v2 handler found for {layer.__class__.__name__}')
+
+        ret, _ = v2_handler(layer_dict, input_names, input_shapes, reader)
+
+        # override / normalize the names used by the v3 graph parser
+        ret['name'] = layer.name
```
calad0i: same, set default with `self.default_config` if possible to simplify the code

hls4ml/converters/keras_v3/qkeras/layer.py
```diff
@@ -0,0 +1,57 @@
...
+
+        # override / normalize the names used by the v3 graph parser
+        ret['name'] = layer.name
+        ret['class_name'] = ret.get('class_name', 'QDense')
+        ret['module'] = layer.__module__
+        ret['input_keras_tensor_names'] = [t.name for t in in_tensors]
+        ret['input_shape'] = [list(t.shape[1:]) for t in in_tensors]
+        ret['output_keras_tensor_names'] = [t.name for t in out_tensors]
+
+        activation = config.get('activation')
```
calad0i, line 33: activation handling here is treated as a special case. The base class shall cover this already though...

calad0i (changes requested):

docs/intro/setup.rst
makoeppel: Done

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: Done

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: I removed the return in the base class.

hls4ml/converters/keras_v3/_base.py
makoeppel: I removed this check since its not needed as @calad0i said.

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: I remove the comment.

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: I moved this class into utils.py

pyproject.toml
makoeppel: I changed this to: optional-dependencies.qkeras-v3 = ["qkeras-v3"] and reverted the optional-dependencies.qkeras.

hls4ml/converters/keras_v3/qkeras/batchnorm.py
makoeppel: @bo3z in the end I did not had to do anything special for the QConvND layers. I added also a test for this `test_qconv_activation_kwarg`.

makoeppel: @calad0i you are correct. I removed the not needed matching conditions. Only for QActivation I need it without a module. Are you fine with this?

makoeppel: Actually @bo3z you are correct. After removing the activation check things are failing. So I created now a layer.py which handles all needed qkeras layers.

pyproject.toml
makoeppel: I fully split this now.

test/pytest/test_keras_v3_api.py
makoeppel: I removed them. They were needed for the built check which I removed.

test/pytest/test_keras_v3_api.py
makoeppel: Yes since the built check was removed I had to test it. Anyway now I added back the built requirement and removed this.

test/pytest/test_qkeras.py
makoeppel: I added a test_qkerasV3.py now and left this untouched.

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: Removed

hls4ml/converters/keras_v3_to_hls.py
makoeppel: Done

hls4ml/converters/keras_v3/qkeras/batchnorm.py
makoeppel: Done

test/pytest/test_qkeras.py
makoeppel: Yes this is need for KerasV3.

docs/intro/status.rst
makoeppel: It should also work in qkeras v3.

hls4ml/model/types.py
makoeppel: Done

hls4ml/model/quantizers.py
makoeppel: Not needed now after fixing the layers. WIll be removed.

hls4ml/converters/keras_v3/qkeras/layer.py
makoeppel: I have now this layer.py which is more or less a base class. I can call it also _base.py if you want.

hls4ml/model/optimizer/passes/qkeras.py
makoeppel: I removed this now since after fixing the layer and activation stuff this is not needed anymore.

hls4ml/converters/keras/qkeras.py
makoeppel: I removed this fully now since all is handled in the v3 paser.

makoeppel:
Thank you for the review!

I addressed all your comments.
Please have a look if its fine now.

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+from hls4ml.converters.utils import IsolatedLayerReader
+
+from ..core import KerasV3LayerHandler
+from .utils import set_default_config
+
+
+class QKerasQActivationHandler(KerasV3LayerHandler):
+    handles = ('qkeras.qlayers.QActivation', 'QActivation')
```
calad0i: Why `QActivation` needs to be defined w/o module? Is it defined via metaprogramming and don't have a fixed path?

hls4ml/converters/keras_v3/qkeras/activation.py
```diff
@@ -0,0 +1,35 @@
...
+        config = layer.get_config()
+        layer_dict = {'config': config, 'class_name': layer.__class__.__name__}
+
+        reader = IsolatedLayerReader(layer)
+        input_shapes = [list(t.shape) for t in in_tensors]
+        input_names = [t.name for t in in_tensors]
+
+        from hls4ml.converters.keras_v2_to_hls import layer_handlers as v2_layer_handlers
+
+        v2_handler = v2_layer_handlers.get(layer.__class__.__name__)
```
calad0i, line 28: If possible, let's avoid using the v2 parser as it shall be eventually removed at some point... No a hard requirement but would be nice if we can do so.

makoeppel:
Note for the tests to work one has to merge https://github.com/fastmachinelearning/example-models/pull/24.

hls4ml/model/quantizers.py
```diff
@@ -140,7 +139,8 @@ def _get_type(self, quantizer_config):
             else:
                 return IntegerPrecisionType(width=width, signed=True)
         else:
-            return FixedPrecisionType(width=width, integer=integer + 1, signed=True)
+            signed = quantizer_config['config'].get('keep_negative', True)
+            return FixedPrecisionType(width=width, integer=integer + int(signed), signed=signed)
```
bo3z: Why this change? Will it affect other quantized frontends?

bo3z:
This looks mostly okay to me.

I only have one question on the quantizers.

I've merged the other PR (example-models), so the tests should now pass. The other failure (in PyTorch) seems to be unrelated (timeout).

Will wait for @JanFSchulte and @calad0i to have another look, but I think we can merge soon.

JanFSchulte:
The test timeout was indeed a fluke, but the test requiring the example model still didn't pass when I just restart the existing test. So lets start the whole test suite from scratch to see if it gets picked up then. 

Otherwise I think I'm happy with this.

hls4ml/model/quantizers.py
makoeppel: I removed this again since it was also only needed because of the old layer handler.

makoeppel:
So the tests fail because the Gitlab CI uses still keras 2 and does not install qkeras-v3. Is this a pre-compiled container?

JanFSchulte:
`hls4ml-testing-keras3` is a predefined conda environment inside our testing container, I think. So using `hls4ml-testing-qkeras-v3` as you do will not work as this doesn't exist, and you should just used the `hls4ml-testing-keras3` one. The missing piece is that you also need to modify this file https://github.com/makoeppel/hls4ml/blob/26a12a79a05bbf175df82dab91f8c8e7d2266ef8/test/pytest/generate_ci_yaml.py to generate the qkeras-v3 tests with the appropriate configuration. You can see in the file that we define a list of tests that use keras 3 and generate them with the correct settings for this case. Unfortunately you will not be able to just add qkeras 3 to that list because installing it together with hgq2 will result in clashes, AFAIK. So you'll need to add an additional case for your tests.

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: I think Keras does not return 'keras.layers.Activation' during serialization.

hls4ml/converters/keras_v3/qkeras/activation.py
makoeppel: I agree but maybe thats something for the future when the v2 parser will be removed.

bo3z: I agree. We could revisit the depracation in the future.

hls4ml/converters/keras_v3/qkeras/activation.py
calad0i: The layer seen by the parser is a python object, and no serialization is used in keras v3 parser though.

bo3z:
There seems to be one unrelated failure in HGQ. @calad0i any ideas?

calad0i:
Looks like usual fluctuation in bit exactness tests, can run again and see what happens...

makoeppel:
> Looks like usual fluctuation in bit exactness tests, can run again and see what happens...

Now all tests passed.

hls4ml/model/quantizers.py
```diff
@@ -170,7 +169,9 @@ def __init__(self, config, xnor=False):
 
     def __call__(self, data):
         data = np.array(data, dtype='float32')
-        y = self.quantizer_fn(data).numpy()
+        y = self.quantizer_fn(data)
+        if hasattr(y, 'numpy'):
```
calad0i: Maybe use `isinstance` check here? Can only tf variable or numpy array here.

hls4ml/converters/keras_v3/qkeras/activation.py
calad0i: This remains

makoeppel:
Hi, should we merge this one? There is an error for the openml.org dataset:

ERROR test_qkeras.py::test_accuracy[latency] - urllib.error.HTTPError: HTTP Error 504: Gateway Time-out

During the tutorial last week we also had an issue with this website. But the tests work when the download of the dataset works.

calad0i:
Minor stuff remains (the qactivation one), otherwise good from my side

makoeppel:
I addressed the last minor changes now.

hls4ml/model/quantizers.py
```diff
@@ -5,6 +5,7 @@
 """
 
 import numpy as np
+import tensorflow as tf
```
calad0i: Move the import `tensorflow` to inside `__call__`, or tf will be needed as a global dependency.

makoeppel: Done

calad0i (approved):

bo3z (approved):

bo3z:
Pending the fix of merge conflicts, this PR can be merged.

PR #1343, opened 2025-07-24
XLS backend

Girjoaba:
# Description

:memo: The implementation of a new backend based on the High-Level Synthesis (HLS) toolchain [XLS](https://google.github.io/xls/) from Google. The purpose of this addition is to increase vendor availability and benefit from the developments that XLS brings to HLS field. XLS is an open-source project, being suitable with the no license requirements of hls4ml.

The changes include the addition of all components required of a new backend: an inherited `xls_backend` class, an `xls_writer` and corresponding optimization passes.

**Dependencies**:

For the XLS backend to work, the original XLS project must be built as described [here](https://google.github.io/xls/) under **Install Latest Release**.  Afterwards, in the hls4ml config object the path to the installed `bazel-bin` must be specified as follows:
```Python
config['xls_bazel_bin_path'] = '$HOME/xls/bazel-bin'
```

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] A new research paper code implementation
- [ ] Other (Specify)

## Tests

Currently, the backend was included in 2 pytests: `test_keras_api.py` and `test_activations.py`. 

The ELU function in `test_keras_api.py` was changed to a RELU for now. The original tests will be reverted back and the backend will be included in all required tests. 

**Test Configuration**:

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have installed and run `pre-commit` on the files I edited or added.
- [ ] I have added tests that prove my fix is effective or that my feature works.

PR #1475, opened 2026-05-14
XLS backend

### LINKED ISSUE #1443: `test_softmax.py` does not test `argmax` and `latency` implementations; `latency` fails
## Quick summary
1. `test_softmax.py` does not actually test non-default implementations (`argmax` and `latency`).
2. If we fix that, tests for `latency` fail, reporting low accuracy.

### Steps to Reproduce

2. Checkout the main branch, commit hash: 6900257f
3. Run `pytest test/pytest/test_softmax.py`
4. Go e.g. to `test/pytest/test_softmax_test_softmax_9,6-input_shape6-18,8-io_parallel-False-latency-Vivado/parameters.h`, line 29, see:
```
    static const nnet::softmax_implementation implementation = nnet::softmax_implementation::stable;
```
instead of expected `nnet::softmax_implementation::latency`.

5. Now, if we replace the line
https://github.com/fastmachinelearning/hls4ml/blob/6900257ff794958b834afbaa41e7c75821d3094e/test/pytest/test_softmax.py#L47
with
```
    cfg['LayerName']['softmax']['implementation'] = strategy
```
then all `latency` tests fail with messages like:
```
Accuracy hls4ml relative to keras: 0.1908
```

6. Now, if we replace https://github.com/fastmachinelearning/hls4ml/blob/6900257ff794958b834afbaa41e7c75821d3094e/test/pytest/test_softmax.py#L19
with
```
    return np.clip(d, -32, 0)
```
the the tests pass.

(Catapult backend fails on my laptop because it doesn't create .so file, but that's unrelated).

vasdommes:
# Description

This PR adds XLS backend. It is based on PR #1343, with most of the code rewritten and new features added.

[Google XLS](https://google.github.io/xls/) is an open-source (Apache 2) High Level Syntesis toolchain that produces an RTL (Verilog or SystemVerilog) design from a high-level description (DSLX or C++).

Adding XLS as a new hls4ml backend allows to generate RTL without vendor-specific dependencies and benefit from the developments that XLS brings to HLS field.

### XLS workflow

XLS backend performs the following transformations:

- `write()`: hls4ml representation -> [DSLX](https://google.github.io/xls/dslx_reference/) project
- `compile()`: DSLX -> [XLS IR](https://google.github.io/xls/ir_overview/) -> [Optimized](https://google.github.io/xls/optimizations/optimizations/) XLS IR
- `build()`: Optimized XLS IR -> [(System)Verilog](https://google.github.io/xls/codegen_options/) ~-> IP~ (**UPD**: Vivado call [removed](https://github.com/fastmachinelearning/hls4ml/pull/1475/commits/ed74aff263a9f31f62ffb9d74ec9ddd8c81e621b), now we produce only Verilog.)

DSLX -> IR -> (System)Verilog conversion is done by XLS.
IP can generated from Verilog file manually e.g. via Vivado or another vendor.

### XLS features

XLS backend supports the following layers: `Input`, `ApplyAlpha`, `BatchNormalization`, `Dense`, `Conv1D`, `DepthwiseConv1D`, `Conv2D`, `DepthwiseConv2D`, `Pooling1D`, `Pooling2D`, `GlobalPooling1D`, `GlobalPooling2D`, `Merge`, `Concatenate`, `Dot`, `Activation`, `HardActivation`, `ParametrizedActivation`, `PReLU`, `Reshape`, `Softmax`, `Transpose`, `TernaryTanh`.

You can override default [codegen options](https://google.github.io/xls/codegen_options/) as follows:
```python
config = hls4ml.utils.config_from_keras_model(model)
# This sets hls_model.config['XLSCodegenFlags']
    model, hls_config=config, backend='XLS',
    xls_codegen_flags={'delay_model': 'asap7', 'generator': 'pipeline', 'use_system_verilog': False}
)
```

DSLX standard library has only signed [FixedPoint](https://github.com/google/xls/pull/2066) type (similar to `ap_fixed`). Thus, unsigned types are not supported.

Currently, XLS backend implements only `IOType: io_parallel`. `Strategy` is ignored.
All operations are fully unrolled.

`io_stream` could be implemented via DSLX [procs](https://google.github.io/xls/dslx_reference/#communicating-sequential-processes-aka-procs). @calad0i and I are going to work on that after finishing this PR.

### Other changes
I made some minor changes in non-XLS code:
- Fixed #1443, since it was needed to test all softmax implementations in XLS.
- [Added](https://github.com/vasdommes/hls4ml/commit/2ec9ef35e9b30c0a6c30132a4f2e84e9d4c86b36#diff-e5cc32b6085fc8e732ef06de16c8d1f6ba56688906f9a6f70bd4d6c2932d2d65R824-R828) hook to `ModelGraph` to call custom `backend.get_top_function()`. This is needed for XLS because it uses optimized XLS IR file instead of `.so` library generated by other backends.
- [Updated](https://github.com/vasdommes/hls4ml/commit/878df2402da0fa223e589479e2e5d3cb8dec58a4) `docs/ir/attributes.rst`. Aside from adding XLS, this commit some other missing layers and attributes.

### Dependencies

XLS backend uses [xls-python](https://calad0i.github.io/pyxls/) to access XLS API. It is enabled by dependency group `xls`:
```
pip install hls4ml[xls]
```
`xls-python` comes with batteries (`libxls.so` and [DSLX standard library](https://google.github.io/xls/dslx_std/)) included, no separate XLS installation is required.
The code has been tested for the version `xls-python=0.1.9875`.

### Known issues

XLS doesn't work with `Dense` layer imported from PyTorch `Linear` layer because of shape mismatch: PyTorch stores `Linear` weights as `(out_features, in_features)`, while hls4ml `Dense` layers use the Keras-style layout `(in_features, out_features)`.

Repro: add XLS backend to [test_pytorch_api.py/test_squeeze](https://github.com/vasdommes/hls4ml/blob/2425a5ff76906287f719bf41f0e7f1142265253a/test/pytest/test_pytorch_api.py#L653) and run the test.

Note that the weights in this test are constant, and other backends flatten them without checking shape.
So, it is unclear whether they handle this situation correctly or not.

## Type of change

- [x] Documentation update

## Tests

>

XLS has been added to the following tests:
`test_activations.py`, `test_auto_precision.py`, `test_binary_cnn.py`, `test_causalpadding.py`, `test_depthconv1d.py`, `test_depthconv2d.py`, `test_keras_api.py`, `test_keras_v3_api.py`, `test_merge.py`, `test_multi_dense.py`, `test_pointwiseconv.py`, `test_pooling.py`, `test_pytorch_api.py`, `test_reshape.py`, `test_sepconv1d.py`, `test_sepconv2d.py`, `test_softmax.py`.

### Test Configuration

Add `xls` dependency, e.g.
```
pip install .[da,testing,testing-keras2,sr,optimization,xls]"
# or: pip install .[da,testing,testing-keras3,sr,xls]"
```
and run tests, e.g.:
```
pytest test/pytest --randomly-dont-reset-seed -k XLS
```

### Notes on performance

Some test cases are very slow for XLS (e.g. ~30 minutes vs ~10 seconds on other backends).
This happens because XLS generates (in `model.compile()`) and uses (in `model.predict()`) an optimized XLS IR code, where all loops are fully unrolled. The resulting file can be huge and thus slow for the likes of `Conv2D`.

During development, I made test faster by reducing dimensions in some tests.
For example, in [test_keras_api.py/test_conv2d](https://github.com/vasdommes/hls4ml/blob/70c7b7bab5c8e6b3080bfd44efcd07635405bb66/test/pytest/test_keras_api.py#L240) I replaced
```
input_shape = (28, 28, 3)
filters=32
```
with
```
input_shape = (14, 14, 3)
filters=8
```

~I haven't pushed such changes, but that could be one of the ways of speeding things up.~

**UPD**: I reduced XLS dimensions in many tests, see [8578a62](https://github.com/fastmachinelearning/hls4ml/pull/1475/commits/8578a62eb17a3b08d1aea0f047d33ca480b7f930) and [75a04fb](https://github.com/fastmachinelearning/hls4ml/pull/1475/commits/75a04fb6ce3639f146d46f42002b12d528384276).

## Checklist


vasdommes:
I reduced XLS dimensions in many tests to make them fast enough, but the test  https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75768174 failed due to 2hr timeout.

```
$ pytest $PYTESTFILE -rA --cov-report xml --cov-report term --cov=hls4ml --junitxml=report.xml --randomly-seed=42 --randomly-dont-reorganize --randomly-dont-reset-seed
============================= test session starts ==============================
platform linux -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0
pytest-randomly: reseed with 42
Using --randomly-seed=42
rootdir: /builds/fastmachinelearning/hls4ml
configfile: pyproject.toml
plugins: anyio-4.12.0, cov-7.0.0, randomly-4.0.1
collected 227 items / 1 skipped
test_einsum_dense.py ..................                                  [  7%]
test_keras_v3_api.py ...........s....................................... [ 30%]
.Terminated
WARNING: step_script could not run to completion because the timeout was exceeded. For more control over job and script timeouts see: https://docs.gitlab.com/ee/ci/runners/configure_runners.html#set-script-and-after_script-timeouts
ERROR: Job failed: execution took longer than 2h0m0s seconds
```

That's surprising - on my laptop, `test_einsum_dense` finishes in 3 min and `test_keras_v3_api` - in 16 min (including 4 min for XLS).

calad0i:
This timeout looks like CI issue, not something real in the code.

vloncar:
If it is not doing synthesis (and it seems that it skips that path), what does it do that takes a long time?

calad0i:
> If it is not doing synthesis (and it seems that it skips that path), what does it do that takes a long time?

I feel like that the slowness is not due to the tests themselves, more likely the runner itself stalls on random tests one in a while...

https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75680057
https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75627913
https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75482738
https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75329237
https://gitlab.cern.ch/fastmachinelearning/hls4ml/-/jobs/75320437

vloncar:
Hmm, seems that all tests always run only on -05, other nodes are never used. Gitlab reports them as "stale". Turns out they are not configured right, always reporting as -05, but only the real -05 is used. I've reconfigured 06 and 08 to be separate, they are in the pool now. Sadly 07 doesn't respond, will need to ping an openlab admin to configure that one. Hopefully then it will be a bit less crowded and tests will run faster

test/pytest/test_softmax.py
```diff
@@ -11,16 +11,18 @@
 
 
 @pytest.fixture()
-def generate_data(input_shape):
+def generate_data(input_shape, implementation):
     shape = (5000, *input_shape)
     d = np.random.normal(0, 2, shape)
     modify_entries = np.random.randint(0, 1, shape) < 0.05
```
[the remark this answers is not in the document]
vasdommes: @calad0i did you mean `np.random.rand(*shape)` when you wrote this line?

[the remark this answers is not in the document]
calad0i: Probably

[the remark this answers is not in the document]
vasdommes: Several `test_softmax` cases fail due to low accuracy if I make this change (on both `main` and `xls_backend` branches) so I'll leave it out of the current PR's scope.
```
>       assert acc_hls4ml >= 0.98
E       assert 0.9638 >= 0.98
```

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+        width=precision.width,
+        integer=integer,
+        signed=precision.signed,
+        rounding_mode=rounding_mode,
+        saturation_mode=saturation_mode,
+    )
+    # Only signed types are supported in XLS
+    if not fixed_precision.signed:
+        if not allow_unsigned:
+            raise ValueError(f'Expected signed precision, got: {precision}')
```
calad0i: Looks like bit exactness issue when data is defined as unsigned type? Consider warn instead of raise since it is a bit-exactness compromise instead of something unrecoverable.

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+        return XLSIntegerType(width=32, signed=True)
+
+
+class XLSFixedPointType:
+    def __init__(self, num_bits, binary_exponent):
+        self.num_bits = num_bits
+        self.binary_exponent = binary_exponent
+
+    @classmethod
+    def from_precision(cls, precision: PrecisionType, allow_unsigned: bool = False):
```
calad0i: Why do you need a new precision type here? The ${BACKEND}WhateverType in other backends are converted by metaclass to add write method, but this one seems to be a plain thin wrapper?

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+        if not allow_unsigned:
+            raise ValueError(f'Expected signed precision, got: {precision}')
+        fixed_precision.signed = True
+        fixed_precision.width += 1
+        fixed_precision.integer += 1
+
+    return fixed_precision
+
+
+def float_to_significand(
```
calad0i, line 132: This is essentially quantize and represent as raw integer. You are assuming the overflow/rounding modes to be WRAP and TRN here, which is not necessarily true. The XOR case looks wrong to me, and I guess we can drop it directly as unsupported.

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+def as_xls_fixed_point_type(type: XLSFixedPointType | PrecisionType, allow_unsigned: bool = False) -> XLSFixedPointType:
+    if isinstance(type, XLSFixedPointType):
+        return type
+    return XLSFixedPointType.from_precision(type, allow_unsigned)
+
+
+# 1d array type. TODO make it explicitly multidimensional?
+class XLSArrayType:
```
calad0i: In the IR, consider keep the arrays as-is in their original shape? Flattening in raw major order is trivial with .ravel() and you can do that in writter.

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+        ]
+        optimization_flow: str = register_flow('optimize', optimization_passes, requires=[init_flow], backend=self.name)
+
+        xls_attributes = [
+            'xls:build_tables',
+            'xls:build_attr',
+        ]
+        xls_attributes_flow: str = register_flow('xls', xls_attributes, requires=[optimization_flow], backend=self.name)
+
+        # TODO: stamp is currently unused, shall we add it to myproject.x, myproject.ir, myproject.opt.ir, ...?
```
calad0i: stamp is not useful for XLS. It is added just to avoid libname collision when using ctypes. Since XLS emulation is jit within libxls, no need for that.

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+
+@functools.lru_cache(maxsize=1)
+def import_xls():
+    try:
+        return importlib.import_module('xls')
+    except ModuleNotFoundError as e:
+        raise ModuleNotFoundError(
+            "XLS backend requires optional dependency 'xls'. "
+            "Please install hls4ml with XLS extras (or install package 'xls')."
+        ) from e
```
calad0i: Import xls directly in the caller and decorate the caller with @requires('xls)

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+        with open(f'{path_no_ext}.ir', 'w') as ir_file:
+            ir_file.write(ir_text)
+
+        opt_ir_text = xls.optimize_ir(ir=ir_text, top=XLSBackend._ir_top_function_name(model))
+        with open(f'{path_no_ext}.opt.ir', 'w') as opt_ir_file:
+            opt_ir_file.write(opt_ir_text)
+
+        # This object can be heavy, so we don't want to cache it unless we call predict().
+        if hasattr(model, '_xls_top_function'):
+            del model._xls_top_function
```
calad0i: Adding top fn as attr was needed for other backends only because they require ctypes binding. Since you are using xls jit, making it an ordinary python function is ok.

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+        x, precision: FixedPrecisionType | Iterable[FixedPrecisionType], dtype: np.typing.DTypeLike
+    ) -> ArrayLike | tuple[ArrayLike, ...]:
+        xls = import_xls()
+        # x: xls.Value
+        match x.get_kind():
+            case xls.c_api.ValueKind.BITS:
+                assert isinstance(precision, FixedPrecisionType), (
+                    f'Precision must be FixedPrecisionType, got {type(precision)}'
+                )
+                return XLSBackend._bits_to_int(x.get_bits()) / (2**precision.fractional)
```
calad0i, line 280: missing overflow and rounding, or ok you are sure that these won't happen

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+            out_precision = [output_var.type.precision for output_var in output_vars]
+            if len(out_precision) == 1:
+                out_precision = out_precision[0]
+            dtype = np.asarray(inputs[0]).dtype
+            output = XLSBackend._xls_ir_to_float(ir_output, out_precision, dtype)
+            # This is the case when len(output_vars) == 1
+            if not isinstance(output, tuple):
+                output = (output,)
+            for i in range(len(output_vars)):
+                outputs[i][:] = np.reshape(output[i], -1)
```
calad0i, line 340: Make this directly the top function; the object to be added in .compile() is the jitted fn/pkg object.

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,429 @@
...
+
+        return top_function
+
+    def build(
```
calad0i, line 348: build() is not needed for XLS. Calling vivado is not correct as XLS emitted Verilog may or may not be used w/ verilog. If you want, make it emit the actual Verilog and **skip** vivado call.

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+# XLS values
+
+
+class XLSInteger:
+    def __init__(self, type: XLSIntegerType | str, value: int | str):
+        self.type = type
+        self.value = value
+
+    @classmethod
+    def u32(cls, value: int | str):
```
calad0i: Seems to be dead method never called? Same for s32.

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+    def __str__(self):
+        return f'{self.element_type}[{self.size}]'
+
+
+# XLS values
+
+
+class XLSInteger:
+    def __init__(self, type: XLSIntegerType | str, value: int | str):
+        self.type = type
```
calad0i: What is type here? Why can it be str?

hls4ml/report/xls_report.py
```diff
@@ -0,0 +1,49 @@
...
+
+def _get_project_name(path) -> str:
+    project_path = Path(path) / 'firmware'
+    sv_files = list(project_path.glob('*.sv'))
+    if not sv_files:
+        raise FileNotFoundError(f'No .sv files found in {project_path}')
+    return sv_files[0].stem
+
+
+def parse_xls_report(hls_dir) -> dict:
```
calad0i: This is vivado report parser. Delete

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+    Returns: x * 2^precision.fractional
+    """
+    if not np.isscalar(x):
+        if not isinstance(x, np.ndarray) or x.dtype.kind != 'f':
+            x = np.asarray(x, dtype=np.float64)
+
+    if isinstance(precision, XnorPrecisionType):
+        # hls4ml stores XNOR weights as bits {0,1};
+        # We convert it to XLS FixedPoint {-1, 1}
+        x = np.where(x == 0, -1, x)
```
calad0i: `x = np.where(x == 0, -1, 1)` instead?

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,501 @@
...
+
+    precision = to_signed_fixed_precision(precision, allow_unsigned)
+
+    width = precision.width
+    frac = precision.fractional
+    scale = 2**frac
+    # TODO support different saturation and rounding modes
+    significand = np.round(x * scale).astype(np.int64)
```
calad0i: This assumes WRAP and RND_CONV. You can use `from quantizers import get_fixed_quantizer_np` and `get_fixed_quantizer_np(round_mode=..., overflow_mode=...)(x, k=, i=, f=)` for saturation and overflow mode supports.

hls4ml/backends/xls/xls_backend.py
vasdommes: Do you mean the case when a fixed-precision number exceeds `sys.float_info.max`?

hls4ml/backends/xls/xls_types.py
vasdommes: Currently we don't have a warning/error for constants like weights and biases (allow_unsigned = True).
But we throw an error (which should be a warning, as you suggest) e.g. when input/output variable type is unsigned.

@jmitrevs noted that we could even skip all warnings and simply mention in the docs that all unsigned types are converted to signed.
In that case, I could simplify the code a bit by removing the `allow_unsigned` parameter (21 usage) .

What do you think? Are such warnings helpful?

hls4ml/backends/xls/xls_types.py
vasdommes: The class `XLSFixedPointType` and its `__str__()` method plays a role similar to `ACFixedPrecisionDefinition` and its `cpp_definition()` method.

I could try and refactor the code to make it similar to `*Definition` / `*Converter` / `TransformTypes` logic used in other backends.

hls4ml/backends/xls/xls_types.py
vasdommes: Array is kept in the original shape. 1d here simply means that in DSLX we represent a multidimensional array as an array of arrays (not as a flat array + shape).

hls4ml/backends/xls/xls_backend.py
calad0i: No, I meant `fixed_point_number = int_repr / 2^frac_bits` only when you are sure that the incoming number fits in the range of the fixed point number w/o rounding or overflowing.

hls4ml/backends/xls/xls_types.py
vasdommes: str was added for convenience, e.g. `XLSInteger(type='u32', value=42)`. It is not actually used now, will remove.

hls4ml/backends/xls/xls_backend.py
vasdommes: `fixed_point_number` equals `int_repr / 2^frac_bits` by definition, doesn't it?

It could be out of range for the `SAT_SYM` overflow mode, but every layer in IR performs rounding and overflowing before returning an answer.

Generally, rounding/overflow could also matter if the output precision of XLS IR differed from the `precision` argument passed to `_xls_ir_to_float()`. But it should never happen - precision always comes from `model.get_output_variables()`.

hls4ml/backends/xls/xls_types.py
vasdommes: These are equivalent since `x` contains only `{0,1}`.
To make it clearer, I switched to `x = np.where(x == 0, -1, 1)` and added an assertion.

hls4ml/backends/xls/xls_types.py
vasdommes: Added correct overflow/rounding via `quantizers`.

XNOR works, tested by `test_binary_cnn`.

hls4ml/backends/xls/xls_types.py
vasdommes: Refactored [d59fd1f](https://github.com/fastmachinelearning/hls4ml/pull/1475/commits/d59fd1f73ec60e44a7e19f5375efd77167d13f43)

hls4ml/backends/xls/xls_types.py
vasdommes: Now we convert everything to signed types and print warnings if model inputs/outputs were unsigned [0ad6cdd](https://github.com/fastmachinelearning/hls4ml/pull/1475/commits/0ad6cdddcfdedc3ff31631c407b7edbe5be967b4)

hls4ml/backends/xls/xls_backend.py
```diff
@@ -0,0 +1,401 @@
...
+            dtype = np.asarray(inputs[0]).dtype
+            output = XLSBackend._xls_ir_to_float(ir_output, out_precision, dtype)
+            # This is the case when len(output_vars) == 1
+            if not isinstance(output, tuple):
+                output = (output,)
+            for i in range(len(output_vars)):
+                outputs[i][:] = np.reshape(output[i], -1)
+
+        # TODO: this duplicates ModelGraph._get_top_function().
+        # NB: ctype is not used in XLS, but it is required by ModelGraph._predict
```
calad0i: Maybe we can warp .predict() directly instead to bypass all of these. Likely it would be cleaner that way

hls4ml/backends/xls/xls_layer_util.py
```diff
@@ -0,0 +1,272 @@
...
+                return [
+                    XLSConstDefinition(name='ACTIVATION_PARAM', value=XLSFixedPointDefinition.from_float(value, precision))
+                ]
+        case _:
+            pass
+    return []
+
+
+def xls_func_name(node: Layer) -> XLSQualifiedName:
+    match node.class_name:
```
calad0i, line 185: Could we do something like f'{node.class_name}{node.index}' to save the whole case-match?

hls4ml/backends/xls/xls_types.py
```diff
@@ -0,0 +1,572 @@
...
+                case 'PReLU':
+                    expected_shape = (input_var.shape[0],)
+                case _:
+                    raise ValueError(f'Unsupported weights for layer {node.class_name}')
+
+        if expected_shape is not None:
+            assert shape_tuple(data.shape) == expected_shape, (
+                f'Weights shape mismatch: expected {expected_shape}, got {data.shape}'
+            )
+
```
calad0i, line 496: Asserting in variable definition is probably a bit misplaced, and we likely don't really need to assert these outside the corresponding layer specific handlers. Maybe remove?

hls4ml/model/graph.py
```diff
@@ -820,7 +820,12 @@ def _compile(self):
             dlclose_func(self._top_function_lib._handle)
         self._top_function_lib = ctypes.cdll.LoadLibrary(lib_name)
 
-    def _get_top_function(self, x):
+    def _get_top_function(self, x, *args, **kwargs):
```
calad0i, line 823: If warp predict() directly, this part could be removed.

hls4ml/backends/xls/xls_backend.py
vasdommes: Actually, I can simply return `ctype = None`.

hls4ml/model/graph.py
vasdommes: The problem is that `_get_top_function` is called inside `_predict`.
https://github.com/fastmachinelearning/hls4ml/blob/2ec9ef35e9b30c0a6c30132a4f2e84e9d4c86b36/hls4ml/model/graph.py#L881-L882

I see two options here:
1. Keep `get_top_function` hook.
2. Remove `get_top_function` hook, lift `self._get_top_function(x)` call from `_predict()` to `predict()`. Then every  `predict()` implementation would need to make an extra `self._get_top_function(x)` call:
```python
    def predict(self, x)
        # <...>
        top_function, ctype = self._get_top_function(x)
        return self._predict(x, top_function, ctype)
```

Option 1 seems lees intrusive to me, although it adds three extra lines of code.

hls4ml/backends/xls/xls_types.py
vasdommes: Ideally yes, everything should be handled earlier. But if something goes wrong either with XLS or with layer handling, as it does in PyTorch Linear layer case:
https://github.com/fastmachinelearning/hls4ml/blob/9fda567412953e498de0948e0805270719c0f968/test/pytest/test_pytorch_api.py#L659-L662

then assertion error
> AssertionError: Weights shape mismatch: expected (1, 5, 3), got (1, 3, 5)

is more readable than XLS error
> layer_6_linear.x:50:5-50:27 TypeInferenceError: type mismatch: FixedPoint<16, -10>[5] vs. FixedPoint<16, -10>[LAYER6_OUT_DIM_1]. The body of function `transform_2d` does not actually return the function's declared return type, which is `FixedPoint<16, -10>[LAYER6_OUT_DIM_1][LAYER6_OUT_DIM_0]`

hls4ml/backends/xls/xls_layer_util.py
vasdommes: No, here we need qualified names for library functions, e.g. `activations::softmax_stable` or `pooling::global_pooling_2d`.

We could put all layer functions into one giant DSLX module (note that DSLX does not allow re-exports), but that's ugly. Also simply returning `node.class_name` and removing `ValueError` would lead to more obscure DSLX errors if layer is not supported.

vasdommes:
`test_softmax` failed for Quartus with `implementation=latency`:
```
FAILED test_softmax.py::test_softmax[16,6-input_shape4-9,6-io_parallel-False-latency-Quartus-standalone] - assert 0.5308 >= 0.98
FAILED test_softmax.py::test_softmax[16,6-input_shape5-9,6-io_stream-False-latency-Quartus-standalone] - assert 0.5216 >= 0.98
FAILED test_softmax.py::test_softmax[9,6-input_shape6-18,8-io_parallel-False-latency-Quartus-standalone] - assert 0.5636 >= 0.98
FAILED test_softmax.py::test_softmax[9,6-input_shape7-18,8-io_stream-False-latency-Quartus-standalone] - assert 0.5598 >= 0.98
```

This became visible because of my fixes for #1443 (enable `latency` and `argmax` in `test_softmax`), which are not directly related to XLS.
I can revert these changes to let CI tests pass for now. @jmitrevs made a similar [fix](https://github.com/fastmachinelearning/hls4ml/commit/5dadf1a4da01083a664615e86006adfa8105d5b8), included in PRs #1476 and #1494.

P.S. Other CI failures seem to be due to timeout (probably unrelated to XLS, same tests work on my machine).

JanFSchulte:
Hi @vasdommes, we should not introduce breaking tests, so I would opt to revert this changes for the moment and fix them in the relevant softmax PRs. 

Otherwise I think this PR is ready. @calad0i are you happy with how your last comments have been addressed? If so, I think we could merge this.

JanFSchulte:
pre-commit.ci autofix

calad0i (approved):

PR #1478, opened 2026-05-21
(Partial) support for SR backend with Vitis HLS

bo3z:
# Description

> :memo: The SR backend doesn't really work with Vitis HLS. There are three reasons for this, two of which are addressed by this PR. The third on is fundamentally un-adressable.

1. The build function of the SR backend always calls vivado_hls, which doesn't exist for Vitis HLS. The current fix calls the appropriate tool (vitis-run or vivado_hls), depending on the compiler passed during convert_from_symbolic_expression.

2. Some math libraries changed versions (from _v0 to _v1) in Vitis 2022.2. Given that hls4ml only supports Vitis HLS > 2022.2, the new code checks which compiler was specified (vivado_hls or vitis_hls) and appropriately sets the library to link against in build_lib.sh.

3. When running hls_model.compile(...), the process looks for Vitis HLS math include header and libraries. These are closed-source/proprietary, and so, compile(...) will faill if no valid Vitis HLS is available (which doesn't happen on other backends; other backends can run every step except build without an HLS compiler). 

I don't see a solution for the 3rd issue, but this PR should at least allow using the SR backend for users for users that have Vitis HLS installed on their system (which could be quite a few users, since Vivado HLS has been deprecated for ~6 years now). This PR also allows us to run the new SR notebook in the proposed tutorial revamp end-to-end without any issues.

## Type of change

## Tests
- The existing SR tests (which default to Vivado HLS) should continue working as expected.
- The "new" feature (Vitis HLS support) has been validated in the new tutorial with Vitis 2024.1: https://github.com/bo3z/hls4ml-tutorial/blob/tutorial-revamp-2026/6_more_models/6b_symbolic_regression.ipynb

## Checklist


hls4ml/converters/__init__.py
```diff
@@ -438,17 +439,24 @@ def convert_from_symbolic_expression(
         part (str, optional): The FPGA part. If set to `None` a default part of a backend will be used.
         clock_period (int, optional): Clock period of the design.
             Defaults to 5.
-        compiler (str, optional): Compiler to use, ``vivado_hls`` or ``vitis_hls``. Defaults to ``vivado_hls``.
-        hls_include_path (str, optional): Path to HLS inlcude files. If `None` the location will be inferred from the
-            location of the `compiler` used. If an empty string is passed the HLS math libraries won't be used during
+        hls_compiler (str, optional): HLS compiler to use. Must be ``'vivado_hls'`` or ``'vitis_hls'`. Defaults to ``'vivado_hls'``.
```
vloncar: Can we add support for `vitisrun` that is used in newer versions? And we could do a performance eval and if it works as well with Vitis as originally with Vivado, we can drop Vivado support altogether.

bo3z: In my mind `vitisrun` is/was the name of the executable. The compiler is still Vitis HLS. But yes, we can make it more consistent and just have vitisrun (since we don't support any versions of Vitis HLS for which vitisrun doesn't exist).

How would we know if it works as well? Do we have some baseline results? I personally wouldn't have access to a machine with Vivado HLS.

vloncar: indeed it may make sense to just call them `vivado` and `vitis` instead of the actual executable.

regarding vivado hls, i see two options: if you have a licence at eth (check `vlm` with local license server if you still have a valid one) i can give you a containerized version that you can run without installation. other option is to use cvmfs if you have a cern account and can log in to lxplus.

JanFSchulte:
pre-commit.ci autofix

brinus:
The vitis_hls compiler path hardcodes vitis-run, which was introduced in Vitis 2023.1. On Vitis ≤ 2022.2, vitis-run is not available even though vitis_hls is — so users on older toolchains will hit a "command not found" error despite passing `hls_compiler='vitis_hls'`.
A related fix was submitted in PR #1493  for the main `VitisBackend.build()` with the same root cause. The same shutil.which fallback would apply here:

```python
if hls_compiler == 'vitis_hls':
  if shutil.which('vitis-run'):
    build_command = 'vitis-run --tcl build_prj.tcl --mode hls'
  elif shutil.which('vitis_hls'):
    build_command = 'vitis_hls -f build_prj.tcl'
  else:
    raise Exception('...')
```

Happy to open a follow-up PR or amend this one if useful.

JanFSchulte:
Looks basically ready to merge, except for this issue of the Vitis executable name. We are officially supporting Vitis starting with 2022.2, so hard-coding vitis-run would indeed break that and we should at the very least update the docs at https://fastmachinelearning.org/hls4ml/intro/status.html to make that official.

JanFSchulte (approved):

PR #1479, opened 2026-05-22
Improve test_conv1d layer checks for io_stream + same padding

frabazz:
# Description
Previously, `test_conv1d` skipped all layer-name and layer-property assertions whenever the backend was Vivado/Vitis with `io_type='io_stream'` and `padding='same'`. In that configuration Vivado/Vitis inserts an extra padding layer after the input, which shifts every subsequent layer index by one.

Instead of skipping the checks entirely, this PR introduces an index offset `off` (1 in that specific case, 0 otherwise) so that all layers except the inserted padding layer are still validated. The padding layer itself is not checked because it lacks attributes such as `name`.

As a result, layer-name and property checks now also run in the `io_stream` + `same` configuration, restoring coverage that was previously dropped.

## Type of change
- [x] Other (Specify): Test coverage improvement

## Tests
The updated `test_conv1d` passes correctly in all configurations, including Vivado/Vitis with `io_type='io_stream'` and `padding='same'`.

## Checklist


JanFSchulte:
Looks good, thanks. Merging.

JanFSchulte (approved):

PR #1480, opened 2026-05-23
Fix stable softmax exp_table indexing sign-bit inefficiency

niteshg97:
Fix stable softmax LUT indexing to avoid wasting the sign bit for nonnegative normalized inputs.

# Description

Problem:-

the stable softmax implementation computes x_max - x_i, which is guaranteed to be nonnegative, however, the existing LUT indexing helper slices the sign bit as part of the exp_table address generation, reducing effective LUT utilization and precision.

Fix:-

Introduce a specialized unsigned lookup indexing helper for stable softmax inputs and use it only in the stable softmax path.,

this preserves existing behavior for latency and legacy implementations while improving LUT address utilization for stable softmax.

### DIFF
--- a/hls4ml/templates/vivado/nnet_utils/nnet_activation.h
+++ b/hls4ml/templates/vivado/nnet_utils/nnet_activation.h
@@ -141,7 +141,6 @@ template <class data_T, unsigned table_size> inline float softmax_real_val_from_
 template <class data_T, unsigned table_size> inline unsigned softmax_idx_from_real_val_unsigned(data_T x) {
     // Slice the top N bits to get an index into the table
     static constexpr int N = ceillog2(table_size); // number of address bits for table
-
     // skip sign bit because inputs are guaranteed nonnegative
     ap_uint<N> y = x(x.width - 2, x.width - N - 1);    // slice the top N bits of input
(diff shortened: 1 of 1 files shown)


jmitrevs:
Note that we also have #1476. Does that fix the issues you are looking at? With that PR the default values become unsigned.

niteshg97:
> Note that we also have #1476. Does that fix the issues you are looking at? With that PR the default values become unsigned.

Thanks for pointing that out, i took a look at #1476, and it seems like the changes to the stable softmax type handling might already cover the sign-bit indexing issue I was trying to fix here.

Also, it looks like the GitLab failures were already present before this PR, so I’ll wait for your guidance on whether this change still adds anything on its own or if #1476 already fully addresses it.

jmitrevs:
I think it's best to use an unsigned type for the unsigned value and not a signed type and ignore the sign bit. So I'll close this PR.

niteshg97:
> I think it's best to use an unsigned type for the unsigned value and not a signed type and ignore the sign bit. So I'll close this PR.

that makes sense..... using an unsigned type directly for values that are always nonnegative is definitely a cleaner approach than handling the sign bit during indexing, thanks for the clarification and for taking the time to review the PR.

PR #1481, opened 2026-05-26
Fix updating dynamic optimizer flows

niteshg97:
# Description

Fixes dynamic flow updates by using the correct `set.add()`API when recording added or removed optimizers.

Previously, calling update_flow() on a callable-backed flow raised an AttributeError because dynamic flow attempted to call `set.put()`. Callable backed flows are used for dynamic optimizer/template flow registration, so this broke flow customization for those flows.

Also adds a focused regression test covering add/remove behavior for a dynamic flow.

## Type of change
Bug fix 

## Tests
 i added a regression test in `test/pytest/test_flows.py` for updating a dynamic flow.

Test command;
```bash
pytest test/pytest/test_flows.py -q

### DIFF
--- a/hls4ml/model/flow/flow.py
+++ b/hls4ml/model/flow/flow.py
@@ -62,10 +62,10 @@ class DynamicFlow(Flow):
         return optimizers
 
     def _add_optimizer(self, opt_name):
-        self._added_optimizers.put(opt_name)
+        self._added_optimizers.add(opt_name)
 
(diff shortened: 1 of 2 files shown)


niteshg97:
pre-commit.ci autofix

vloncar (approved):
Indeed a bug. I can't remember what was the idea behind using `put()` here. Seemed unused for many years

PR #1482, opened 2026-05-28
fix(bit_exact): add produce_kif handler for ZeroPadding1D/2D

LarocheC:
# Description

`ZeroPadding1D` and `ZeroPadding2D` had no `_produce_kif` handler in `hls4ml/model/optimizer/passes/bit_exact.py`, so bit-exact propagation through any model containing them raised `NotImplementedError`. This came up when bringing a TCM-based model with explicit zero-padding (used as a causal-padding alternative for depthwise dilated Conv1D) through `convert_from_keras_model`.

Fix: padding inserts exact zeros, which fit any signed fixed-point grid with `k=0, i=0, f=0`. The output kif is therefore the input kif padded with zeros along the spatial axis/axes (1D pads axis 0, 2D pads axes 0 and 1 — `channels_last`).

The change is local to `bit_exact.py`: two new `@_produce_kif.register` handlers plus the corresponding imports.

## Type of change

## Tests

A new pytest module `test/pytest/test_bit_exact_zeropadding.py` covers three cases:

1. `test_zeropadding1d_produce_kif` — `ZeroPadding1D + Conv1D(valid)` converts via Vitis backend and `hls_model.predict` matches `keras.Model.predict` to `atol=1e-2`.
2. `test_zeropadding2d_produce_kif` — analogous for `ZeroPadding2D + Conv2D(valid)`.
3. `test_causal_via_zeropadding1d_matches_native_causal` — `ZeroPadding1D(pad_left=K-1) + Conv1D(valid)` produces exactly the same Keras output as `Conv1D(padding='causal')` (pure-Keras regression guard).

Without the patch, tests 1 and 2 fail at conversion time with `NotImplementedError`. With the patch, all three pass.

**Test Configuration**:
- Python 3.13, Keras 3.14.1 (torch backend on CPU for local verification; tests use `model.predict()` and so are backend-agnostic).
- `pytest test/pytest/test_bit_exact_zeropadding.py -v` → 3 passed.

## Checklist

- [ ] I have made corresponding changes to the documentation. (No user-facing API change; supported-layer list now covers a path that previously raised.)

### DIFF
--- a/hls4ml/model/optimizer/passes/bit_exact.py
+++ b/hls4ml/model/optimizer/passes/bit_exact.py
@@ -35,6 +35,8 @@ from hls4ml.model.layers import (
     Reshape,
     Softmax,
     Transpose,
+    ZeroPadding1D,
+    ZeroPadding2D,
 )
 from hls4ml.model.optimizer import ModelOptimizerPass, OptimizerPass
 from hls4ml.model.optimizer.passes.hgq_proxy_model import FixedPointQuantizer, UnaryLUT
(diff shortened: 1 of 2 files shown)


test/pytest/test_bit_exact_zeropadding.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+
+os.environ.setdefault('KERAS_BACKEND', 'tensorflow')
+
+import keras  # noqa: E402
+
+import hls4ml  # noqa: E402
+
+
+def _convert(model, output_dir):
```
calad0i: bit_exact flow won't be triggered this way. Either use some hgq layers, or use qkeras and set `bit_exact=True` to trigger it.

hls4ml/model/optimizer/passes/bit_exact.py
```diff
@@ -677,6 +679,34 @@ def _(layer: Embedding):
     return k, i, f
 
 
+@_produce_kif.register
+def _(layer: ZeroPadding1D):
+    k_in, i_in, f_in = get_input_kifs(layer)[0]
+    pad_left = int(layer.attributes['pad_left'])
+    pad_right = int(layer.attributes['pad_right'])
+    # channels_last: kif shape is (in_width, n_chan); pad axis 0.
+    pad_shape = ((pad_left, pad_right),) + ((0, 0),) * (k_in.ndim - 1)
```
calad0i, line 689: Maybe assert for channels_last?

test/pytest/test_bit_exact_zeropadding.py
```diff
@@ -0,0 +1,83 @@
...
+
+    hls_model = _convert(model, str(tmp_path / 'zpad1d'))
+    hls_model.compile()
+
+    rng = np.random.default_rng(0)
+    xin = rng.standard_normal((2, 16, 8)).astype(np.float32)
+    y_keras = model.predict(xin)
+    y_hls = hls_model.predict(xin).reshape(y_keras.shape)
+
+    np.testing.assert_allclose(y_keras, y_hls, atol=1e-2)
```
calad0i: You can check for array_equal if bit_exact pass fires, if no super large bitwidth is used.

calad0i (changes requested):
Minor changes needed, good otherwise.

test/pytest/test_bit_exact_zeropadding.py
LarocheC: I rewrote the tests to use HGQ2 QConv layers around the ZeroPadding layer so the bit_exact pass actually fires (verified it now hits the handlers; reverting them makes the test raise NotImplementedError). Done in d686d725.

hls4ml/model/optimizer/passes/bit_exact.py
LarocheC: Added, matching the existing Conv1D/Conv2D handler.

test/pytest/test_bit_exact_zeropadding.py
LarocheC: Switched to assert_array_equal now.

calad0i (approved):

calad0i:
LGTM, merge after tests pass

calad0i:
One unrelated failure on recurrent nn, merging

PR #1485, opened 2026-06-04
fix(bit_exact): support grouped/depthwise Conv1D/2D in produce_kif

LarocheC:
# Description

The `produce_kif` handler for `Conv1D`/`Conv2D` in the `bit_exact` precision propagation pass assumes a non grouped kernel, where the kernel's second to last axis spans every input channel (`kernel.shape[-2] == n_chan`). A grouped or depthwise convolution stores only `in_per_group` input channels per filter, so the `im2col` buffer no longer matches the full channel input and conversion of a model containing a grouped or depthwise quantized conv fails inside the pass:

```
ValueError: could not broadcast input array from shape (48,) into shape (3,)
```

This is reproduced with an HGQ2 `QConv1D(..., groups=g)` for any `g > 1` (the pass is triggered by the `FixedPointQuantizer` the layer inserts).

The fix detects the grouped case (`kernel.shape[-2] != n_chan`) and processes each group as an independent standard convolution over its own channel slice, then concatenates the per group results along the channel axis. Depthwise is the degenerate `groups == n_chan` (`in_per_group == 1`) case. The non grouped branch is the original code, left byte for byte unchanged, so existing models are unaffected.

Scope: this fixes precision propagation only. After this change a grouped or depthwise model converts instead of crashing, but the converter and backend codegen still emit a dense matmul for the grouped case, so the compiled numerics are not yet correct for `groups > 1`. Routing grouped/depthwise convolutions to a real grouped kernel is a separate follow up; this PR is the prerequisite for it, since any grouped conv support needs the `bit_exact` pass to handle grouped kernels rather than crash on them. The non grouped path is unaffected.

## Type of change

## Tests

Added `test/pytest/test_bit_exact_grouped_conv.py`. For grouped, depthwise and dense (control) `QConv1D` and `QConv2D` models it runs the full `bit_exact` conversion flow (which raised `ValueError` before this fix for `groups > 1`) and asserts that the output precision the pass assigns represents the quantized Keras output exactly, with no rounding and no saturation. The check is done both on the single assigned `result_t` and per channel through `produce_kif`, across the Vivado, Vitis and oneAPI backends.

To reproduce:

```
pytest test/pytest/test_bit_exact_grouped_conv.py
```

**Test Configuration**: Keras 3.14 (torch backend), HGQ2 0.1.8, Python 3.13. Result: 18 passed.

## Checklist

- [ ] I have made corresponding changes to the documentation. (not applicable, internal pass fix)

### DIFF
--- a/hls4ml/model/optimizer/passes/bit_exact.py
+++ b/hls4ml/model/optimizer/passes/bit_exact.py
@@ -486,14 +486,46 @@ def _(layer: Conv1D | Conv2D):
     bias = _bias.data if _bias is not None else 0
     k_in, i_in, f_in = get_input_kifs(layer)[0]
     k_in, i_in, f_in = pad_arrs(layer, 0, k_in, i_in, f_in)
-    k_in, i_in, f_in = im2col(kernel.shape, k_in, i_in, f_in)
-    k_in, i_in, f_in = stride_arrs(layer, k_in, i_in, f_in)
-    kernel = kernel.reshape(-1, kernel.shape[-1])
-    qint_in = QIntervalArray.from_kif(k_in, i_in, f_in)
-    qint_out = qint_in @ kernel
-    qint_out = qint_out + bias
-    k, i, f = qint_out.to_kif()
-    return k.astype(np.int16), i, f
+
+    in_per_group = kernel.shape[-2]
+    n_chan = k_in.shape[-1]
+    if in_per_group == n_chan:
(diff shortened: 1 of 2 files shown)


LarocheC:
Same as in #1486, I've reduced the verbosity: the module docstring and per-test docstrings are down to one or two lines, and the inline commentary in the bit_exact handler is now a single short note on the grouped-case handling.

JanFSchulte (approved):

PR #1496, opened 2026-06-29
fix(pytorch): place unsqueeze axis at the correct position

areporeporepo:
## Description

The PyTorch `unsqueeze` parser placed the new size-1 axis using
`output_shape.index(output_shape[squeeze_dim])` — i.e. it searched for the
*value* of the indexed dimension. `list.index()` returns the first dimension
that shares that size, so the axis was inserted in the wrong position whenever:

- the dim is negative (e.g. the very common `torch.unsqueeze(x, dim=-1)`), or
- the target dimension's size matches an earlier dimension.

Negative dims were also never normalized to torch's accepted range.

### Example (on `main`)

For an input of shape `(N, 4)`:

| call | expected `target_shape` | parsed on `main` |
|------|-------------------------|------------------|
| `torch.unsqueeze(x, dim=-1)` | `[4, 1]` | `[1, 4]` ❌ |

And for `(N, 4, 4)`, `torch.unsqueeze(x, dim=2)` should give `[4, 1, 4]` but
`main` produces `[1, 4, 4]`, because `list.index(4)` returns the first `4`.

### Fix

Insert the size-1 axis directly at `squeeze_dim`, normalizing negative dims to
the range torch accepts (`[-(D+1), D]`). This is a small, self-contained change
in `hls4ml/converters/pytorch/reshape.py`.

### Testing

Added `test_unsqueeze` to `test/pytest/test_pytorch_api.py`, parametrized over
backends and io types and mirroring the existing `test_squeeze`. It exercises
`torch.unsqueeze(x, dim=-1)` and asserts the resulting reshape reports
`target_shape == [4, 1]`. The test fails on `main` (`[1, 4]`) and passes with
this change. The existing `test_squeeze` (which covers `unsqueeze(x, dim=1)`)
continues to pass, and `pre-commit` passes on both edited files.

https://claude.ai/code/session_01SVVaAFkshNYfkJLo6GLukb

### DIFF
--- a/hls4ml/converters/pytorch/reshape.py
+++ b/hls4ml/converters/pytorch/reshape.py
@@ -78,11 +78,15 @@ def parse_unsqueeze_layer(operation, layer_name, input_names, input_shapes, node
         squeeze_dim = node.args[1]
     else:  # Specified as unsqueeze(x, dim=n)
         squeeze_dim = node.kwargs['dim']
-    # torch.unsqueeze inserts a new axis of size 1 at position 'squeeze_dim'.
-    # Normalize negative dims (valid range is [-(D+1), D]) so list.insert places
-    # the axis at the correct location regardless of duplicate dimension sizes.
+    # torch.unsqueeze inserts a new axis of size 1 at position 'squeeze_dim' and accepts
+    # dim in [-(D+1), D]. Reject out-of-range values (list.insert would otherwise silently
+    # clamp them to a wrong shape) and normalize negative dims, so the axis lands at the
+    # correct location regardless of duplicate dimension sizes.
+    ndim = len(output_shape)
+    if not -(ndim + 1) <= squeeze_dim <= ndim:
+        raise Exception(f'Dimension {squeeze_dim} is out of range for unsqueeze of a {ndim}D tensor')
     if squeeze_dim < 0:
-        squeeze_dim += len(output_shape) + 1
+        squeeze_dim += ndim + 1
     output_shape.insert(squeeze_dim, 1)
 
     layer['target_shape'] = output_shape.copy()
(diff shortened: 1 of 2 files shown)


hls4ml/converters/pytorch/reshape.py
```diff
@@ -78,9 +78,16 @@ def parse_unsqueeze_layer(operation, layer_name, input_names, input_shapes, node
         squeeze_dim = node.args[1]
     else:  # Specified as unsqueeze(x, dim=n)
         squeeze_dim = node.kwargs['dim']
-    # insert() will add an element before the index, unsqueeze expects the location
-    index = output_shape.index(output_shape[squeeze_dim])  # + 1
-    output_shape.insert(index, 1)
+    # torch.unsqueeze inserts a new axis of size 1 at position 'squeeze_dim' and accepts
```
JanFSchulte, line 81: Please remove this lengthy agent-style comment, as it clutters the code.

hls4ml/converters/pytorch/reshape.py
```diff
@@ -78,9 +78,16 @@ def parse_unsqueeze_layer(operation, layer_name, input_names, input_shapes, node
...
+    # dim in [-(D+1), D]. Reject out-of-range values (list.insert would otherwise silently
+    # clamp them to a wrong shape) and normalize negative dims, so the axis lands at the
+    # correct location regardless of duplicate dimension sizes.
+    ndim = len(output_shape)
+    if not -(ndim + 1) <= squeeze_dim <= ndim:
```
JanFSchulte, line 86: This condition will never be reached in a real model. Please disregard the suggestion by copilot and revert your last commit to remove this unnecessary code.

JanFSchulte (changes requested):

JanFSchulte:
@areporeporepo Could you have a look at the requested changes? There is now also a conflict that needs to be resolved. We would love to merge this, if possible.

PR #1497, opened 2026-06-29
Re-enable safe checkout action

jmitrevs:
# Description

Checkout v7 made the default behavior safer for pull_request_target events in the CI. We verified that the action is safe, so we are re-enabling the behavior. Please check our assessment.

## Type of change

- [x] Other (Specify) - CI

## Tests

The CI should work after merging.

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have installed and run `pre-commit` on the files I edited or added.
- [ ] I have added tests that prove my fix is effective or that my feature works.

PR #1499, opened 2026-06-30
add tensorflow as the KERAS_BACKEND for qkerasV3 tests

jmitrevs:
# Description

This should fix the qkeras3 pytests issues introduced with qkeras_v3==1.2

## Type of change

QKeras v3 1.2 needs to have the `KERAS_BACKEND` environment variable set, so this sets it in the CI. 

- [x] Other (Specify) - CI

## Tests

This is a CI fix so it will only be visible in the CI. Compare with the qkeras3 pytest in #1497.

## Checklist

### DIFF
--- a/test/pytest/generate_ci_yaml.py
+++ b/test/pytest/generate_ci_yaml.py
@@ -171,7 +171,11 @@ def generate_test_yaml(test_root='.'):
         name = 'qkerasV3'
         test_files = ' '.join([str(path.relative_to(test_root)) for path in batch_paths])
         batch_need_example_model = int(any([qkeras3_need_examples[i] for i in batch_idxs]))
-        diff_yml = yaml.safe_load(template.format(name, '.pytest-qkeras-v3-only', test_files, batch_need_example_model))
+        diff_yml = yaml.safe_load(
+            template_keras3_backend.format(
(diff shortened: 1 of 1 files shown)


calad0i:
Sounds like a bug from upstream in qkeras v3 to me. In keras v3, if `KERAS_BACKEND` is not set, it shall be regarded as `tensorflow`.

jmitrevs:
However, we need a workaround till it is fixed. It seems better to explicitly set it vs to lock the version to an older version of qkeras3.

calad0i:
I would do version lock in this case

jmitrevs:
I don't have a strong preference either way. Let's see what others say. I don't see a particular downside in setting the KERAS_BACKEND explicitly.

makoeppel:
I can set the backend default to tensorflow in qkerasv3.

makoeppel:
I fixed the default backend to tensorflow in qkerasv3.

jmitrevs:
So I think this PR is no longer necessary. The latest version of QKeras3 no longer fails in this case. It is kind of odd, though, that currently pquant tests use the template_keras3_backend in the CI while keras3 tests do not. Feel free to close, though.

PR #1503, opened 2026-07-09
Fix non-quantized QConv2DBatchnorm conversion

### LINKED ISSUE #928: Non-quantized QKeras layers break conversion
## Quick summary
While a bit nonsensical, QKeras layers without quantizers are legal. For example `QDense(n_units)` (without `kernel_quantizer=something`) fails conversion since we expect the quantizers to be present in the converter. In fact, the IR instance of `Dense` doesn't need a quantizer, so it is just the converter that is problematic.

### Steps to Reproduce

Create any model with QKeras layers and don't specify quantizers.

### Expected behavior
The model should convert to IR.

### Actual behavior
The converter fails.

### Possible fix
We need to go through `hls4ml/converters/keras/qkeras.py` and check if supported layers pass without the quantizers specified. The check already exists for bias, we should add it for weights.

### Additional context
I wish this was caught on QKeras side and creation of quantized models without a quantizer is prohibited, but that's out of our control.

hadronic-arvind:
## Summary

- Allow `QConv2DBatchnorm` layers without weight or bias quantizers to convert after batch-normalization folding.
- Preserve folded unquantized weights and bias when their corresponding quantizer is `None`.
- Add matched Keras 2/QKeras and Keras 3/QKeras-v3 regressions for both `Latency` and `Resource` strategies.

Fixes #928.

## Tests

- `pre-commit run --files hls4ml/model/layers.py test/pytest/test_qkeras.py test/pytest/test_qkerasV3.py`
- Keras 2/QKeras focused regression: 2 passed.
- Keras 3/QKeras-v3 focused regression: 2 passed.
- Keras 2/QKeras cross-backend acceptance subset: 14 passed (Vivado, Vitis, Quartus; `io_parallel` and `io_stream`).
- Keras 3/QKeras-v3 cross-backend acceptance subset: 14 passed (Vivado, Vitis, Quartus; `io_parallel` and `io_stream`).
- Conversion-only validation passed for Vivado, Vitis, Quartus, Catapult, and oneAPI, including the Vivado `Resource` path.

### DIFF
--- a/hls4ml/model/layers.py
+++ b/hls4ml/model/layers.py
@@ -689,21 +689,25 @@ class Conv2DBatchnorm(Conv2D):
     def initialize(self):
         super().initialize()
         folded_weights, folded_bias = self._get_folded_weights()
+        weight_q = self.get_attr('weight_quantizer')
         if self.model.config.is_resource_strategy(self) and self.model.config.backend.name in [
             'Vivado',
             'VivadoAccelerator',
             'Catapult',
         ]:
             self.weights['weight'].data_unquantized = np.transpose(folded_weights, axes=[3, 0, 1, 2])
-            self.weights['weight'].data = self.get_attr('weight_quantizer')(self.weights['weight'].data_unquantized)
+            self.weights['weight'].data = (
+                weight_q(self.weights['weight'].data_unquantized)
+                if weight_q is not None
+                else self.weights['weight'].data_unquantized
+            )
 
         else:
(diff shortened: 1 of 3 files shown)


vloncar:
Do you actually have a model with QConv2DBatchnorm that requires this? We don't actively support this layer anymore and it will likely be removed in the future.

Rule of thumb: inquire about an issue before blindly telling your agent to solve it.

PR #1507, opened 2026-07-19
Handle empty optional ONNX inputs in shape parsing

### LINKED ISSUE #1266: hls4ml config_from_onnx_model fails when using a Resize node with no ROI (Brevitas -> QONNX)
## Prerequisites
- [x] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary
hls4ml config_from_onnx_model fails when using a Resize node with no ROI.
 
## Details
hls4ml config_from_onnx_model fails when using a Resize node with no ROI comming from the conversion of a QuantUpsample from Brevitas.

### Steps to Reproduce
2. Checkout the master branch, with commit hash: [77b8331]
3. Run code below

```
import torch.nn as nn
import torch.nn.functional as F
import brevitas.nn as qnn
from brevitas.export import export_qonnx
import torch
import qonnx
from qonnx.core.modelwrapper import ModelWrapper
from qonnx.util.cleanup import cleanup_model
from qonnx.transformation.channels_last import ConvertToChannelsLastAndClean
from qonnx.transformation.qcdq_to_qonnx import QCDQToQuant
from qonnx.transformation.gemm_to_matmul import GemmToMatMul
import onnx

def init_weights(m):
    if isinstance(m, nn.Conv2d) or isinstance(m, nn.ConvTranspose2d):
        nn.init.kaiming_normal_(m.weight, mode="fan_out", nonlinearity="leaky_relu")
        if m.bias is not None:
            nn.init.constant_(m.bias, 0)
    elif isinstance(m, nn.BatchNorm2d):
        nn.init.constant_(m.weight, 1)
        nn.init.constant_(m.bias, 0)

class test_model(nn.Module):
    def __init__(self):
        super(test_model, self).__init__()

        self.quant_inp = qnn.QuantIdentity(bit_width=4, return_quant_tensor=True)
        self.upsample = qnn.QuantUpsample(scale_factor=2)

        for m in self.modules():
            init_weights(m)

    def forward(self, x):
        x1 = self.quant_inp(x)
        x2 = self.upsample(x1)
        

        return x2

model = test_model()
export_qonnx(model, torch.randn(1, 1, 25, 25), export_path='qmodel.onnx')

model = ModelWrapper('qmodel.onnx')
model = cleanup_model(model)
model = model.transform(ConvertToChannelsLastAndClean())
model = model.transform(GemmToMatMul())
model = cleanup_model(model)
onnx.save(model.model, 'transformed_model.onnx')

import hls4ml
from hls4ml.converters import convert_from_onnx_model
from hls4ml.utils.config import config_from_onnx_model

config = hls4ml.utils.config.config_from_onnx_model(model)
```

### Expected behavior
Sucessfull creation of config from model.

### Actual behavior

```
Warning:  it is recommended to pass the backend to "config_from_onnx_model"
Output layers:  ['Resize_0']
Input shape: [1, 25, 25]
Topology:
Layer name: Quant_0, layer type: Quant, current shape: [[1, 1, 25, 25]]
  File "/home/user/project/ConversionQuantUpsample.py", line 54, in <module>
    config = hls4ml.utils.config.config_from_onnx_model(
  File "/home/user/project/.venv/lib/python3.10/site-packages/hls4ml/utils/config.py", line 492, in config_from_onnx_model
    layer_list, _, _ = hls4ml.converters.parse_onnx_model(model)
  File "/home/user/project/.venv/lib/python3.10/site-packages/hls4ml/converters/onnx_to_hls.py", line 244, in parse_onnx_model
    input_shapes = get_input_shape(onnx_model.graph, node)
  File "/home/user/project/.venv/lib/python3.10/site-packages/hls4ml/converters/onnx_to_hls.py", line 76, in get_input_shape
    raise RuntimeError(f'Could not find the shape for input {inp}')
RuntimeError: Could not find the shape for input
```

### Additional context
When printing the `node.input` you get `['Quant_0_out0', '', 'Resize_0_param0']`. It seems it's the `''` that is causing the issue. In addition when checking the model with [Netron](https://netron.app/) in the inputs of the Resize node there is no roi.

niteshg97:
## Summary

Fixes #1266.

In ONNX, optional inputs that are not provided are represented by an empty string (`""`). For example, a `Resize` node can have inputs like : ['Quant_0_out0', '', 'Resize_0_param0']

where the second entry corresponds to the optional ROI input.

At the moment, `get_input_shape()` attempts to resolve the shape for every input, including the empty placeholder. Since an empty string does not correspond to a valid tensor, this results in,

 RuntimeError: Could not find the shape for input

This change simply skips empty input placeholders while keeping the existing behavior unchanged for valid input names. If a non empty input is missing its shape, the same error is still raised.

## Changes

- Ignore empty input placeholders in `get_input_shape()`.
 - Update the docstring to clarify that shapes are returned only for non empty inputs.
- Add a regression test for a `Resize` node with the optional ROI input omitted.

## Testing

```bash
pytest test/pytest/test_onnx_to_hls.py
```

### DIFF
--- a/hls4ml/converters/onnx_to_hls.py
+++ b/hls4ml/converters/onnx_to_hls.py
@@ -49,20 +49,22 @@ def get_global_input_shape(graph, inp):
 
 
 def get_input_shape(graph, node):
-    """Return the input shapes of the node in the model
+    """Return the non empty input shapes of the node in the model
 
     Arguments:
         graph:  the onnx graph
         node:  the onnx node for which the input is desired
 
     Returns:
-        list of lists: The shapes of all the inputs
+        list of lists: The shapes of all the non empty inputs
 
     Raises:
         StopIteration:  If the an input name is not found in the graph
     """
     rv = []
     for inp in node.input:
+        if not inp:
+            continue
         # first try regular variables
         vals = [x for x in graph.value_info if x.name == inp]
         if not vals:
--- a/test/pytest/test_onnx_to_hls.py
+++ b/test/pytest/test_onnx_to_hls.py
@@ -1,0 +1,27 @@
+from onnx import TensorProto, helper
+
+from hls4ml.converters.onnx_to_hls import get_input_shape
+
+
+def test_get_input_shape_skips_empty_optional_inputs():
+    input_tensor = helper.make_tensor_value_info('global_in', TensorProto.FLOAT, [1, 1, 25, 25])
+    scales_tensor = helper.make_tensor_value_info('resize_scales', TensorProto.FLOAT, [4])
+    output_tensor = helper.make_tensor_value_info('global_out', TensorProto.FLOAT, [1, 1, 50, 50])
+
+    resize_node = helper.make_node(
+        'Resize',
+        name='resize',
+        inputs=['global_in', '', 'resize_scales'],
+        outputs=['global_out'],
+        mode='nearest',
+    )
+
+    graph = helper.make_graph(
+        nodes=[resize_node],
+        name='ResizeWithoutRoiGraph',
+        inputs=[input_tensor],
+        outputs=[output_tensor],
+        value_info=[scales_tensor],
+    )
+
+    assert get_input_shape(graph, resize_node) == [[1, 1, 25, 25], [4]]


JanFSchulte:
Thanks for the PR. What is not clear to me if this actually leads to proper behavior of the converted model. Inferring the correct shape is one thing, but has it been verified that the resulting hls4ml model gets wired correctly to skip the missing input, or will the downstream code still expect it?

niteshg97:
> Thanks for the PR. What is not clear to me if this actually leads to proper behavior of the converted model. Inferring the correct shape is one thing, but has it been verified that the resulting hls4ml model gets wired correctly to skip the missing input, or will the downstream code still expect it?

Thanks for bringing this up. I tested the full conversion flow after making this change,the original issue is fixed: config_from_onnx_model() now successfully parses a `Resize` node with an omitted ROI input (["X", "", "scales"]), whereas previously it failed in get_input_shape().

I also tested  convert_from_onnx_model(). The conversion now gets past parsing, but it fails later in the `ResizeRemoveConstants` optimizer. from what I found, it assumes the ROI input always resolves to a `Constant` node and calls `get_attr()` on it unconditionally. When the ROI is omitted (represented by  ""  in ONNX), `node.get_input_node("")` returns `None`, which results in an AttributeError.

so it looks like this PR fixes the original parsing issue reported in #1266, but there's also a separate downstream issue in the optimizer that needs to handle omitted optional ROI inputs.

JanFSchulte:
After discussing this internally, we prefer not to allow parsing of models that we do not support internally. Instead, the models should be parsed correctly by QONNX, and will go in the directions of ONNX models being required to be cleaned by QONNX to ensure compatibility. For this specific issue, a PR fixing it in QONNX already exist. So thank you very your contribution, but we prefer to close here at this point.

niteshg97:
Thanks for the clarification, that makes sense.

I understand the reasoning for keeping this handling in QONNX rather than in hls4ml. Thanks for taking the time to review the PR and explain the intended conversion flow.

PR #1518, opened 2026-08-05
migrate da strategy provider to alkaid

calad0i:
# Description

Replacing da strategy provider `da4ml` to `alkaid`. No user facing breaking changes except for installing alkaid instead of da4ml.

- LUT packaging for CMVM ops with 1-bit inputs
  - Ternary adder fusion reducing up to 33% for add/sub heavy graphs in RTL doesn't translate to vitis HLS for some reason. Disabled for now. 
- General support of unknown layers fallback to symbolic traced implementation
  - All layers in HGQ2
  - Vivado/Vitis only for now, extensible to oneapai

## Type of change

- [x] Breaking change (da strategy now needs alkaid instead of da4ml)

## Tests

Added da fallback implementation of LUT layer and QGRU layer.

**Test Configuration**:

## Checklist

- [x] all


JanFSchulte:
The `QGRU` still fails because there's no linear activation in hls4ml/templates/vivado/nnet_utils/nnet_recr_activations.h. Adding it fixes the compiliation, but I see issues with the `model.predict()` call like this ` tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: Exception encountered when calling QGRUCell.call().` coming from HGQ2 itself. Not sure if I'm running the tests in an incorrect environment offline or if there is really something wrong with QGRU in HGQ2, can you please verify @calad0i ?

calad0i:
Weird. Locally I didn't see the same issue. Could be keras version interacting? Checking

hls4ml/backends/vivado/passes/distributed_arithmetic.py
```diff
@@ -125,23 +126,22 @@ def transform(self, model: 'ModelGraph', node: Layer):
 
         k, i, f = get_kernel_inp_kif(node)
         hard_dc = int(os.environ.get('DA_HARD_DC', 2))
-        options = {'hard_dc': hard_dc, 'search_all_decompose_dc': True}
-        inp = FixedVariableArray.from_kif(k, i, f, HWConfig(1, -1, -1), solver_options=options)
+        options: solver_options_t = {'hard_dc': hard_dc, 'search_all_decompose_dc': True}
+        inp = FVArray.from_kif(k, i, f, solver_options=options)
```
JanFSchulte, line 130: Here you are dropping the `HWConfig` argument. I see that the default in alkaid is ` HWConfig(1, 1, -1)`, so different from what was passed here before. Just checking if that is an intended change.

hls4ml/backends/vivado/passes/distributed_arithmetic.py
```diff
@@ -444,16 +447,44 @@ def transform(self, model: 'ModelGraph', node: DACombinational):
...
+
+        fn_name_internal = f'{fn_name}_internal' if len(inp_names) > 1 else fn_name
+        _name = fn_name_internal if len(inp_names) > 1 else fn_name
+        comb_logic, _ = hls_logic_and_bridge_gen(
+            comb, _name, flavor=flavor, pragmas=['#pragma HLS INLINE'], print_latency=True
+        )
 
-        fn_cpp = f'{namespace}::{fn_name}<{inp_t}, {out_t}>({inp_name}, {out_name});'
+        # When there's multiple inputs, make a wrapper doing the concatenation and rename the original fn.
+        if len(inp_names) > 1:
```
JanFSchulte, line 467: I think it would be good to add a test for the multi-input case.

hls4ml/backends/vivado/passes/distributed_arithmetic.py
calad0i: This is intended; the parameter is used for alkaid internal surrogate, and the config changed

hls4ml/backends/vivado/passes/distributed_arithmetic.py
JanFSchulte: @calad0i can you comment why this case is not really reachable at the moment for the record? I'm fine with merging without the test, but would like to have it documented.

calad0i: It can be reachable by any new layer implementation, like any transformer blocks. But those are already registered in hls4ml,and you need to patch to disable the current path to enable fallback. Or, some new layer is to be proposed and implemented.

JanFSchulte (approved):

PR #1527, opened 2026-08-16
Agent docs

vloncar:
✨  oooh a shiny new template ✨ 

## Description

`AGENTS.md` (added in #1514) tells an AI agent how to behave in this repository. It does not tell it how hls4ml works. An agent therefore learns the codebase by exploring it, and gets it wrong in the same few ways every time: assuming one backend's conventions hold for all of them, generalizing from `Dense` to layer families that share none of its structure, and quoting C synthesis estimates as if they were measured results.

This PR adds that knowledge as plain Markdown under `.agents/`, with `AGENTS.md` routing to it. Twelve documents: how a model is converted and run, a map of the pipeline, the frontends, the optimizer passes, the C++ kernels, precision and debugging, evaluating an implementation, standing up a backend, toolchain access, contributing a change, reporting an issue, and a template for site-specific setup.

Three things shaped how they are written:

- **They say what is a contract and what is one backend's convention.** The `strategy` field, the `kernel` typedef and the `nnet_utils` layout exist in the Vivado family; Quartus and oneAPI have neither, oneAPI uses pipes and its own template classes, and the symbolic backend has no kernels at all. Documentation that presents the Vivado shape as "how hls4ml works" is what sends a contributor down the wrong path.
- **They contain no site-specific content.** No paths, tool versions or environment assumptions. `local-setup.template.md` is a skeleton each group fills in for itself. If people are interested, I can provide my cvmfs-based setup that would work well for CERN people (onsite or users through lxplus/lxtunnel)
- **They are assistant-neutral.** Contributors use Claude Code, Codex, Copilot, Cursor and others. The documents carry front matter that is a superset of what those tools want, and `.agents/agent_adapters.py` renders them into each layout on demand. Generated output is git-ignored, so `.agents/` stays the only copy anyone edits. Pointing an assistant straight at `.agents/` also works.

Documentation about mechanisms rots faster than the code it describes, so `test/pytest/test_agent_docs.py` checks that every repository path these documents mention still exists, that the front matter parses, and that the links between them resolve. A renamed module fails the suite.

Relates to #1514

## Type of change

- [x] Documentation
- [x] Build, CI or tooling

## Affected areas

**Backends:**
- [x] Backend-independent (core IR, optimizer, `hls4ml.model`)

**Frontends:**
- [x] Not related to a frontend

**Components:**
- [x] Packaging, build or CI

## Configurations affected and exercised

Not applicable — this PR adds documentation, a helper script and a test. No conversion path is touched.

## Impact on generated HLS

[...]

PR #1531, opened 2026-08-24
Use HGQ's trained softmax lookup tables in the generated HLS

### LINKED ISSUE #1523: hls4ml HGQ2 issues: softmax LUT mismatch + negative bitwidths
# hls4ml HGQ2 issues: softmax LUT mismatch + negative bitwidths

## Summary

For HGQ2-quantized models (in this case using `QLinformerAttention` layers), **da4ml `comb_trace` is bit-accurate to Keras**, while **hls4ml diverges** due to two issues:

1. **Softmax LUT values** — hls4ml uses `std::exp()`/`1/x` at runtime instead of HGQ's trained tables
2. **Negative bitwidths** — `extract_fixed_quantizer_config` produces garbage types for learned zero-width channels

### Root cause

hls4ml's C++ softmax (`nnet_utils/nnet_activation.h`) populates LUT tables at first call:

```cpp
void init_exp_table(...) {
    for (i = 0; i < table_size; i++) {
        table_out[i] = std::exp(x);    // generic exp()
    }
}

void init_invert_table(...) {
    for (i = 0; i < table_size; i++) {
        table_out[i] = 1 / x;          // generic 1/x
    }
}
```

HGQ's `QSoftmax` uses `QUnaryFunctionLUT` layers (`exp_table`, `inv_table`) that store **trained LUT values** computed during HGQ training with quantization applied. hls4ml's `_bit_exact=True` correctly uses the table **sizes** and **types** from HGQ quantizer specs, but the **values** are generic `exp()`/`1/x`, not HGQ's trained weights.

This is explicitly documented in hls4ml source:
```python
# bit_exact.py:780-782
if not node.attributes.get('_bit_exact', False):
    # Softmax is not bit-exact by default
    warn(f'Softmax layer {node.name} is converted from a frontend '
         f'not supporting bit-exact softmax.')
```

###  first attention block (`q_softmax`, stable implementation)

| | HGQ (trained) | hls4ml  |
|---|---|---|
| Exp table (8 entries) | `[0.75, 0.75, 0.75, 0.75, 0.75, 0.25, 0.0, 0.0]` | different `exp(x * scale)` values |
| Inv table (512 entries) | mostly 0.0, upper block 0.5 | different `1/x` values |

## Per-output-dimension error (Keras vs hls4ml)

| Dim | Keras mean | HLS mean | MAE | corr |
|-----|-----------|----------|-----|------|
| 0 | 0.036 | 0.025 | 0.026 | 0.55 |
| 1 | 0.051 | 0.050 | 0.028 | 0.86 |
| 2 | -0.009 | -0.017 | 0.023 | 0.89 |
| 3 | -0.150 | -0.138 | 0.015 | 0.52 |
| 4 | -0.138 | -0.128 | 0.034 | 0.55 |
| 5 | 0.046 | 0.053 | 0.020 | 0.85 |
| 6 | 0.011 | 0.011 | 0.019 | 0.82 |
| 7 | 0.025 | 0.026 | 0.014 | 0.63 |
| 8 | -0.059 | -0.032 | 0.031 | 0.87 |
| 9 | 0.068 | 0.031 | 0.050 | 0.80 |
| 10 | -0.008 | -0.010 | 0.018 | 0.59 |
| 11 | 0.065 | 0.053 | 0.028 | 0.88 |
| 12 | 0.048 | 0.051 | 0.023 | 0.82 |
| 13 | -0.040 | -0.044 | 0.010 | 0.81 |
| 14 | 0.027 | 0.008 | 0.028 | 0.85 |
| 15 | -0.007 | 0.024 | 0.040 | 0.66 |

Manually widening the softmax output type from `ap_ufixed<1,0>` to `ap_ufixed<4,0>` (and downstream types to match attention1) produced **identical** error — the discrepancy is inside the LUT arithmetic before any truncation.

## Comparison plots

da4ml comb_trace matches Keras exactly (bit-accurate).

hls4ml diverges — scatter points deviate from the diagonal.

Direct da4ml vs hls4ml comparison shows the same discrepancy.

### Root cause

hls4ml's `extract_fixed_quantizer_config` (`hgq2/_base.py:19`) computes total bitwidth as `B = k + i + f` directly from HGQ quantizer specs:

```python
# hls4ml/converters/keras_v3/hgq2/_base.py:28-29
k, i, f = internal_q.kif
k, B, I = k, k + i + f, k + i   # B = total bits, I = integer bits
```

HGQ2 uses **heterogeneous quantizers** — per-element `(k, i, f)` where some channels learn **negative total bitwidths** (`k + i + f < 0`) as the training zeroes them out. When `B < 0`, the resulting `FixedPrecisionType(-5, 0)` produces garbage C++ types:

```
ap_fixed<-5,0>          ← negative width, template error
ap_ufixed<2,32>         ← 2 bits with 32 integer bits
```

This affected multiple layers including attention2's query/key input quantizers and the QK einsum output.

### Evidence (before fix)

```
attention2_query_iq_t:      ap_ufixed<2,32>     ← garbage
attention2_attention2_QK_t: ap_ufixed<2,32>     ← garbage
attention2_attention2_aV_t: ap_ufixed<1,0>      ← too narrow
```

### Fix applied

Clamp `B >= 1` and `I >= 0` in the extract function:

```python
B = np.clip(B, 1, None)
I = np.clip(I, 0, None)
```

Same fix for `fixed_quantizer_to_hls4ml_t` in the softmax handler. After fix:

```
attention2_query_iq_t:      ap_fixed<2,1>        ← correct
attention2_attention2_QK_t: ap_fixed<4,2>        ← correct
```

JanFSchulte:
## Description

`hgq.layers.QSoftmax` trains two lookup tables, `exp_table` and `inv_table`. `QSoftmaxHandler`
propagated only their **sizes and types**; the generated C++ rebuilt the **contents** at runtime
from `std::exp` and `1/x` (`nnet::init_exp_table` / `nnet::init_invert_table`). Those
reconstructions are not guaranteed to reproduce the values HGQ trained, and the table was
addressed with a bit-slice whose width came from a different quantity than the address word.

The tables cannot be built at conversion time. For `implementation='latency'` their domain is the
softmax **input** precision, which only the `bit_exact` pass decides. `QSoftmaxHandler` therefore
stashes two domain-parameterised builders on the layer, and a new Vivado pass,
`materialize_softmax_tables`, calls them once the address type is final and turns the result into
weight arrays. It runs in the `vivado:optimize` flow next to `fix_softmax_table_size`, after
`bit_exact` and before `transform_types`.

The pass also forces `exp_table_size == 2**width` of the address type. `softmax_idx_from_real_val`
slices the top `ceillog2(table_size)` bits of the address word, so when the two disagree the table
is read at each bin's lower edge, or the slice runs off a narrower word. This is an independent
defect from the table contents and is a plausible larger contributor to the error reported in the
issue — see "Numerical behaviour" below.

The kernels gain `softmax_lut` / `softmax_multidim_lut` entry points taking the two tables as
arguments. `softmax_latency` / `softmax_stable` were split into an `_impl` body plus the existing
runtime-initialising entry point, which is unchanged, so plain Keras and QKeras softmax generate
exactly the code they did before. Separate names rather than overloads of `softmax`, so that
`softmax_multidim`'s `#pragma HLS allocation instances = softmax<CONFIG_T>` still names a single
function.

The same issue reports invalid types such as `ap_ufixed<2,32>`. That spelling is `bit_exact`'s
deliberate sentinel for `B < 1`, not itself a bug, but it is reached through a real one: HGQ's
heterogeneous quantizers can train a channel to a negative total bitwidth. `extract_fixed_quantizer_config`
now clamps `B` to **0**, which is the constant-zero encoding `generate_mask_fn` already understands,
rather than to 1, which would bring a pruned channel back as a live 1-bit one. `generate_mask_fn`'s
`if b == 0` is widened to `b <= 0` so the other four frontends that build `mask_kbi` are covered too.

Fixes #1523

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Breaking change (existing configs, APIs or generated code behave differently)
- [ ] New frontend / backend
- [ ] Documentation
- [ ] Build, CI or tooling

## Affected areas

**Backends:**
- [x] Vivado, Vitis (VivadoAccelerator inherits `vivado:optimize` and so also gets the pass; not exercised)
- [ ] Backend-independent (core IR, optimizer, `hls4ml.model`)

Nothing under `hls4ml/model/` changes. Quartus, oneAPI, Catapult and Libero are untouched: the pass
is registered under the Vivado backend only, and their softmax kernels keep building tables at
runtime exactly as before.

**Frontends:**
- [x] Keras v3 (HGQ2 handlers only)

**Components:**
- [x] IR (layers + model graph) / optimizer passes
- [x] C++ templates / HLS sources under `hls4ml/templates/`
- [x] Packaging, build or CI (one entry in `generate_ci_yaml.py`)

The axis this change is keyed on is the softmax `implementation`, not Strategy.

| Backend | `io_type` | `implementation` | Verified |
|---|---|---|---|
| Vivado | io_parallel | stable, latency | pytest + csim, 1D and multidim inputs |
| Vivado | io_stream | stable, latency | pytest + csim, 1D and multidim inputs |
| Vitis | io_parallel | stable, latency | pytest + csim, 1D and multidim inputs |
| Vitis | io_stream | stable, latency | pytest + csim, 1D and multidim inputs |
| Vivado, Vitis | both | argmax, legacy | pass returns early; covered by the existing `test_softmax.py` |
| VivadoAccelerator | — | — | not run |

Deliberately left out: no synthesis run of any configuration (see below), and no HGQ2 softmax on
Quartus / oneAPI / Catapult / Libero, which the HGQ2 frontend is not tested on today
(`test_hgq_layers.py`, `test_hgq_players.py` and `test_hgq2_mha.py` are all Vivado/Vitis only).

**New configuration axis introduced by this PR:** none.

## Impact on generated HLS

The change is confined to HGQ2 `QSoftmax`. For every other softmax the emitted call, the config
struct and the kernel body are byte-identical to `main`; `test_hgq2_softmax.py` carries an explicit
negative control asserting a plain `keras.layers.Softmax` still gets the two-argument call.

For HGQ2 `QSoftmax` the call gains two arguments and the project gains two weight headers
(`exp_table<n>.h`, `inv_table<n>.h`), while the runtime `init_*_table` loops disappear from that
layer.

**Numerical behaviour:** intentionally changed. After this PR, `hls_model.predict` on the test
models is bit-exact with Keras — `np.testing.assert_array_equal`, not `assert_allclose` — for all
16 covered combinations.

One caveat worth recording, since it bears on how the issue is attributed. Before implementing this
I compared HGQ's trained tables against what `init_exp_table` / `init_invert_table` reconstruct for
default `QSoftmax` configurations, and they matched **bit for bit** (the C++ assigns through
`exp_table_t`, which already carries HGQ's `RND_CONV`/`SAT`). On those models the reported error
therefore came from the address-width mismatch, not from the table values. Both are fixed here, and
carrying the trained tables makes the result independent of `exp_scale` and of the generic
reconstruction; but I have not isolated the two effects on the reporter's model.

| Model / test | Backend & version | Part | Latency (cycles) | II | LUT | FF | DSP | BRAM |
|---|---|---|---|---|---|---|---|---|
| before | — | — | not run | | | | | |
| after  | — | — | not run | | | | | |

**No synthesis was run.** C simulation proves the arithmetic; it cannot see two things that need a
real build: whether `#pragma HLS allocation instances = softmax_lut<CONFIG_T>` resolves as intended
in `softmax_multidim_lut`, and the QoR effect of `function_instantiate` on the table arguments.
Both are flagged here rather than assumed.

## Tests

New file `test/pytest/test_hgq2_softmax.py`, 19 tests:

- `test_hgq2_softmax_uses_trained_tables`, parametrised over backend × `io_type` × `stable` ×
  input shape (16 cases). Asserts the weights exist and are sized `2**addr_width`; that they carry
  the `NamedType` `bit_exact` derived rather than one re-inferred from the data (this catches
  `AttributeDict.__setitem__` silently overwriting `<name>_t` when a `WeightVariable` is stored);
  that the contents equal an independent recomputation straight from `hgq` and **differ** from a
  naive `exp(x * exp_scale)` reconstruction, so the test cannot pass vacuously; that the generated
  call names both arrays and the headers are written and included; and finally
  `assert_array_equal(r_hls, r_keras)`.
- `test_plain_keras_softmax_keeps_runtime_tables` — negative control, two-argument call preserved.
- `test_dead_channels_stay_zero` — a heterogeneous quantizer channel forced to a negative total
  width stays `b == 0` and renders as `out[3] = 0;`.

`test_hgq2_softmax` is added to `KERAS3_LIST` in `generate_ci_yaml.py`, so CI runs it under
`.pytest-keras3-only` alongside `test_hgq2_mha`. Without that it would be batched into a default
job, whose environment has Keras 2 and no `hgq`.

Commands run, all from `test/pytest` on this branch:

```
pytest test_hgq2_softmax.py -q -p no:randomly            # 19 passed
pytest test_softmax.py -q -p no:randomly -k "Vivado or Vitis"   # 116 passed
pytest test_sparsepixels.py test_bit_exact_zeropadding.py -q -p no:randomly
                                                          # 17 passed, 2 failed
pre-commit run --files <every file in this diff>          # all hooks passed
pre-commit run --hook-stage manual --all-files check-manifest
                                                          # only uninitialised submodules reported
```

The two failures are `test_bit_exact_zeropadding.py::test_bit_exact_zeropadding{1,2}d[io_parallel-oneAPI]`.
They fail identically on an unmodified `origin/main` worktree here, so they are pre-existing and
unrelated — this PR does not touch oneAPI.

`test_hgq_layers.py` and `test_hgq_players.py`, which cover the `QUnaryLUTHandler` refactor, could
**not** be run: HGQ v1 is not installed in my environment (`ModuleNotFoundError: No module named
'HGQ'`). Instead I checked the refactor directly, running the extracted helpers and the verbatim
pre-refactor algorithm side by side on `QUnaryFunctionLUT` layers with relu, sigmoid, tanh and exp
activations; the tables are byte-identical. That refactor also fixes a latent bug it inherited: the
old code called `layer.oq(table[None, ...])`, which fails on any layer built for rank > 2 — which
is exactly what `QSoftmax`'s sublayers are for a multidimensional input.

**Test configuration:** Linux, Python 3.10, Keras 3 on the TensorFlow backend, `hgq` 0.1.8. No HLS
toolchain — `compile()` builds against the bundled `ap_types/` with g++.

## AI assistance disclosure

Tool(s) and model(s): Claude Code (Claude Opus 5)

Where it was used: all files in this diff — the optimizer pass, the converter changes, the kernel
split, and the tests — from prompts describing the issue.

- [x] **I am the author of this contribution and take full responsibility for it.** ...
- [x] I verified the generated code against real hls4ml and HLS semantics ...
- [x] All numbers, logs and test results quoted in this PR come from runs I actually performed ...
- [x] I have the right to submit this work under the project's licence ...

## Checklist

**Required:**
- [x] I have read the [contributing guidelines](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I added tests under `test/pytest` covering this change.
- [x] I self-reviewed the full diff ...

**If applicable:**
- [ ] Documentation under `docs/` updated. — not applicable; this is a bug fix with no new
      user-facing option, and no `docs/` page describes the softmax table construction.
- [ ] No new build or synthesis warnings. — cannot confirm, no synthesis was run.
- [ ] Public API / config schema changes are documented. — none.

## Release note

```release-note
HGQ2 QSoftmax layers now emit the lookup tables trained by HGQ instead of rebuilding them at
runtime in the generated C++, making Vivado and Vitis output bit-exact with Keras.
```

---

### DIFF
--- a/hls4ml/backends/fpga/passes/hgq_proxy_model.py
+++ b/hls4ml/backends/fpga/passes/hgq_proxy_model.py
@@ -34,7 +34,8 @@ def generate_mask_fn(
     masks = []
     to_fixed = to_acfixed if backend.lower() in ['oneapi', 'quartus'] else to_apfixed
     for idx, (k, b, i) in enumerate(zip(Ks, Bs, Is)):
-        if b == 0:
+        if b <= 0:
+            # b <= 0 means nothing is representable: the channel is a constant zero
             fn = f'out[{idx}] = 0;'
         else:
             fn = f'out[{idx}] = {to_fixed(k, b, i, RND, SAT)}(inp[{idx}]);'
--- a/hls4ml/backends/vivado/passes/core_templates.py
+++ b/hls4ml/backends/vivado/passes/core_templates.py
@@ -250,6 +250,9 @@ activ_function_template = 'nnet::{activation}<{input_t}, {output_t}, {config}>({
 param_activ_function_template = (
     'nnet::{activation}<{input_t}, {param_t.name}, {output_t}, {config}>({input}, {param}, {output});'
 )
+softmax_lut_function_template = (
+    'nnet::{activation}<{input_t}, {output_t}, {config}>({input}, {output}, {exp_table}, {inv_table});'
+)
 
 activ_include_list = ['nnet_utils/nnet_activation.h', 'nnet_utils/nnet_activation_stream.h']
 
@@ -348,6 +351,14 @@ class SoftmaxFunctionTemplate(FunctionCallTemplate):
         use_multidim = use_multidim and node.model.config.get_config_value('IOType') == 'io_parallel'
         params['activation'] = 'softmax' if not use_multidim else 'softmax_multidim'
         params['config'] = '{}_config{}'.format(node.get_attr('activation'), node.index)
+
+        # HGQ2 softmax carries its trained tables as weights; every other softmax lets the
+        # HLS code build them at runtime.
+        if 'exp_table' in node.weights and 'inv_table' in node.weights:
+            params['activation'] += '_lut'
+            params['exp_table'] = node.get_weights('exp_table').name
+            params['inv_table'] = node.get_weights('inv_table').name
+            return softmax_lut_function_template.format(**params)
 
         return self.template.format(**params)
 
--- a/hls4ml/backends/vivado/passes/hgq_softmax_tables.py
+++ b/hls4ml/backends/vivado/passes/hgq_softmax_tables.py
@@ -1,0 +1,68 @@
+from warnings import warn
+
+from hls4ml.model.layers import Layer, Softmax
+from hls4ml.model.optimizer import OptimizerPass
+from hls4ml.model.types import FixedPrecisionType
+
+
+class MaterializeSoftmaxTables(OptimizerPass):
+    """Turn HGQ's trained softmax exp/reciprocal lookup tables into weight arrays.
+
+    Without this the generated C++ rebuilds both tables at runtime from std::exp and 1/x
+    (nnet::init_exp_table / nnet::init_invert_table), which does not reproduce HGQ's own
+    quantized values. QSoftmaxHandler stashes the builders as _exp_table_fn / _inv_table_fn;
+    both take the (k, i, f) of the type the C++ addresses the table with, which is known only
+    here: for the latency implementation it is the softmax input precision, decided by
+    bit_exact. This pass therefore runs after bit_exact and before transform_types.
+    """
+
+    def match(self, node: Layer):
+        # Keyed on the stash rather than on the layer type: only the HGQ frontend leaves it,
+        # and it must always be removed again, being a callable no saved model can hold.
+        return isinstance(node, Softmax) and '_exp_table_fn' in node.attributes
+
+    def transform(self, model, node: Layer):
+        exp_table_fn = node.attributes.pop('_exp_table_fn')
+        inv_table_fn = node.attributes.pop('_inv_table_fn')
+
+        impl = node.get_attr('implementation')
+        if impl not in ('latency', 'stable'):
+            return False  # argmax and legacy use no lookup table
+
+        # The exp table is indexed by the normalized input (x_max - x) for the stable
+        # implementation and by the layer input itself for the latency one. Reading the
+        # element precision off the variable is only unambiguous because this runs before
+        # transform_types wraps io_stream variables.
+        if impl == 'stable':
+            exp_inp_t: FixedPrecisionType = node.attributes['inp_norm_t'].precision
+        else:
+            exp_inp_t = node.get_input_variable().type.precision
+        inv_inp_t: FixedPrecisionType = node.attributes['inv_inp_t'].precision
+
+        for name, addr_t, fn in (('exp_table', exp_inp_t, exp_table_fn), ('inv_table', inv_inp_t, inv_table_fn)):
+            # softmax_idx_from_real_val slices the top ceillog2(table_size) bits of the address
+            # word. Sizing the table as 2**width makes that the whole word, so no low bits are
+            # dropped and the slice cannot run off the end of a narrower word.
+            size = 1 << int(addr_t.width)
+            if int(node.get_attr(f'{name}_size') or 0) != size:
+                warn(
+                    f'{node.name}: {name}_size {node.get_attr(f"{name}_size")} does not match the {addr_t.width}-bit '
+                    f'address type {addr_t}; overriding to {size} to keep table indexing bit-exact.',
+                    stacklevel=1,
+                )
+            node.set_attr(f'{name}_size', size)
+            k = int(bool(addr_t.signed))
+            data = fn((k, addr_t.integer - k, int(addr_t.width) - addr_t.integer))
+
+            # Capture the type bit_exact derived first: storing a WeightVariable under `name`
+            # makes AttributeDict overwrite `{name}_t` with the variable's own type.
+            named_t = node.attributes[f'{name}_t']
+            node.set_attr(f'{name}_data', data)
+            node.add_weights_variable(
+                name=name, var_name=name + '{index}', precision=named_t.precision, type_name=named_t.name
+            )
+            node.get_weights(name).type = named_t
+            node.attributes[f'{name}_t'] = named_t
+            model.config.layer_name_precision[f'{node.name}_{name}'] = str(named_t.precision)
+
+        return False
--- a/hls4ml/backends/vivado/vivado_backend.py
+++ b/hls4ml/backends/vivado/vivado_backend.py
@@ -171,6 +171,7 @@ class VivadoBackend(FPGABackend):
             'vivado:inplace_stream_flatten',
             'vivado:skip_softmax',
             'vivado:fix_softmax_table_size',
+            'vivado:materialize_softmax_tables',
             'infer_precision_types',
             'vivado:distributed_arithmetic_codegen',
             'vivado:distributed_arithmetic_einsum_codegen',
--- a/hls4ml/converters/keras_v3/hgq2/_base.py
+++ b/hls4ml/converters/keras_v3/hgq2/_base.py
@@ -1,6 +1,7 @@
 from collections.abc import Sequence
 from math import prod
 from typing import TYPE_CHECKING, Any
+from warnings import warn
 
 import numpy as np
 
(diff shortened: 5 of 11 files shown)


JanFSchulte:
I made this mostly as an example to test @vloncar's agent instructions from #1527, including the new PR template. At first glance, I think Claude did a decent job following the instructions. Still a bit verbose in the comments, but harmless compared to other things I have seen. It did flag a few things it would have fixed on the fly, and acknowledged that it was instructed to stay within the scope of the requested fix. So that part of the instructions seem to work well. 

As for the fix itself, it seems a bit clunky to me, but does work, so we can discuss if we want to actually merge it or come up with something better.

PR #1534, opened 2026-08-26
Keras frontends: stop silently misparsing unsupported layer configurations

vloncar:
# Keras frontends: stop silently misparsing unsupported layer configurations

## Description

This PR is the Keras half of a systematic audit of the frontends for silent misparses: legal, realistic layer configurations that convert without complaint but compute something different from the model the user trained. The rule applied throughout: **never convert a model that computes different math than the one that was trained.** Each case is resolved according to where it sits:

**Ingested faithfully into the IR** (no kernel computes these yet — backend-side rejection is the immediate follow-up PR, together with a general capability check):

- *Conv `dilation_rate`*: was dropped on the floor — the emitted model computed an undilated convolution with the undilated output shape. The parsers now store the `dilation` attribute (a field the conv config structs already have, and which the PyTorch parser already sets) and compute padding/output shape from the dilated kernel extent. Asymmetric 2D dilation is rejected (the IR carries a single dilation value). Supersedes the parser-level reject of #1521.
- *`ReLU(max_value=...)`* (ReLU6): silently dropped in v2, rejected in v3. Now ingested as a new `ClippedReLU` parametrized activation. The Vivado-family templates already ship unmapped `relu_max`/`relu6` kernels, so actual support is close. `max_value` combined with `negative_slope`/`threshold` stays rejected.
- *`GRU(reset_after=False)`*: was parsed into an `apply_reset_gate` attribute that no backend reads, with a bias split that assumed the `reset_after=True` weight layout (garbage bias values). Now parsed faithfully (`apply_reset_gate='before'`, flat bias, zero recurrent bias), standalone and inside `Bidirectional`, on both frontends.
- *`Bidirectional(merge_mode != 'concat')`*: the attribute was stored but consumed by nothing, so `sum`/`mul`/`ave` silently produced concat behavior with the wrong `n_out`. Now ingested with the correct `n_out`. `merge_mode=None` (two separate output tensors) is rejected — the Bidirectional layer cannot represent it.

**Rejected with a clear error** (previously silently wrong, out of scope for hls4ml):

- *Grouped convolution (Keras v2)*: `groups` was ignored entirely. Depthwise-shaped groups (`groups == in_channels == filters`) are now routed to DepthwiseConv, mirroring the v3 (#1486) and PyTorch (#1524) behavior; other groupings are rejected, consistent with those parsers. A grouped `QConv` carries its kernel quantizer to the depthwise weight.
- *`go_backwards=True`* on non-bidirectional RNNs (was silently parsed as a forward RNN); `stateful=True` prints a warning.
- *`Bidirectional(SimpleRNN)`*: the backend implements only LSTM/GRU cells (and the v2 weight lookup was broken for SimpleRNN).
- *`BatchNormalization`* with the axis not pointing at the last dimension (`n_filt` was taken from the last dimension regardless, normalizing along the wrong axis).
- *`Dot(normalize=True)`* (cosine similarity — was silently computed as a plain dot product), *`Embedding(mask_zero=True)`* (masking silently dropped), and v3 no-op preprocessing layers that change shape (`RandomCrop` center-crops at inference time).

**Fixed** (previously crashed or computed wrong values):

- RNN `use_bias=False` crashed the v2 parser (the zeros substitute read its shape from the missing bias).
- v3 `Bidirectional(GRU)` bias handling: `isinstance(rnn_layer.cell, keras.layers.GRU)` is never true (the cell is a `GRUCell`), so biases were stored unsplit and the recurrent bias was never set.
- Plain `ReLU()` through the v3 handler was silently classified as `ThresholdedReLU(0.0)` (branch-order bug; numerically identical, wrong layer class).
- `ZeroPadding2D` with a `(sym_h, sym_w)` padding config: a typo wrote `pad_bottom` twice and never set `pad_right`.
- Global pooling `keepdims=True`: the parser read the flag but the layer classes always emitted `[n_filt]`, so the graph's shape bookkeeping silently diverged from Keras (numerics unaffected only because the kept dimensions are 1s). The GlobalPooling1D/2D layer classes now honor a `keepdims` attribute.
- `LayerNormalization(center=False)` / `(scale=False)` fed `None` weight data into the graph; zeros/ones are now substituted. Negative axis values (how Keras 3 serializes the default) are accepted.

All changes apply to both the Keras v2 handlers (legacy models and QKeras, which delegates to them) and the Keras v3 handlers. A follow-up PR applies the same audit to the PyTorch frontend. Until the backend-side rejection lands, the four ingested configurations parse correctly but no kernel computes them — that follow-up is the necessary counterpart of this PR.

Relates to #1521, #1486, #1524.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Breaking change (existing configs, APIs or generated code behave differently)
- [ ] New frontend / backend
- [ ] Documentation
- [x] Build, CI or tooling (test output directory naming)

## Affected areas

**Backends:**
- [ ] None directly
- [x] Backend-independent (core IR, optimizer, `hls4ml.model`)

**Frontends:**
- [x] Keras v3
- [x] Keras v2 (including QKeras delegation)

**Components:**
- [x] IR (layers + model graph) / optimizer passes — `GlobalPooling1D/2D.keepdims`, `ClippedReLU` registration in `layer_map`
- [ ] C++ templates / HLS sources under `hls4ml/templates/`
- [ ] Packaging, build or CI

| Backend | `io_type` | Strategy | Verified |
|---|---|---|---|
| all (parsing is backend-independent) | all | all | pytest (C simulation) with Vivado backend |

**New configuration axis introduced by this PR**: none user-facing. New IR attributes: `dilation` on conv layers (Keras parsers now set the field the config structs already had), `keepdims` on global pooling, `apply_reset_gate='before'` now reachable, `ClippedReLU` activation. No backend supports the first, third and fourth yet; the follow-up PR makes every backend reject them explicitly.

## Impact on generated HLS

**Numerical behaviour:** Only models using the affected layer configurations are impacted, in three ways: (1) configurations listed under "rejected" no longer produce (wrong) generated code — conversion raises; (2) `keepdims=True` global pooling now emits the kept-dims output shape (values unchanged — verified bit-identical in the existing `test_globalpooling` tests, which pass unmodified except for added shape assertions); (3) previously-crashing configurations (`use_bias=False` RNNs, `center/scale=False` LayerNorm) and previously-wrong ones (v3 Bidirectional GRU biases, `ZeroPadding2D` tuple padding, ReLU `negative_slope`/`threshold`) now generate correct code, verified against Keras predictions by C simulation. Models not using any affected configuration generate identical code. No synthesis was run — this is a frontend/parse-level change; the resource/latency table is deliberately left out.

## Tests

- `test_keras_converter.py`: extended from the single #1521 test into the direct (handler-level) test suite for the v2 parsers — ingestion attributes, weight handling and every reject, exercised deterministically under any installed Keras version. The #1521 dilation test is updated from asserting the reject to asserting the ingestion.
- `test_keras_v3_api.py`: the same parse/reject matrix through real Keras 3 models (runs in the Keras 3 CI environment), plus the v3-only fixes: Bidirectional GRU bias (numeric), plain-ReLU classification, RandomCrop no-op shape guard.
- Per-layer files: `test_rnn.py` (`use_bias=False` numeric for all three cells, Bidirectional GRU numeric), `test_keras_api.py` (ReLU `negative_slope`/`threshold` numeric, grouped-depthwise routing numeric), `test_layernorm.py` (`center=False`/`scale=False` numeric), `test_globalpooling.py` (output shape assertions added to the existing keepdims-parametrized tests), `test_batchnorm.py`, `test_embed.py`, `test_merge.py` (rejects).
- `conftest.py` / `synthesis_helpers.py`: the `test_case_id` fixture now prefixes its value with `hls4mlprj_`, so every test that builds its output directory from it (the vast majority) produces project directories covered by `.gitignore`; baseline synthesis report lookups strip the prefix, so the existing baseline files keep their names. Split into its own commit.
- `test_qkeras.py`: the behavior propagates through QKeras delegation (dilated `QConv2D` ingested, depthwise-shaped grouped `QConv2D` routed with its quantizers, `QLSTM(use_bias=False)` converts, `QGRU(reset_after=False)` parsed faithfully). The existing `test_qgru` passed `reset_after='False'` — a truthy string — so it never tested the `False` path; it now passes `True` explicitly.

**Test configuration** (OS, Python, ML framework version, HLS tool version): Linux (EL9), Python 3.11. Two environments: TensorFlow 2.14.1 / Keras 2.14 with the `jmitrevs/qkeras` fork (mirrors the default CI environment), and TensorFlow 2.21 / Keras 3.15 with qkeras-v3 and HGQ2 (mirrors the Keras 3 CI environment). C simulation via g++; no HLS tool involved.

## Breaking changes and migration

**What breaks:** Models using a configuration listed under "rejected" above previously converted silently (computing the wrong function) and now raise `NotImplementedError` naming the layer and option. Additionally, until the backend-rejection follow-up lands, the four ingested configurations (dilation, `reset_after=False`, `max_value`, non-concat `merge_mode`) parse correctly but still compute the wrong function if compiled — the follow-up closes exactly that gap.

**How users migrate:** Remove or replace the unsupported option (e.g. retrain with `reset_after=True`, drop `mask_zero`, use an explicit supported activation). Models that do not use any affected option are unaffected.

## AI assistance disclosure

Tool(s) and model(s): Claude Code (Fable 5)

Where it was used: the full branch — frontend audit, parser changes, IR attributes and all tests — under prompt-level direction with minor manual changes and review of the scope decisions (what is ingested vs rejected, test organization) by the author.

- [x] **I am the author of this contribution and take full responsibility for it.** I have read and understood every line and can explain and defend it in review.
- [x] I verified the generated code against real hls4ml and HLS semantics — no invented APIs, config keys, pragmas or citations.
- [x] All numbers, logs and test results quoted in this PR come from runs I actually performed, not from agent output.
- [x] I have the right to submit this work under the project's licence, and to my knowledge it does not reproduce third-party code that would conflict with it.

## Checklist

**Required:**
- [ ] I have read the [contributing guidelines](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I added tests under `test/pytest` covering this change (a bug fix should add a test that fails on `main` and passes here).
- [ ] I self-reviewed the full diff and it contains no leftover debug code, commented-out blocks or unrelated changes.

**If applicable:**
- [ ] Documentation under `docs/` updated.
- [ ] No new build or synthesis warnings.
- [ ] Public API / config schema changes are documented.

## Release note

```release-note
Fixed several Keras parser bugs; Keras layer options that hls4ml cannot faithfully implement now raise an error at conversion time instead of being silently misparsed.
```

### DIFF
--- a/hls4ml/converters/keras/convolution.py
+++ b/hls4ml/converters/keras/convolution.py
@@ -8,9 +8,7 @@ def parse_conv1d_layer(keras_layer, input_names, input_shapes, data_reader):
 
     layer = parse_default_keras_layer(keras_layer, input_names)
 
-    dilation = keras_layer['config'].get('dilation_rate', [1])[0]
-    if dilation != 1:
-        raise NotImplementedError(f'Layer {layer["name"]}: dilation_rate > 1 is not supported.')
+    layer['dilation'] = keras_layer['config'].get('dilation_rate', [1])[0]
 
     (*_, layer['in_width'], layer['n_chan']) = parse_data_format(input_shapes[0], layer['data_format'])
 
@@ -35,8 +33,29 @@ def parse_conv1d_layer(keras_layer, input_names, input_shapes, data_reader):
     layer['filt_width'] = keras_layer['config']['kernel_size'][0]
     layer['stride_width'] = keras_layer['config']['strides'][0]
 
+    groups = keras_layer['config'].get('groups', 1)
+    if groups != 1:
+        # groups == in_channels == filters is a depthwise conv; reshape the kernel from
+        # (kernel_size, 1, n_chan) to the (kernel_size, n_chan, 1) depthwise layout.
+        # Other grouped shapes have no hls4ml kernel and are rejected.
+        if (
+            groups != layer['n_chan']
+            or layer['n_filt'] != layer['n_chan']
+            or layer['class_name'] not in ['Conv1D', 'QConv1D']
+        ):
+            raise NotImplementedError(
+                'Grouped convolution is only supported as depthwise (groups == in_channels == filters); '
+                f'layer {layer["name"]} has groups={groups}, in_channels={layer["n_chan"]}, filters={layer["n_filt"]}.'
+            )
+        kernel = layer.pop('weight_data')
+        layer['class_name'] = 'DepthwiseConv1D'
+        layer['depthwise_data'] = kernel.reshape(*kernel.shape[:-2], layer['n_chan'], 1)
+        layer['depth_multiplier'] = 1
+
+    # Padding and output size follow the dilated (effective) kernel extent
+    dilated_filt_width = (layer['filt_width'] - 1) * layer['dilation'] + 1
     (layer['out_width'], layer['pad_left'], layer['pad_right']) = compute_padding_1d(
-        keras_layer['config']['padding'], layer['in_width'], layer['stride_width'], layer['filt_width']
+        keras_layer['config']['padding'], layer['in_width'], layer['stride_width'], dilated_filt_width
     )
 
     if layer['data_format'] == 'channels_last':
@@ -52,6 +71,12 @@ def parse_conv2d_layer(keras_layer, input_names, input_shapes, data_reader):
     assert 'Conv2D' in keras_layer['class_name']
 
     layer = parse_default_keras_layer(keras_layer, input_names)
+
+    dilation_rate = keras_layer['config'].get('dilation_rate', [1, 1])
+    if dilation_rate[0] != dilation_rate[1]:
+        # The conv config structs carry a single dilation value
+        raise NotImplementedError(f'Layer {layer["name"]}: asymmetric dilation_rate {dilation_rate} is not supported.')
+    layer['dilation'] = dilation_rate[0]
 
     (*_, layer['in_height'], layer['in_width'], layer['n_chan']) = parse_data_format(input_shapes[0], layer['data_format'])
 
@@ -78,6 +103,28 @@ def parse_conv2d_layer(keras_layer, input_names, input_shapes, data_reader):
     layer['stride_height'] = keras_layer['config']['strides'][0]
     layer['stride_width'] = keras_layer['config']['strides'][1]
 
+    groups = keras_layer['config'].get('groups', 1)
+    if groups != 1:
+        # groups == in_channels == filters is a depthwise conv; reshape the kernel from
+        # (*kernel_size, 1, n_chan) to the (*kernel_size, n_chan, 1) depthwise layout.
+        # Other grouped shapes have no hls4ml kernel and are rejected.
+        if (
+            groups != layer['n_chan']
+            or layer['n_filt'] != layer['n_chan']
+            or layer['class_name'] not in ['Conv2D', 'QConv2D']
+        ):
+            raise NotImplementedError(
+                'Grouped convolution is only supported as depthwise (groups == in_channels == filters); '
+                f'layer {layer["name"]} has groups={groups}, in_channels={layer["n_chan"]}, filters={layer["n_filt"]}.'
+            )
+        kernel = layer.pop('weight_data')
+        layer['class_name'] = 'DepthwiseConv2D'
+        layer['depthwise_data'] = kernel.reshape(*kernel.shape[:-2], layer['n_chan'], 1)
+        layer['depth_multiplier'] = 1
+
+    # Padding and output size follow the dilated (effective) kernel extent
+    dilated_filt_height = (layer['filt_height'] - 1) * layer['dilation'] + 1
+    dilated_filt_width = (layer['filt_width'] - 1) * layer['dilation'] + 1
     (
         layer['out_height'],
         layer['out_width'],
@@ -91,8 +138,8 @@ def parse_conv2d_layer(keras_layer, input_names, input_shapes, data_reader):
         layer['in_width'],
         layer['stride_height'],
         layer['stride_width'],
-        layer['filt_height'],
-        layer['filt_width'],
+        dilated_filt_height,
+        dilated_filt_width,
     )
 
     if layer['data_format'] == 'channels_first':
--- a/hls4ml/converters/keras/core.py
+++ b/hls4ml/converters/keras/core.py
@@ -1,5 +1,7 @@
 import math
 
+import numpy as np
+
 from hls4ml.converters.keras_v2_to_hls import get_weights_data, keras_handler, parse_default_keras_layer
 from hls4ml.model.quantizers import BinaryQuantizer, TernaryQuantizer
 from hls4ml.model.types import IntegerPrecisionType
@@ -76,7 +78,32 @@ def parse_activation_layer(keras_layer, input_names, input_shapes, data_reader):
     elif layer['class_name'] == 'ELU':
         layer['activ_param'] = keras_layer['config'].get('alpha', 1.0)
     elif layer['class_name'] == 'ReLU':
-        layer['class_name'] = 'Activation'
+        max_value = keras_layer['config'].get('max_value', None)
+        if max_value is not None and max_value == float('inf'):
+            max_value = None
+        negative_slope = keras_layer['config'].get('negative_slope', 0.0)
+        threshold = keras_layer['config'].get('threshold', 0.0)
+        if max_value is not None and (negative_slope != 0.0 or threshold != 0.0):
+            raise NotImplementedError(
+                f'Layer {layer["name"]}: ReLU with max_value combined with threshold or negative_slope is not supported.'
+            )
+        if negative_slope != 0.0 and threshold != 0.0:
+            raise NotImplementedError(f'Layer {layer["name"]}: ReLU must have threshold == 0 or negative_slope == 0.')
+        if max_value is not None:
+            layer['class_name'] = 'ClippedReLU'
+            layer['activation'] = 'clippedrelu'
+            layer['activ_param'] = max_value
+        elif negative_slope != 0.0:
+            layer['class_name'] = 'LeakyReLU'
+            layer['activation'] = 'LeakyReLU'
+            layer['activ_param'] = negative_slope
+        elif threshold != 0.0:
+            layer['class_name'] = 'ThresholdedReLU'
+            layer['activation'] = 'ThresholdedReLU'
+            layer['activ_param'] = threshold
+        else:
+            layer['class_name'] = 'Activation'
+            layer['activation'] = 'relu'
     elif layer['class_name'] == 'PReLU':
         if keras_layer['config'].get('shared_axes') is not None:
             raise Exception('PReLU with shared_axes other than None is not supported in hsl4ml')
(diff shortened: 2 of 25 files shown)


hls4ml/model/layers.py
```diff
@@ -851,11 +851,15 @@ class GlobalPooling1D(Layer):
     _expected_attributes = [
         Attribute('n_in'),
         Attribute('n_filt'),
+        Attribute('keepdims', value_type=bool, default=False),
         ChoiceAttribute('pool_op', ['Max', 'Average'], configurable=False),
     ]
 
     def initialize(self):
-        shape = [self.attributes['n_filt']]
+        if self.attributes['keepdims']:
```
JanFSchulte: Breaks (Q)ONNX frontend because the default defined above in the `expected_attributes` is not set when `initialize` is run, that happens afterwards. So using `self.get_attr('keepdims', False)` should fix this.

hls4ml/converters/keras/core.py
```diff
@@ -76,7 +78,32 @@ def parse_activation_layer(keras_layer, input_names, input_shapes, data_reader):
...
+        negative_slope = keras_layer['config'].get('negative_slope', 0.0)
+        threshold = keras_layer['config'].get('threshold', 0.0)
+        if max_value is not None and (negative_slope != 0.0 or threshold != 0.0):
+            raise NotImplementedError(
+                f'Layer {layer["name"]}: ReLU with max_value combined with threshold or negative_slope is not supported.'
+            )
+        if negative_slope != 0.0 and threshold != 0.0:
+            raise NotImplementedError(f'Layer {layer["name"]}: ReLU must have threshold == 0 or negative_slope == 0.')
+        if max_value is not None:
+            layer['class_name'] = 'ClippedReLU'
```
JanFSchulte, line 93: `ClippedReLU` should be added to the list of allowed activations in the keras v2 parser at https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/converters/keras_v2_to_hls.py#L231

JanFSchulte: I think it also needs to be handled in https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/model/layers.py#L981 to get the correct activation function name in the backend.

hls4ml/converters/keras/convolution.py
```diff
@@ -8,9 +8,7 @@ def parse_conv1d_layer(keras_layer, input_names, input_shapes, data_reader):
 
     layer = parse_default_keras_layer(keras_layer, input_names)
 
-    dilation = keras_layer['config'].get('dilation_rate', [1])[0]
-    if dilation != 1:
-        raise NotImplementedError(f'Layer {layer["name"]}: dilation_rate > 1 is not supported.')
+    layer['dilation'] = keras_layer['config'].get('dilation_rate', [1])[0]
```
JanFSchulte, line 11: That doesn't actually does not enable dilation, at least in Vivado/Catapult, since the calls for example here https://github.com/fastmachinelearning/hls4ml/blob/f98254690a5e9948ef111a033ff7a0d9a29c8741/hls4ml/backends/vivado/passes/im2col_codegen.py#L33 for 1D and 2D don't pass the dilation argument to the `generate_conv1d/2d_line_buffer_fn` functions and so it gets ignored. 

It also seems that in the 2D case the dilation parameters are not set correctly in https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/vivado/passes/convolution_templates.py, as they are not include. However, the actual code does not use them at all, so that doesn't really matter. We just set 2 parameters here to 1 that are never changed, but also never used https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/templates/vivado/nnet_utils/nnet_conv2d.h#L34-L35. Same I think for the 1D case, except that the value is set correctly there, but also not used.

JanFSchulte: Dilation is also not supported for streaming, at least it seems to give very wrong results. Also, oneAPI further rejects it in it's backend passes, so that would need to be adjusted https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/backends/oneapi/passes/convolution_templates.py#L93-L94

hls4ml/converters/keras/recurrent.py
```diff
@@ -158,6 +174,9 @@ def parse_bidirectional_layer(keras_layer, input_names, input_shapes, data_reade
     layer['n_timesteps'] = input_shapes[0][1]
     layer['n_in'] = input_shapes[0][2]
     layer['merge_mode'] = keras_layer['config']['merge_mode']
+    if layer['merge_mode'] is None:
```
JanFSchulte, line 177: Btw, seems like sum / ave / mul merge_modes are parsed correctly with the correct `n_out`, but that doesn't reach the kernel, which is written for `concat` anyway.

hls4ml/converters/keras/recurrent.py
```diff
@@ -44,20 +51,24 @@ def parse_rnn_layer(keras_layer, input_names, input_shapes, data_reader):
     )
 
     if layer['bias_data'] is None:
```
JanFSchulte, line 53: The birectional case down below is missing a handling of `bias=False` along the lines of the code here, resulting in a variety of crashes depending on the exact configuration.

hls4ml/model/layers.py
vloncar: Good catch — `initialize()` indeed runs before the expected-attribute defaults are applied, and the ONNX parser doesn't set `keepdims` at all. Fixed with `get_attr('keepdims', False)` in ee3ee99e7.

hls4ml/converters/keras/core.py
vloncar: Right, and it was a bit worse than a missing entry: since the returned layer dict carries an `activation` key, the dispatcher treated it as an embedded activation and split off a duplicate `Activation` layer. Added `ClippedReLU` to the list with a model-level test asserting the resulting layer structure.

hls4ml/converters/keras/convolution.py
vloncar: Known, there will be a followup PR that adds proper handling of (un)supported features. Still working on that. Doesn't break now, so it's fine here

hls4ml/converters/keras/recurrent.py
vloncar: Same as dilation rate, will be properly rejected once we land the feature matrix PR

hls4ml/converters/keras/recurrent.py
vloncar: Good catch, fixed. The bidirectional branch now substitutes zeros of the gate width (and the recurrent bias for GRU), same as the standalone path, and I've added numeric tests for Bidirectional GRU/LSTM with `use_bias=False`.

JanFSchulte (approved):

JanFSchulte:
Thanks for the fixes. Looks good to me. Do we want to merge it now or wait for the IR/backend PR and have it as a package?

vloncar:
I'd merge it now. The backend guards PR will affect all backends and future developments will need to be aware of that feature (it's not intrusive and difficult to manage, but it qualifies as a core change). I'd like a wider reflection on that PR before it is merged. In practice this means we talk about it in dev meetings and leave some time for interested people to think over. Keeping this PR and #1535 out until that lands introduces long delays for something that is already 90% helpful without the guards.

PR #1535, opened 2026-08-26
PyTorch frontend: stop silently misparsing unsupported layer configurations

vloncar:
# PyTorch frontend: stop silently misparsing unsupported layer configurations

## Description

This PR is the PyTorch half of the frontend audit for silent misparses started in the Keras parser PR (it is branched on top of that PR and should be rebased once it merges; the last two commits are the ones to review). The rule applied is the same: **never convert a model that computes different math than the one that was trained.**

**Fixed** (previously computed wrong values, or crashed far from the cause):

- *`'same'` padding*: the PyTorch padding helpers hard-coded `pad_top = 1`, used an output-size formula only correct for kernel size 3, and put the odd padding pixel on the opposite side from PyTorch — every `Conv1d(padding='same')` with kernel size other than 3 computed wrong values. `Conv2d` with string padding never reached the helper at all (it indexed into the string, `pad_top = 's'`). The `'same'` branches now implement PyTorch's actual convention (total padding `dilation * (kernel - 1)`, extra pixel on the bottom/right, stride 1 only, output size = input size), and `Conv2d` takes all four pads from the helper.
- *Which element of the RNN output tuple is consumed*: `nn.RNN/LSTM/GRU` return `(sequence, state)`, but the tracer discarded the index — `out, _ = self.lstm(x)` followed by per-timestep layers silently computed on the **last hidden state only**. The parser now inspects the consumed indices: the sequence output sets `return_sequences=True` (with the correct output shape), the final hidden state keeps the previous behavior, and stacked RNNs consuming the sequence output now convert (previously a `KeyError`). Consuming the LSTM cell state, consuming both tuple elements, or slicing the outputs raises a clear error.
- *RNN `bias=False`*: crashed with `AttributeError` (the bias tensors do not exist on the module), and the zeros substitute below the crash used the wrong dimension (the input width instead of the gate width).
- *Scalar merge operands*: the pre-pass that turns scalars in `add/mul/sub` into constant inputs always replaced argument 1 — `2 - x` replaced `x` and kept the scalar, computing garbage — and the merge output shape was copied from input 0, which for a scalar-first expression is the size-1 constant. The scalar's actual position is now used and the output shape comes from the tensor input.
- *`torch.cat` with a positional `dim`*: the dimension was read only from kwargs and defaulted to 0 (the batch axis), so `torch.cat([a, b], 1)` crashed (or, with a numeric batch entry, would have been silently wrong).
- *Functional calls with positional or omitted arguments*: `F.leaky_relu(x, 0.2)`, `F.elu(x, 0.5)`, `F.threshold(x, t, v)`, `F.softmax(x, dim)` and the functional poolings read their parameters only from kwargs — positional spellings crashed with `KeyError`, and `F.leaky_relu(x)` (defaults) crashed too. All of them now read positional-or-keyword with the correct defaults. The functional average poolings also hard-coded `count_include_pad=True`; the flag is now honored.
- *`view` with a literal batch size*: `x.view(4, -1)` included the literal batch in the `-1` deduction, producing a wrong dimension. The batch entry is now dropped before deducing.
- *`nn.Upsample(size=...)`*: `scale_factor` is `None` in this form, crashing with a `TypeError`. The output size is now taken from `size` when it is an integer multiple of the input size, and rejected otherwise.
- *`LayerNorm(elementwise_affine=False)`* (and `bias=False`): fed `None` weight data into the graph; identity values are now substituted, mirroring the Keras-side `center`/`scale` fix.

**Ingested faithfully into the IR** (matching the Keras PR; backend-side rejection is the immediate follow-up):

- *`nn.ReLU6`, `nn.Hardtanh(0, x)` and `F.relu6`*: previously rejected as unsupported layers; now ingested as the `ClippedReLU` parametrized activation introduced by the Keras parser PR. `Hardtanh` with `min_val != 0` stays rejected (it is not a clipped ReLU).

**Rejected with a clear error** (previously silently wrong):

- *`nn.Threshold(threshold, value)` with `value != 0`*: the layer outputs `value` below the threshold, but only the `value == 0` case maps to `ThresholdedReLU`; the value was silently dropped.
- *`LSTM(proj_size > 0)`*: the projection weights were ignored, giving shape mismatches downstream.
- *`BatchNorm*(track_running_stats=False)`*: normalizes with per-batch statistics even in eval mode, which no hls4ml kernel can represent.
- *Pooling with `ceil_mode=True`* (output sizes were silently computed with floor), *max pooling with `dilation != 1`* (silently ignored), *`AvgPool2d(divisor_override=...)`* (silently ignored).
- *`Conv2d` with different dilation factors per dimension*: the second factor was silently dropped (the IR carries a single dilation value, same as the Keras-side reject).
- *`torch.cat` along the batch dimension*, and *broadcasting in elementwise merges* (mismatched non-scalar shapes had the output shape of input 0 silently assumed).

The audit items that overlap open PRs are left out: the `unsqueeze` insert-position bug is fixed by #1496, and grouped/depthwise conv routing was already handled for PyTorch by #1524.

Relates to #1521, #1524, #1496 and the Keras parser audit PR.

## Type of change

- [x] Bug fix (non-breaking)
- [x] Breaking change (existing configs, APIs or generated code behave differently)
- [ ] New frontend / backend
- [ ] Documentation
- [ ] Build, CI or tooling

## Affected areas

**Backends:**
- [x] None directly

**Frontends:**
- [x] PyTorch

**Components:**
- [ ] IR (layers + model graph) / optimizer passes
- [ ] C++ templates / HLS sources under `hls4ml/templates/`
- [ ] Packaging, build or CI

| Backend | `io_type` | Strategy | Verified |
|---|---|---|---|
| all (parsing is backend-independent) | all | all | pytest (C simulation) with Vivado backend |

*The SimpleRNN numeric tests run on the Quartus backend, due to the known behavior of the Vivado SimpleRNN kernel.*

**New configuration axis introduced by this PR**: none. The `ClippedReLU` activation (introduced in the Keras parser PR) becomes reachable from the PyTorch frontend; no backend computes it yet, and the backend-rejection follow-up covers it for both frontends at once.

## Impact on generated HLS

**Numerical behaviour:** Only models using the affected configurations are impacted: (1) configurations listed under "rejected" no longer produce (wrong) generated code — conversion raises; (2) previously-wrong configurations (`'same'` padding with kernel size ≠ 3, sequence-consuming RNN models, scalar-first merges, functional average pooling with `count_include_pad=False`, `view` with a literal batch size) and previously-crashing ones (string-padded `Conv2d`, positional functional arguments, `bias=False` RNNs, `Upsample(size=...)`, `LayerNorm` without affine parameters, stacked RNNs) now generate correct code, verified against PyTorch predictions by C simulation. Models not using any affected configuration generate identical code. No synthesis was run — this is a frontend/parse-level change; the resource/latency table is deliberately left out.

## Tests

- `test_pytorch_api.py`: `'same'`/`'valid'` string padding swept over kernel sizes 1–7 (1D) and 3–4 (2D) against PyTorch predictions; functional activations and poolings with positional or omitted arguments (numeric); ReLU6/Hardtanh ingestion as ClippedReLU (parse assertions); rejects for asymmetric dilation, `Hardtanh(min_val != 0)`, `Threshold(value != 0)`, `ceil_mode`, pooling `dilation`, `divisor_override`; `view` literal-batch deduction.
- `test_recurrent_pytorch.py`: sequence-consuming models (`return_sequences=True`) for GRU/LSTM (Vivado) and SimpleRNN (Quartus) with more than one timestep, a stacked GRU on the sequence output, functional (`F.relu`) and method (`.flatten`) consumers of the sequence output, state-consuming models for GRU and LSTM, one RNN's final state as another's initial state (parse), `bias=False` for all three cells — all numeric against PyTorch except the state-chaining parse check; rejects for the cell state, both-outputs, sliced-output and `proj_size` patterns. (The pre-existing tests in this file consume the sequence output with a single timestep, which is why the misparse was invisible; they now parse as `return_sequences=True` and still pass unmodified.)
- `test_merge_pytorch.py`: scalar operands in both positions (`x+2.5`, `x-2.5`, `2.5-x`, `x*0.5`, numeric), positional `torch.cat` dim (numeric), rejects for batch-axis concatenation and non-scalar broadcasting.
- `test_batchnorm_pytorch.py` / `test_layernorm_pytorch.py` / `test_upsampling_pytorch.py`: `track_running_stats=False` reject; `elementwise_affine=False` numeric; `Upsample(size=...)` numeric and the non-integer-multiple reject.

**Test configuration** (OS, Python, ML framework version, HLS tool version): Linux (EL9), Python 3.11, PyTorch (CPU). C simulation via g++; no HLS tool involved.

## Breaking changes and migration

**What breaks:** Models using a configuration listed under "rejected" above previously converted silently (computing the wrong function) or crashed with a confusing error, and now raise `NotImplementedError` naming the layer and option. Models that consumed the RNN sequence output previously produced a last-state-only layer; they now convert to the sequence-producing layer they actually describe, so downstream shapes change to the correct ones.

**How users migrate:** Remove or replace the unsupported option (e.g. `ceil_mode=False`, `reset_after`-style retraining is not needed here; for RNNs consume either the sequence or the final state, not both). Models that do not use any affected option are unaffected.

## AI assistance disclosure

Tool(s) and model(s): Claude Code (Fable 5)

Where it was used: the full branch — frontend audit, parser changes and all tests — under prompt-level direction and review of the scope decisions (what is fixed vs ingested vs rejected, test organization) by the author.

- [x] **I am the author of this contribution and take full responsibility for it.** I have read and understood every line and can explain and defend it in review.
- [x] I verified the generated code against real hls4ml and HLS semantics — no invented APIs, config keys, pragmas or citations.
- [x] All numbers, logs and test results quoted in this PR come from runs I actually performed, not from agent output.
- [x] I have the right to submit this work under the project's licence, and to my knowledge it does not reproduce third-party code that would conflict with it.

## Checklist

**Required:**
- [x] I have read the [contributing guidelines](https://github.com/fastmachinelearning/hls4ml/blob/main/CONTRIBUTING.md).
- [x] I added tests under `test/pytest` covering this change (a bug fix should add a test that fails on `main` and passes here).
- [x] I self-reviewed the full diff and it contains no leftover debug code, commented-out blocks or unrelated changes.

**If applicable:**
- [x] Documentation under `docs/` updated.
- [ ] No new build or synthesis warnings.
- [ ] Public API / config schema changes are documented.

## Release note

```release-note
Fixed several PyTorch parser bugs, notably 'same' padding and RNN output handling; PyTorch layer options that hls4ml cannot faithfully implement now raise an error at conversion time instead of being silently misparsed.
```

### DIFF
--- a/docs/attr_doc_gen.py
+++ b/docs/attr_doc_gen.py
@@ -92,6 +92,14 @@ def print_attrs(attrs, file):
             file.write('  * Available in: ' + ', '.join(attr.available_in) + '\n\n')
 
 
+def strip_trailing_blank_lines(file_name='attributes.rst'):
+    """Ends the file with exactly one newline, as the pre-commit hooks require."""
+    with open(file_name) as file:
+        content = file.read()
+    with open(file_name, mode='w') as file:
+        file.write(content.rstrip('\n') + '\n')
+
+
 def write_all_attributes(all_layers_attrs):
     with open('attributes.rst', mode='w') as file:
         file.write('================\n')
@@ -127,6 +135,8 @@ def write_all_attributes(all_layers_attrs):
                 file.write('---------------------------\n')
                 print_attrs(attr_list.unique_backend_attrs, file)
 
+    strip_trailing_blank_lines()
+
 
 def write_only_configurable(all_layers_attrs):
     with open('attributes.rst', mode='w') as file:
@@ -141,6 +151,8 @@ def write_only_configurable(all_layers_attrs):
             config_attrs = attr_list.only_configurable
             if len(config_attrs) > 0:
                 print_attrs(config_attrs, file)
+
+    strip_trailing_blank_lines()
 
 
 if __name__ == '__main__':
--- a/docs/ir/attributes.rst
+++ b/docs/ir/attributes.rst
@@ -3646,4 +3646,3 @@ Backend-specific attributes
   * "LineBuffer" implementation is preferred over "Encoded" for most use cases. This attribute only applies to io_stream.
 
   * Available in: Vivado, VivadoAccelerator, Vitis, Catapult, Coyote
-
--- a/docs/ir/attributes.rst
+++ b/docs/ir/attributes.rst
@@ -87,19 +87,19 @@ Backend-specific attributes
 
   * The number of times each multiplier is used by controlling the amount of pipelining/unrolling. Lower number results in more parallelism and lower latency at the expense of the resources used.Reuse factor = 1 corresponds to all multiplications executed in parallel, and hence, the lowest possible latency.
 
-  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Libero, XLS
+  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Coyote, Libero, XLS
 
 * table_size: int (Default: 1024)
 
   * The size of the lookup table used to approximate the function.
 
-  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Libero, XLS
+  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Coyote, Libero, XLS
 
 * table_t: NamedType (Default: fixed<18,8,TRN,WRAP,0>)
 
   * The datatype (precision) used for the values of the lookup table.
 
-  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Libero, XLS
+  * Available in: Vivado, VivadoAccelerator, Vitis, Quartus, Catapult, SymbolicExpression, oneAPI, Coyote, Libero, XLS
 
 ParametrizedActivation
 ======================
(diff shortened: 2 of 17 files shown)


JanFSchulte:
Stress testing this one didn't reveal any issues not already flagged in the keras one. So lets see the test results. If there are no surprises, this one can also go in.

JanFSchulte (approved):

PR #1538, opened 2026-08-31
patch einsum, leakyrelu name

brightneedle:
# Description

## Summary

This PR fixes two issues in bit-exact precision and interval propagation:

- Recognizes both `leakyrelu` and `leaky_relu` activation names when inferring precision for `ParametrizedActivation`.
- Correctly applies direct-sum axes when propagating numeric and quantized intervals through Einsum operations.
- Uses `np.ones_like` when constructing fallback signedness arrays, preserving the expected shape and dtype.

## Motivation and context

Keras may serialize Leaky ReLU as `leaky_relu`, while the bit-exact optimizer previously handled only `leakyrelu`. The unmatched name caused the optimizer to use overly conservative fallback precision.

Einsum recipes may also contain input axes that must be reduced before transposition and contraction. The interval implementation ignored these direct-sum axes, producing incorrect shapes or bounds for equations containing input-only reduction dimensions. This PR aligns interval execution with the existing numeric Einsum recipe implementation.

## Dependencies

No new dependencies are required.

## Type of change

## Tests

Ran the existing quantized-interval test suite:
`pytest test/pytest/test_qinterval.py -q`

### DIFF
--- a/hls4ml/model/optimizer/passes/bit_exact.py
+++ b/hls4ml/model/optimizer/passes/bit_exact.py
@@ -693,7 +693,7 @@ def _(layer: Activation):
             i = np.minimum(i, 1)
             f = np.full_like(f, 126, dtype=np.int16)
         case _:
-            k = np.ones(k, dtype=np.int16)
+            k = np.ones_like(k, dtype=np.int16)
             i = np.full_like(i, 126, dtype=np.int16)
             f = np.full_like(f, 126, dtype=np.int16)
     return k, i, f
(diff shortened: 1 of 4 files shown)


vloncar:
Needs tests to showcase the fix. We don't accept "trust me bro" PRs 😛

vloncar:
Sorry, I just noticed, can you amend the commits to include your name? Looks like you didn't configure git fully, so you commit with your computer username and as "Your Name", and then it doesn't link back to you on GitHub. It will make it easier for us to note you as a contributor in the future.

JanFSchulte:
Thanks for the proposed fixes. In general, this looks good. There is one concern about the direct-sum fix in qinterval.py. It's good, but we have another entry point to einsum calculations in https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/utils/einsum_utils.py#L197 that would also need the fix. So the best solution would be to fix the calculations directly in `parse_einsum` in the `einsum_utils` so we don't have to apply the same fix in two different places.

PR #1540, opened 2026-09-02
Serialize NumPy arrays nested inside lists and tuples

### LINKED ISSUE #1526: _serialize_array_attrs() does not serialize NumPy arrays contained inside lists or tuples.
## Prerequisites

* [✓] Test that the bug appears on the current version of the master branch. Make sure to include the commit hash of the commit you checked out.
* [✓] Check that the issue hasn't already been reported, by checking the currently open issues.
* [✓] If there are steps to reproduce the problem, make sure to write them down below.
* [   ] If relevant, please include the hls4ml project files, which were created directly before and/or after the bug.

## Quick summary

`_serialize_array_attrs()` does not serialize NumPy arrays contained inside lists or tuples. As a result, architectures containing attributes such as `mask_kbi`, which can contain a tuple of NumPy arrays, fail when the resulting model architecture dictionary is serialized to JSON with `TypeError: Object of type ndarray is not JSON serializable`.

### Steps to Reproduce

1. Checkout the master branch, with commit hash: `b90fb06736baa0908a8995fc1cf4ac4a7d1c241f`.
2. Convert a model that produces an architecture attribute containing NumPy arrays inside a tuple, such as an HGQ model containing `FixedPointQuantizer` layers.
3. Serialize the model architecture dictionary to JSON.

A minimal example to reproduce the problem:

```python
import json
import numpy as np

model_arch = {
    "layer": {
        "mask_kbi": (
            np.array([[0, 1, 0]], dtype=np.int16),
            np.array([[1, 2, 3]], dtype=np.int16),
        )
    }
}

# Serialize the array attributes
_serialize_array_attrs(
    model_arch,
    "layer",
    dest_dir
)

# Serialize the resulting architecture
json.dumps(model_arch)
```

The current `_serialize_array_attrs()` implementation recursively processes dictionaries and directly encountered NumPy arrays, but does not recursively process arrays contained inside tuples or lists.

### Expected behavior

All NumPy arrays contained anywhere within the model architecture dictionary should be serialized to `.npy` files and replaced by the corresponding `@ndarray:...` references before the architecture is serialized to JSON.

For example:

```python
{
    "mask_kbi": (
        "@ndarray:layer_mask_kbi_0.npy",
        "@ndarray:layer_mask_kbi_1.npy",
    )
}
```

The resulting architecture should then be JSON serializable.

### Actual behavior

NumPy arrays contained inside tuples or lists are left unchanged.

For example:

```python
{
    "mask_kbi": (
        np.ndarray(...),
        np.ndarray(...),
    )
}
```

Attempting to serialize the architecture then raises:

```text
TypeError: Object of type ndarray is not JSON serializable
```

This occurs because `_serialize_array_attrs()` handles dictionaries and direct `np.ndarray` values, but does not recurse into tuple/list values.

### Possible fix

Extend `_serialize_array_attrs()` to recursively process lists and tuples, serializing any NumPy arrays contained within them. 

### Additional context

The issue is relevant to model architectures generated from HGQ-related layers. In the affected architecture, `mask_kbi` is represented as a tuple containing multiple `np.ndarray` objects with `dtype=int16`. The arrays are nested several levels inside the layer state.

cycsmail:
# Description

`_serialize_array_attrs()` only matches `np.ndarray` values, so arrays sitting inside a list or tuple are left untouched and `json.dump()` on the model architecture fails with "Object of type ndarray is not JSON serializable". The `mask_kbi` tuple of `FixedPointQuantizer` layers hits this. Fixes #1526.

The serializer now walks lists and tuples too (recursively, including dicts inside them) and saves each array as `<prefix>_<index>.npy` with the usual `@ndarray:` reference, so `mask_kbi` comes out as `["@ndarray:layer_mask_kbi_0.npy", ...]` like the issue expects. np.integer/np.floating scalars inside sequences are converted the same way they already are for dict values. I added the matching handling in `_deserialize_array_attrs()` as well, otherwise a saved `.fml` with `mask_kbi` loads back as a list of strings and the fix doesn't round-trip.

## Type of change

## Tests

New test in `test/pytest/test_serialization_utils.py`: builds a layer dict with a `mask_kbi` tuple of int16 arrays plus a nested list case, checks the `@ndarray:` references, that `json.dumps` then succeeds, and that a JSON round-trip plus `_deserialize_array_attrs` gives the arrays back.

```
pytest -q test/pytest/test_serialization_utils.py
```

1 passed with the fix, and 1 failed with `serialization.py` reverted to main, so it does catch the bug. The issue's minimal reproducer gives the expected output too. It only needs numpy and base hls4ml, unlike the end-to-end tests in `test_serialization.py`, which need qkeras/tensorflow/qonnx and a compiler and which I did not run locally. `ruff check`, `ruff format --check` and `flake8` with the repo's config are clean on both files.

**Test Configuration**: Linux, Python venv with numpy and hls4ml installed editable, no vendor tools.

## Checklist

### DIFF
--- a/hls4ml/utils/serialization.py
+++ b/hls4ml/utils/serialization.py
@@ -189,18 +189,37 @@ def _serialize_array_attrs(attr_dict, layer_name, dest_dir):
     for attr_name, attr_val in attr_dict.items():
         if isinstance(attr_val, dict):
             _serialize_array_attrs(attr_val, layer_name, dest_dir)
-        if isinstance(attr_val, np.ndarray):
+        if isinstance(attr_val, (np.ndarray, list, tuple)):
             # arr_name ensures a nicer name for the data of weight variables and avoids name-clashing
             arr_name = layer_name
             if 'name' in attr_dict and attr_dict['name'] != layer_name:
                 arr_name += '_' + attr_dict['name']
             arr_name += '_' + attr_name
-            serialized_name = _serialize_ndarray(attr_val, arr_name, dest_dir)
-            attr_dict[attr_name] = serialized_name
+            if isinstance(attr_val, np.ndarray):
+                attr_dict[attr_name] = _serialize_ndarray(attr_val, arr_name, dest_dir)
+            else:
+                attr_dict[attr_name] = _serialize_sequence(attr_val, arr_name, layer_name, dest_dir)
         if isinstance(attr_val, np.integer):
             attr_dict[attr_name] = int(attr_val)
         if isinstance(attr_val, np.floating):
             attr_dict[attr_name] = float(attr_val)
+
+
+def _serialize_sequence(seq, name_prefix, layer_name, dest_dir):
+    serialized = []
+    for idx, val in enumerate(seq):
+        if isinstance(val, dict):
+            _serialize_array_attrs(val, layer_name, dest_dir)
+        if isinstance(val, np.ndarray):
+            val = _serialize_ndarray(val, name_prefix + '_' + str(idx), dest_dir)
+        elif isinstance(val, (list, tuple)):
+            val = _serialize_sequence(val, name_prefix + '_' + str(idx), layer_name, dest_dir)
+        if isinstance(val, np.integer):
+            val = int(val)
+        if isinstance(val, np.floating):
+            val = float(val)
+        serialized.append(val)
+    return tuple(serialized) if isinstance(seq, tuple) else serialized
 
 
 def _serialize_ndarray(arr, name_prefix, dest_dir):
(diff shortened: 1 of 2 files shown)


hls4ml/utils/serialization.py
```diff
@@ -189,20 +189,39 @@ def _serialize_array_attrs(attr_dict, layer_name, dest_dir):
...
             attr_dict[attr_name] = int(attr_val)
         if isinstance(attr_val, np.floating):
             attr_dict[attr_name] = float(attr_val)
 
 
+def _serialize_sequence(seq, name_prefix, layer_name, dest_dir):
+    serialized = []
+    for idx, val in enumerate(seq):
+        if isinstance(val, dict):
+            _serialize_array_attrs(val, layer_name, dest_dir)
```
JanFSchulte, line 212: If there are 2 dicts in the sequence, this will result in their output files having the same name after serialization, so the second will overwrite the first. Is there a reason not to use the same `name_prefix + '_' + str(idx)` as below?

hls4ml/utils/serialization.py
```diff
@@ -189,20 +189,39 @@ def _serialize_array_attrs(attr_dict, layer_name, dest_dir):
...
+        if isinstance(val, np.ndarray):
+            val = _serialize_ndarray(val, name_prefix + '_' + str(idx), dest_dir)
+        elif isinstance(val, (list, tuple)):
+            val = _serialize_sequence(val, name_prefix + '_' + str(idx), layer_name, dest_dir)
```
JanFSchulte, line 216: Tuples get flattened into arrays in `json.dump`, and so get returned as lists instead of tuples after deserialization. Not a problem as we explicitly cast to tuples in the relevant places in the code, but we should add a comment here to document that.

JanFSchulte (changes requested):
Thanks, looks good to me. A have two minor comments before merging.

cycsmail:
Thanks for the review, both good catches.

1. Dicts inside a sequence now get the same per-index prefix as the other branches (`name_prefix + '_' + str(idx)`), so two dicts in one list no longer write to the same file. Added a test with two dicts in one list that each hold an array under the same key: both `.npy` files exist with distinct names and the round trip restores both arrays. The deserialize side needed no change since it just follows the `@ndarray:` references.

2. Added a comment where the tuple is rebuilt: `json.dump` flattens tuples, so they come back as lists after deserialization, and callers cast back to tuples where it matters.

JanFSchulte (approved):

PR #1541, opened 2026-09-06
Add Conv1D dilation support for io_parallel

AaravGaurdev:
# Description

Add dilated Conv1D support for Vivado and Vitis using `io_parallel`.

The requested dilation rate is now passed to the existing Conv1D im2col code. The `io_stream` path does not support dilation yet and raises an error.

## Type of change

This addresses the Conv1D dilation bug reported in #1508.

## Tests

Ran the Conv1D dilation and Keras converter tests:

```bash
python -m pytest -q test/pytest/test_conv1d_dilation.py test/pytest/test_keras_converter.py
```

All 35 tests passed. Pre-commit checks also passed.

**Test Configuration**:

- Python 3.13.12
- TensorFlow 2.21.0
- Keras 3.15.1
- Vivado and Vitis with `io_parallel`

## Checklist


JanFSchulte:
Thanks! Three main comments:
- Did you ever run HLS synthesis for a model generated with this configuration? Successful C++ emulation is great, but we would like to see that it produces synthesizable HLS code as well.
- This should probably also be added for the separable Conv1D case in the Vivado backend pass
- The ONNX converter sets a different variable name for the dilation parameter, so this fix would not work in that case: https://github.com/fastmachinelearning/hls4ml/blob/main/hls4ml/converters/onnx/convolution.py#L52. This would be a good opportunity to fix that and make it consistent with the keras one.

AaravGaurdev:
Thanks for the feedback. I’ve run C++ simulation but not HLS synthesis yet, so I’ll do that next. I’ll also update SeparableConv1D and fix the dilation attribute name in the ONNX converter.

PR #1544, opened 2026-09-08
Add runtime-selectable weights for BramFactor parameters

dimdano:
# Description

This PR adds runtime-selectable parameter banks for hls4ml designs using externally exposed parameters. The main goal is to allow multiple trained parameter sets of the same model architecture to reside on the FPGA and select one per inference, without modifying or re-running HLS for the synthesized compute IP. 

Parameters that may vary between banks are selected explicitly per layer with ``ExternalParameters``.

1. ``hls_model.build(synth=True)`` produces the normal hls4ml compute IP with selected parameters exposed.
2. ``pack_banks()`` validates and packs complete model parameter sets.
3. ``package()`` verifies the synthesized parameter interfaces and generates banked storage, wrapper/control RTL, and Vivado Tcl.
4. Vivado builds the complete design from the original HLS RTL and generated banking logic.

The synthesized hls4ml compute IP is kept unchanged. The manifest records the parameter layout and interface expected by hls4ml, while synthesis reports and generated RTL are used to verify the actual interface before packaging. Unsupported layouts or interface geometries are rejected. Agentic AI was used as a development aid for probing, tests, and bug fixing.

Current verified scope is Vitis, ``io_parallel``, ``Strategy: Resource``, Dense layers over rank-1/2/3 inputs ( lowered ``PointwiseConv1D/2D`` layers are also handled) , and ``ap_fixed`` parameters using ``TRN``/``WRAP``. Dense reuse factor 1 (no addressable BRAM interface) and packed words wider than 4096 bits (existing HLS wide-interface limit) are outside the implemented flow and out of scope.

## Type of change

- [x] Documentation update

## Tests

Tests cover:

- manifest generation, packing, validation, and unsupported configurations;
- Dense and Pointwise parameter layouts across supported reuse factors;
- frontend consistency for Keras and PyTorch;
- synthesized-interface inspection and refusal of incompatible RTL;
- end-to-end RTL simulation of multi-bank Dense and Pointwise designs;
- bank-selection/handshake behavior and invalid bank IDs;
- Vivado synthesis of the generated wrapper and Tcl flow.

Tested on Vitis 2024.2 and 2025.2. Synthesis and RTL integration tests are needed because the key risks are in the generated hardware interface, timing, and wrapper behavior, which Python-level tests cannot validate.

Run with or without synthesis/RTL tests:

```bash
RUN_SYNTHESIS=true pytest -q \
    test/pytest/test_external_parameters.py \
    test/pytest/test_runtime_weights.py
```

## Checklist


vloncar:
I think it would be better to adopt a new name for this feature and then we can retire/remove `BramFactor`. This is a different implementation that only briefly touches the (naively-implemented) feature with the same name. There are already issues and prior mentions of `BramFactor` that you don't want to be associated with.

As for the implementation itself, one early comment i can give is that you seem to rely on constructing and passing around plain dicts instead of objects. This is very error prone in the future when things inevitably change, maing it harder to maintain. Make them objects, later on you can add more information to them to make them smart enough to know how to write themselves as a string. This makes writer component simpler, and overall more in line with the rest of the codebase.

Also, function for parsing reports already exists. If it is missing features you need, better to add them there rather than creating another parser.

